PR #5, opened 2010-12-22
html 5 audio

a-zb:
Added sound.seek(tm) where tm is a time to seek to.
Added a little longer shoot.ogg version of shoot.wav.
Modified game.js in elevator demo to use above.


louisstow:
Will close this and leave the other HTML5 audio open

PR #20, opened 2011-01-20
Mouse.areaMap() bug fixing

adngdb:
In mouse.areaMap(), if arrays are passed (for example, areaMap([0,0], [0,10], [10,10], [10,0]); ), points are modified to offset from the origin of the screen and not the origin of the object. This is explained in the API : http://craftyjs.com/api/mouse-9/-areaMap--.html

But, it is not the case when you give a polygon to areaMap. I guess it is a mistake, so here is a simple patch to this bug. 

Simply adding : 
poly.shift(this.x, this.y);

When there is only a single parameter, meaning a Polygon object. 

Thanks for this brilliant framework ! :)

Adrian

### DIFF
--- a/src/controls.js
+++ b/src/controls.js
@@ -93,6 +93,9 @@ Crafty.c("mouse", {
 				args[i][0] += this.x;
 				args[i][1] += this.y;
 			}
+			else {
+			    poly.shift(this.x, this.y);
+			}
 			
 			poly = new Crafty.polygon(args);
 		}


louisstow:
Thanks for that. Will add for the next version. I plan to introduce SAT collision so polygons may get a bit of a makeover.

louisstow:
Fixed

adngdb:
What you did to fix this isn't right: if you always do this.map.shift(this._x, this._y);, then you have to remove the first manuel shift on the arguments. For the moment, it works if a polygon is given, but not anymore if arrays are given. ;-)

louisstow:
Oops my mistake.

```
//offset the polygon if it was passed through arguments
if(arguments.length === 1) 
    this.map.shift(this._x, this._y);
```

How does that look?

adngdb:
Ow I saw my commit, ha ha that's stupid: of course the 

<pre>else {
    poly.shift(this.x, this.
}</pre>


had to be at the end of the if () and not after the for (). 

But in the way you wrote it, why don't you just remove the lines 

for(;i<l;i++) {
    args[i][0] += this.x;
    args[i][1] += this.y;
}

It will work fine then, and you optimize a little your code.

louisstow:
Nice. Commited and added to the changelog crediting you.

PR #37, opened 2011-02-20
Add a function to return desired FPS

danhealy:
Is there a better way to get this value?  I am writing an animation component and need it.  I know I could just save the number used to initialize, but it seems like this should be available from the core.


louisstow:
Sure will add to the core.

I would like a method that calculates the real FPS based on previous ticks. Will have to do some work on that.

danhealy:
I added a change to animation that allows you to specify a FromX index after a ToX and have it animate backwards.

louisstow:
Maybe add support for Stats.js

PR #52, opened 2011-03-18
Adding optional callback for onHit. Will be called everytime two objects just stopped colliding.

ponychicken:
Through this one could do a action menu which appears when near to an object and disappears when leaving.


louisstow:
Cheers will merge with core and credit you.

Do you have a website or name you want me to put in?

ponychicken:
cool, could you use leo-koppelkamm.de / Leo Koppelkamm

PR #61, opened 2011-04-02
2 fixes

sorenbs:
.requires made chainable and the dependency of collision component on 2D made explicit with requires.

### DIFF
--- a/crafty.js
+++ b/crafty.js
@@ -1291,6 +1291,10 @@ Crafty.polygon.prototype = {
 
 Crafty.c("collision", {
 
+    init: function() {
+        this.requires('2D');
+    },
+
 	collision: function(poly) {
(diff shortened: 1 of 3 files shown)


sorenbs:
ahh, i see. I did find crafty.js to be a bit unwieldy :-)

Is it okay with you that i poke around and submit pull requests when i find something, or is there some other way you would prefer?

louisstow:
How about I give you commit permission and you create a branch?

louisstow:
Ok you have permission now. May I ask that you read through the style guide? https://github.com/louisstow/Crafty/wiki/Style-Guide 

I need to follow the rules too haha. Now that Crafty is getting quite big, need to take some measures to keep it clean.

PR #111, opened 2011-08-11
Isometric view - diamond rather than rectangle

ajacksified:
I've replaced a few lines of code for the isometric view, so that the output is a diamond rather than a rectangle. Most isometric applications that I've come in contact with are far more natural to develop using this paradigm. The X value increases along the top-left edge, and the Y value increases along the bottom-right edge.

### DIFF
--- a/src/isometric.js
+++ b/src/isometric.js
@@ -35,12 +35,11 @@ Crafty.extend({
 		* @see Crafty.isometric.size
 		*/
 		place: function(x,y,z, obj) {
-			var m = x * this._tile + (y & 1) * (this._tile / 2),
-				n = y * this._tile / 4,
-				n = n - z * (this._tile / 2);
-				
-			obj.attr({x: m  + Crafty.viewport._x, y: n  + Crafty.viewport._y}).z += z;
-			return this;
+			var xPos = .5 * (x * this._tile) + (y * .5 * this._tile);
+      var yPos = ((.25 * this._tile) * (y - x)) + this._tile + (-1 * .5 * z * this._tile);
+
+      obj.attr({ x: xPos,  y: yPos }).z += z;
+      return this;
 		}
 	}
 });


ajacksified:
I've put together a demo here: http://www.olivinelabs.com/labs/crafty/iso.html (I apologize for the quality of the code, it's a retrofit of the first crafty demo I've ever put together, and it was done somewhat hastily)

louisstow:
This looks good!

I had a thought, maybe isometric could be a component instead of a namespaced function.

```
Crafty.c("DiamondTile", {
      size" 0,

      DiamondTile: function(size) {
           this.size = size;
      },


       place: function(x, y, z) {
            ....
       }
});
```

That way we could have both.

ajacksified:
Yeah, that makes sense. That feels a little more natural with the way the rest of the crafty system works.

louisstow:
Great. So I'll change the current Crafty.isometric to a component. Should I call it 'RectIsometric' or something and yours `DiamondIsometric`?

ajacksified:
Sure, that sounds good.

ajacksified:
Whoops, wrong button. Closed instead of commented. Although, this could probably be closed anyway, since it's being rewritten rather than pulled.

hex:
What is the status of this issue? Has the diamond grid been implemented? I am looking in the snapshot and I can't find it.

louisstow:
Waiting on a new pull request.

techwraith:
Hey @ajacksified what's the status on this? I'd love to use it :)

ajacksified:
Hey, @Techwraith; I misunderstood and thought @louisstow was rewriting the old isometric engine _and_ my changes as components. I'll rewrite mine and add it in as a pull request sometime soon!

techwraith:
@ajacksified Thanks dude, looking forward to being able to use this :)

louisstow:
My bad. I did change it but didn't commit to master because API changes are annoying and invalidate existing tutorials.

ajacksified:
@louisstow If you throw it in a separate branch, I can copy the pattern and apply my changes; or are you planning on leaving isometric in as core functionality instead of a module indefinitely?

sorenbs:
I would love to get your thoughts on the layer proposal here https://groups.google.com/forum/?fromgroups#!topic/craftyjs/de0qGN4u-Ho

The basic idea is that we introduce layers as a general abstraction in Crafty. Entities can be added to a layer and layers can be configured to do different things to its entities. Some basic examples are z-ordering with respect to other layers, transparency, shifting the entire layer etc. I think we will do this no matter what. With this system in place I thought that we might as well use it to define the way entities are positioned. In that way the isometric placement is configured on the layer, instead of in the entity (using a component) or explicitly (using a placer object).

ajacksified:
I think that having some kind of "layout" layer that handles a pixel-based or isometric (or anything else) makes more sense conceptually. In fact, in order to get the diamond isometric view to work properly, I had to know the total width anyway (due to an issue somewhere deep within crafty's collision detection that doesn't like negative z-values.)

With the entities added to a layer, the entity could have a "move" function exposed that fires off an event; the layer would then changes the entity's position relative to units on its layer (pixels, or isometric diamonds [or even portions of an isometric square]).

That said, these layers could still be components - no reason the layer system itself can't be a component, just like regular entities.

PR #117, opened 2011-08-25
added callback function to tweens

jhocking:


mpetrovic:
What does this solve that binding to TweenEnd doesn't?

jhocking:
I didn't know about the TweenEnd binding because that's not on the main
website yet; doing it with a callback function seems like a simpler approach
to accomplishing the same goal, but yeah it's the same goal.

I just started using Crafty yesterday and wanted a callback function for
tweens (the build on the website doesn't have that). After I added that and
suggested the change to Louis, he asked me to put it in the engine myself
and submit a pull request.

Louis mentioned the new TweenEnd binding in his reply to me, so I guess he
wanted both approaches on the repository and then people can debate on the
merits or something. For example, can you send parameters to the TweenEnd
function like you can with a callback?

-Joe Hocking
www.newarteest.com

mpetrovic:
No, the only parameter TweenEnd gives is which property the tween is ending for. 

What parameters would you expect to be used?

jhocking:
Well I used the parameter to pass the originating object to the callback, so
that the tweened object can be deleted when the tween completes. That
particular parameter might be unnecessary though, since I can access that
through the tween's target or something.

More generally, it's just handy to be able to pass different parameters to
the callback function, so different tweens can trigger different reactions
without needing to repeat code with a bunch of different functions.

-Joe Hocking
www.newarteest.com

mpetrovic:
As with any event binding, you can access the object the event is firing for with 'this'. 

I can see the case for more parameters in a function called at the end of a tween. If the current parameter list isn't enough, I think we should come up with a list of useful parameters to pass into TweenEnd bindings and pass them all into it. This'll be more consistent with the rest of Crafty, and more consistent across TweenEnd callbacks.

jhocking:
Well, how about instead of making a list of parameters (since really those
parameters could be anything at all; for example, maybe the user wants to
have some other object in the scene affected, and it's a different object
for each tween) there could be an optional callback_params object that gets
passed to the tween() function. So like:

object1.tween({x: x_60, y: y_60}, 4, {object2});

And then in the function bound to TweenEnd you can access the tween's
callback_params. That syntax looks a little goofy to me though; it seems
clearer to explicitly say "this is the function, and these are the
parameters" like so:

object1.tween({x: x_60, y: y_60}, 4, onFinish, {object2});

-Joe Hocking
www.newarteest.com

jhocking:
on reflection though, it's just as easy to just attach those callback
parameters to the object directly. I keep forgetting that JavaScript is so
dynamic that you can just stick new fields onto an object on the fly.

Kind of ugly syntax though, I dunno I'll have to test this in my project to
see how it'd work. Gimme a minute.

-Joe Hocking
www.newarteest.com

mpetrovic:
objects like that would be available anyway, even without attaching them to the object itself

``` javascript
object2 = something
object1.tween({x: 30, y: 40}, 50).bind("TweenEnd", function (prop) {
    object2.doSomething();
});
```

is perfectly valid and will reference the same object2 variable. The only reason it wouldn't work is if the function wasn't defined in the same scope as the variable.

The only arguments an event handler would need is something to differentiate between 2 triggers of the same event.

jhocking:
huh I'm pretty new to all this, so I didn't know you can do that. I'll try
that in my testing project, that looks like it'll do what I need.

-Joe Hocking
www.newarteest.com

jhocking:
hm well I actually meant a different object2 for each tween, so something
like the tweened objects are created in a loop and the TweenEnd callback
uses a different entry from the array of objects. In my case, I have a
series of lights and a different light goes off for each tween:

for (var i = 0; i < 4; i++) {
    var x = Math.pow(-1, i);
    var y = 1;
    if (i > 1) {y = -1;}

```
var explode = Crafty.e("2D, Canvas, Tween, explode");
explode.tween({x: x*60, y: y*60}, 4).bind("TweenEnd", function f(obj) {
    lights[i].on();
    this.destroy();
    this.unbind("TweenEnd", f);
});
this.attach(explode);
```

}

Does that work? I haven't tested it yet, but it seems like that would only
use lights[3] for all the callback functions. Regardless, I'm starting to
think I should just revert the changes I made and use the TweenEnd binding.
As I explained, I only committed those changes after Louis suggested I
should.

-Joe Hocking
www.newarteest.com

mpetrovic:
I'm not sure how that'll play out. Try it and see what happens. My first guess is that it would use light[3] for all of them, but closures are a tricky thing.

jhocking:
Well I tested that and it turns out the array "lights" doesn't even exist
within the bind function, because I got an error. Oh well, it was simple
enough to just set explode.light=lights[i]; before setting the tween, and
then this.light.on(); inside the bind function.

Sorry for wasting your time chasing down a feature that was already in the
development version. Another question: I notice the Tween component used to
have the line this.unbind("EnterFrame", d); but that's been commented out.
Why isn't that needed any more?

-Joe Hocking
www.newarteest.com

mpetrovic:
Before last weekend, every time .tween() was called, Crafty would add an EnterFrame handler. When the tween was over, it would remove it. Having multiple handlers caused some issues, namely, confusing behavior when .tween() was called in the middle of an already running tween. So I refactored Tween to use a single EventHandler and a list of properties. Now there's only one event handler, it's only bound the first time tween is called, it handles tweening every property, and it should never be unbound until the entity is destroyed.

PR #119, opened 2011-08-25
"recompiled" minified code

jhocking:
After I reverted the tween callback in order to use the new TweenEnd binding I noticed that wasn't in the minified code yet.


louisstow:
If you made some changes to the source, could you modify the files under `src/`. Or is that a new build?

jhocking:
It's just a new build, no changes to the source. I'm new to this project so
sorry if I got mixed up; do you not update the minified code along with
source?

-Joe Hocking
www.newarteest.com

jhocking:
I had initially modified the source in order to add a callback function to
tweens, but after a discussion with user mpetrovic I decided to abandon my
changes and use the TweenEnd binding that I didn't know about because it's
not on the website yet.

-Joe Hocking
www.newarteest.com

jhocking:
okay now I've made a few changes to src/

I added getters for .parent and .numChildren to the 2D component. This way an object can keep track of what it's attached to, and how many children that parent has.

PR #128, opened 2011-09-27
That bug with runaway tweens: misused 'delete'

mildmojo:
Here's the one-line fix for runaway tweens that I mentioned on Twitter. Changed 'delete prop;' to 'delete this._step[k];' inside a 'for prop in...' loop. Here's some info from people who understand JS a lot better than me:

http://stackoverflow.com/questions/208105/how-to-remove-a-property-from-a-javascript-object
http://perfectionkills.com/understanding-delete/

### DIFF
--- a/crafty.js
+++ b/crafty.js
@@ -4247,7 +4247,7 @@ function tweenEnterFrame(e) {
 		this[k] += prop.val;
 		if (prop.rem-- == 0) {
 			this.trigger("TweenEnd", k);
-			delete prop;
+			delete this._step[k];
 			this._numProps--;
 		}
 	}


mildmojo:
Er, sorry, this should be against the dev branch.

mildmojo:
Or maybe not... The dev branch tween code is completely different. I guess this is useful for a point release.

mpetrovic:
Make the change in the src files. We compile crafty.js from those.

The dev branch uses old Tween code, which is why its so different.

mildmojo:
Gotcha, resubmitted. I'm still new to forks & pull requests. Thanks for being patient.

Issue #137, opened 2011-10-24
Collision normal bug


maxencefrenette:
There is defintely a bug with SAT collision normal. I don't understand the code enough to fix it.

Blame : 6a950d73026f736cc58fc9e2153e1afb5bd1db13

mpetrovic:
what's the bug?

gworley3:
It's certainly possible that there's a bug because it doesn't detect the side hit by position but by amount of overlap.  This allows the use of SAT without introducing a lot of extra code to keep track of the sides and it's usually right.  What's the bug, though?  Can you show us how to reproduce it?

maxencefrenette:
The normal is wrong.

maxencefrenette:
Oh, sorry, didn't saw your comment, growley3.

First of all, i think that the code you added at line 241 should also be added between line 198 and 199. Second of all, your identation is wrong. And finally, here is a fiddle : http://jsfiddle.net/8Nns4/ .

gworley3:
The return is only at 241 because we need to loop through all the sides of both polygons.  As for indentation, well, I use spaces instead of tabs in my editor so I never noticed it, but if it bothers someone they're welcome to fix it.  Finally, as for the fiddle, the issue appears, without actually stepping through the code, to be that you were expecting a normal of (-1, 0), but got (0, 1) because the amount of overlap for the left side was less than or equal to the overlap for the top side, and the top side was considered first by SAT.

I can't think of a good solution right now.  Overlap works well for guessing which side was hit when you have a small polygon hitting a bigger one, but in this case I see the problem.  The ideal solution is to figure out which side was hit, but the current SAT implementation doesn't really allow it and putting in that feature will technically make the algorithm non SAT.  Not sure what effect a change like this might have on speed, but I'd encourage you to take a look at the problem if it's affecting you.

maxencefrenette:
I dedicate this bugfix to chrome developer tools and its breakpoints :) .

louisstow:
Fixed see commit https://github.com/louisstow/Crafty/commit/7db85a109184fc421d13c3e9c58c4a57bf1b7e1d

PR #138, opened 2011-10-24
Fix for #137

maxencefrenette:


louisstow:
Changes to crafty.js don't mean anything because it gets generated. Just an FYI if you didn't know already.

maxencefrenette:
Yup, i know. I just generated it to debug. Should i discard changes to crafty.js with a new commit ?

louisstow:
Fixed in latest commit

PR #139, opened 2011-10-27
I needed the ability to pass my own explicit IDs to Crafty.e(), so I added it.

gabehollombe:
Woudl you consider adding this to core?

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -742,20 +742,30 @@ Crafty.extend({
 	* @param componentList - List of components to assign to new entity
 	* @sign public Entity Crafty.e(String component1[, .., String componentN])
 	* @param component# - Component to add
-	* Creates an entity. Any arguments will be applied in the same 
+	* @sign public Entity Crafty.e(Object id, String componentList)
+	* @param id - A specific ID to use for this entity
+	* @param componentList - List of components to assign to new entity
+	* Creates an entity. Any componentList arguments will be applied in the same 
 	* way `.addComponent()` is applied as a quick way to add components.
 	*
 	* From here, any component added will augment the functionality of 
 	* that entity by assigning the properties and methods to that entity. 
 	*/
 	e: function() {
-		var id = UID(), craft;
+		var params = Array.prototype.slice.call(arguments);
+		params.shift();
+		if (params.length == 2) {
+			id = params.shift()
+		}
+		else {
+			id = UID();
+		}
 		
 		entities[id] = null; //register the space
 		entities[id] = craft = Crafty(id);
 		
-		if(arguments.length > 0) {
-			craft.addComponent.apply(craft, arguments);
+		if(params.length > 0) {
+			craft.addComponent.apply(craft, params);
 		}
 		craft.addComponent("obj"); //every entity automatically assumes obj
 		


sorenbs:
That looks usefull. The old code supports this case:
Crafty.e("Comp1", "Comp2")
The new code does not, but i don't know if anybody actually uses that.

louisstow:
Why would you need a custom ID? For DOM elements?

gabehollombe:
> Why would you need a custom ID? For DOM elements?
> Because I'm creating entities server-side and sending serialized
> attributes over the wire to the client, and I want the client entities
> to end up with the same IDs that I've made for them on the server.

-g

On Fri, Oct 28, 2011 at 8:51 AM, Louis Stowasser
reply@reply.github.com
wrote:

louisstow:
Ah I see. I agree with sorenbs. Maybe you could check if the first argument is a Number than assume it's an ID.

sorenbs:
I ended up solving the same scenario by having a server side mapping between client IDs and server IDs. Keeping the IDs in sync is definitely a simpler approach. One drawback though is that you will have to generate all IDs serverside, or else the next serverside generated id will clash with an existing entity. Unless, of cause you generate the IDs some other way on the server, say by starting at a very large number and decrementing.

Anyway, i think the number check would be a good solution.

gabehollombe:
Why do IDs have to be numbers? Any unique string should be OK for IDs
too, right?

On Fri, Oct 28, 2011 at 11:59 AM, sorenbs
reply@reply.github.com
wrote:

gabehollombe:
Why do IDs have to be numbers? Any unique string should be OK for IDstoo, right?
On Fri, Oct 28, 2011 at 11:44 AM, Louis Stowasser
reply@reply.github.com
wrote:

louisstow:
It's an internal thing. Like event handlers and the selectors.

sorenbs:
Yeah, the fact that IDs are numbers can be exploited in the implementation, like with the number checking trick above. Even if it is possible to have entities with IDs that are not numbers i think it would be best not to, such as not to constrain future changes to Crafty.
Any compelling reasons why IDs should be arbitrary strings?

gabehollombe:
Nah. I'm happy for Crafty to require integer IDs on entities.  I'll
put that change together and submit another pull request. =-)

On Fri, Oct 28, 2011 at 12:57 PM, sorenbs
reply@reply.github.com
wrote:

mpetrovic:
wouldn't it be easier to store the server id as an attribute on the client side entity? attributes could be anything, so just Crafty.e('Server').attr('server_id', 5) would work fine.

louisstow:
I agree with Matt. Better to assign an attribute that won't potentially break something internally.

gabehollombe:
Great observation, Matt.  But does Crafty provide a way to access entities
by attr values?  I'd like to be able to fetch entities by server_id, for
example.

-g

louisstow:
Nah only components. Though could easily just create an object where the key is the server_id and the value is the entity.

mpetrovic:
Extending Crafty is trivial. Adding a method to find an entity with a given attribute would be simple.

gabehollombe:
_nod nod_ Both ways are great and valid suggestions. I was just wondering
if Crafty had anything like that built-in before I go adding my own stuff.
=-) Thanks, gents.  Let's close this pull request and leave things as they
are; I'll use a different attribute to store server-side-id on my crafty
entities client-side.

-g

PR #140, opened 2011-10-31
Rounding issue on tween

ghost:
when using mycharacter.tween({x: -18, y: -96}, 100); the x and y properties ends like this: up: -95.68px; left: -17.36px; instead of x=96 and y=18 as its supposed to be

### DIFF
--- a/src/animation.js
+++ b/src/animation.js
@@ -218,7 +218,7 @@ Crafty.c("Tween", {
 			}
 			
 			for (var prop in props) {
-				this._step[prop] = {val: (props[prop] - this[prop] )/duration, rem: duration};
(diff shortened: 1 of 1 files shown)


mpetrovic:
Something's not right here. The final value of this[k] should've been the number given when Tween is called. This is a band-aid covering up a real problem in Tween.

louisstow:
You're right. My initial thought was floating point inaccuracy. Then maybe it wasn't ending on the correct number of frames. Really not sure.

mpetrovic:
The final value that goes to k should be an integer, but we would need to
store the current value as a float. I'm not sure why the final value is a
weird float unless something else is rounding the values.

PR #162, opened 2012-01-04
HitBox Components

BlackScorp:
This Components displayes Crafty.polygon wired or solid(uses for debugging only), but it has some bugs Actually. if your Viewport is modified the hitboxes will be displayed several times. clearrect would help but adding clearRect didnt displayes the Wired Hitboxes. 

Maybe anyone can help on fixing it

### DIFF
--- a/src/hitbox.js
+++ b/src/hitbox.js
@@ -1,0 +1,65 @@
+/**@
+* #Collision
+* @category 2D
+* Components to display Crafty.polygon Array for debugging collision detection
+*/
+Crafty.c("WiredHitBox", {
+    init:function(){
+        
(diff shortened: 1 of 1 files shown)


ahilles107:
You must provide valid docs for this component.

BlackScorp:
it is a component, i thought it is clear if you just make Crafty.e("2D , Canvas, Collision,WiredHitBox").collision(new Crafty.polygon([10,20],[20,50],[20,30]));  but first i need help on fixing it, maybe you can help me?:D

louisstow:
If you just add an example that should be fine. Will pull anyway.

PR #164, opened 2012-01-05
a small bug;)

BlackScorp:


mpetrovic:
I wouldn't worry too much about any of the php code. It's not ready for use at all. The fact that it's there is due to an error on my part when merging storage into master.

PR #168, opened 2012-01-07
Matrix2D addition and Vector2D changes

dubrowgn:
- Added Crafty.math.Matrix2D class
- Made some changes to Vector2D:
  - Added passing Vector2D instance as parameter in constructor
  - Added .angleBetween(Vector2D)
  - Renamed .getAngle() to .angleTo() and simplified
  - Renamed .getCopy() to .clone() for consistency with Crafty
  - Renamed .squaredDistance to .distanceSq() for consistency
  - Added .divide(Vector2D)
  - Added .equals(Vector2D)
  - Renamed .getLength() to .magnitude() for consistency (.length is already a member of Function)
  - Added .magnitudeSq()
  - .multiply() now only accepts Vector2D for consistency with .add(), .divide(), and .subtract()
  - Fixed a potential bug in .normalize() when normalizing vectors of magnitude 0
  - Added .scale(Number[, Number])
  - Renamed .setLength() to .scaleToMagnitude() for consistency
  - .setValues() now accepts x and y coordinates in addition to another Vector2D instance
  - Added .toString()
  - Added .translate(Number[, Number]);
- Added /tests/math.html which currently contains 67 unit tests for Matrix2D/Vector2D


louisstow:
The comments need to be slightly changed for the documentation generator. Needs an @ symbol after the first /**. The name needs a # in front of it (#.add). See the style guide: https://github.com/louisstow/Crafty/wiki/Style-Guide

Issue #127, opened 2011-09-25
Compare memory consumption on FF6 between 0.5 and 0.4.5


sorenbs:
See this discussion: https://groups.google.com/forum/#!topic/craftyjs/v_IKimZctQo

Currently the DOM implementation leaks memory in firefox. When the new drawing implementation from DEV branch is merged to master (scheduled for the 0.5 release) it should be tested if it fixes the memory leak.
If not decide what to do about it.

louisstow:
What's the latest on this issue? Does the latest snapshot help at all?

sorenbs:
This is still a problem in the snapshot. I have not tested the dev branch.
On 22 Oct 2011 17:39, "Louis Stowasser" <
reply@reply.github.com>
wrote:

sorenbs:
There might be some useful stats on about:memory

Also, this is an interesting read on the improvements coming in FF11 onwards memory-wise http://blog.mozilla.com/nnethercote/2011/11/01/spidermonkey-is-on-a-diet/

ahilles107:
This issue shoud be closed i think.

sorenbs:
I still think this is an issue that we should do something about. 
I have added a description to make this issue more actionable. Thoughts?

ahilles107:
we must have more data about this memory leak.

sorenbs:
Then i guess the first phase is to gather more data :-)

What i know is, 

1) when you run this program it eats infinite memory on FF (tested on FF 8) but not in chrome or IE

setInterval(function () {
        var a = Crafty.e("2D, DOM, Color").color("red").attr({w:20, h:20})
        .bind("EnterFrame", function () {
            this.x += 1;
            if (this.x > 400)
                this.destroy();
        });
    }, 1);

2) If you use Canvas instead of DOM there is no problem
3) Installing adblock plus makes firefox eat memory faster

This is a problem because many games (including the fruit assassin demo) creates and destroy a lot of entities.
My guess is that somehow we are not freeing DOM elements correctly, but i have not done any thorough analysis.

ahilles107:
On my ff 9 (ubuntu) it works without any leaks.

sorenbs:
Just tested fruit assassin on FF 11, windows. The memory buildup is much slower than on FF 8, but there is still a clear  pattern. mem usage increases slowly, then it drops suddenly (but not as much as it had increased) and the cycle is repeated.
After 20 minutes it cycles between 220 and 250 mb (it used 130 at startup).
[...]

PR #170, opened 2012-01-08
Memory Issues #127

BlackScorp:
With this small change i dont have Memory Issues on Ubuntu and Firefox 8, maybe iam Wrong can someone Test it?

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -684,7 +684,7 @@ Crafty.extend({
 		prev: (+new Date),
 		current: (+new Date),
 		fps: 0,
-
+                curTime:Date.now(),
 		init: function () {
 			var onFrame = window.requestAnimationFrame ||
 					window.webkitRequestAnimationFrame ||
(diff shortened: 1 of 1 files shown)


sorenbs:
closes #127
This seems to solve the problem on FF 11. But i find it a little concerning that the code you changed was not in the version of Crafty the bug was originally reported against :-)

I will merge and see if i can get the reporter to test it. Thank you!

BlackScorp:
sorry about that, but iam actually not good with github didnt knew that the issue was made on another crafty version

sorenbs:
I just meant that the problem was discovered on an older version of Crafty that did not have the part of the code that you changed, which means that the problem would have to be somewhere else, but perhaps we fixed the old problem and introduced this new one... That could explain it.

BlackScorp:
i just think that my crafty does not crash anymore in Firefox even if i have more than 300 entities in my scene, it lags but not crashes anymore

sorenbs:
That's great. We really should have some kind of automatic performance test that could run on each commit so we can keep track of how we are doing performance wise. It is all too easy to introduce problems like this one. I guess that will be my next  major Crafty related project.

PR #171, opened 2012-01-09
Support for other repositories

ahilles107:
Hi, im back ;)
I do this today. CraftyComponents.com will supported this change today.

### DIFF
--- a/crafty.js
+++ b/crafty.js
@@ -6,7 +6,7 @@
 * Dual licensed under the MIT or GPL licenses.
 */
 
-(function(window, undefined) {
+(function(window, undefined) {
 
 /**@
 * #Crafty
@@ -999,11 +999,11 @@ Crafty.bind("Load", function() {
 window.Crafty = Crafty;
 })(window);
 
-//wrap around components
+//wrap around components
 (function(Crafty, window, document) {
 
 
-/**
+/**
 * Spatial HashMap for broad phase collision
 *
 * @author Louis Stowasser
@@ -1162,7 +1162,7 @@ Entry.prototype = {
 parent.HashMap = HashMap;
 })(Crafty);
 
-Crafty.map = new Crafty.HashMap();
+Crafty.map = new Crafty.HashMap();
 var M = Math,
 	Mc = M.cos,
 	Ms = M.sin,
@@ -2061,7 +2061,7 @@ Crafty.matrix.prototype = {
 }
 
 
-/**@
+/**@
 * #Collision
 * @category 2D
 * Component to detect collision between any two convex polygons.
@@ -2317,7 +2317,7 @@ Crafty.c("Collision", {
 });
 
 
-/**@
+/**@
 * #DOM
 * @category Graphics
 * Draws entities as DOM nodes, specifically `<DIV>`s.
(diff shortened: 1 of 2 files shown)


sorenbs:
While i agree it should be easy to use an alternative repository i think we should keep crafty-modules.com as the default to use if no repository is specified.

crafty-modules.com is open source, open for commiters and intended to be as much a community effort as crafty itself. 

The way i see it, the place where support for alternative repositories can be a great benefit is if you are developing your game as closed source and want to keep all sources on your internal network.

I suppose we allow the following two ways of using Crafty.modules:

Crafty.modules("http://craftycomponents.com", { moveto: 'DEV' }, function () {
    //module is ready
    Crafty.e("MoveTo, 2D, DOM");
});

Crafty.modules({ moveto: 'DEV' }, function () {
    //module is ready
    Crafty.e("MoveTo, 2D, DOM");
});

ahilles107:
Sorry for second commit ;/ it's change nothing.

With crafty-modules.com is one main problem - not many people know scala ;/

CraftyComponents.com is writen in PHP (on Symfony2 (https://github.com/symfony/symfony), Twig (https://github.com/fabpot/Twig), and Doctrine (https://github.com/doctrine/doctrine2)) - its biggest comunity on github. 

CraftyComponents.com will be open sorce to. (i must create repo on github, and clear parameters.ini). I and others from FlyWithMonkey have a lot ideas for this page (ex. comments for components, contributors rankings etc.). And in scala we canot help ;/

Then why don't choose CraftyComponents.com (I know - it was a lot of code, but in our side too).

ahilles107:
https://github.com/ahilles107/CraftyComponents - CraftyComponents is open source now. Tomorrow i will write docs for instalation.

sorenbs:
If you and your company want to commit to supporting craftycomponents.com i think we should let that be the default. It's been 5 years since i have done any php, so it should be interesting to read through the code and see what has changed since then.

We would like to release a new version of Crafty soonish, and i would like the module system to be fully operational first. How much time do you think you need?

I have one feature request also: always lowercase the filenames, so http://cdn.craftycomponents.com/MoveTo-1.8.js -> http://cdn.craftycomponents.com/moveto-1.8.js

ahilles107:
Nice to hear this.

Time depends on the number of features. For now we want add cron for updating components one time each day, comments for components and added jsfiddle examples on component page. We also want to start work on Crafty Game Boilerplate - but it's not exactly related with craftycomponents.com

for now it should work for moveTo. Now I working with github parser for load crafty.js (paths in package.json files object)

sorenbs:
It would be nice if you could support github post-receive hooks to have components updated instantly. That produces a much better workflow when discussing and showing changes to somebody.

You should also consider serving the js files from something like s3 to get higher availability. As you are using a subdomain this can always be changed later.

ahilles107:
We can support post-recive hooks. We use rackspace nad we can create cdn on it.

sorenbs:
Sounds good!

Venlig Hilsen
Søren

2012/1/10 Paweł Mikołajczuk <
reply@reply.github.com

sorenbs:
You could consider using http://disqus.com/ for comments.

sorenbs:
When i try to update at
http://craftycomponents.com/components/single/8/crafty i get a 503 server
error.

Venlig Hilsen
Søren

PR #172, opened 2012-01-09
^M problem in stage.html

pengyu:
There is ^M in stage.html. I fixed it. And the format was dos. I changed it to unix.


ahilles107:
Maybe you can remove this PR and do it again (after second commit we don't see changes) ;/ Can You?

pengyu:
I'm not sure if I understand you.

Do you mean the current repository has it fixed?

I just pulled it. It is not fixed.

Or you mean there is not change in the "second commit" ("79ad13d"). In that case, there is not much difference between 4388c13 and 79ad13d. I just change the format from dos to unix (can be seen in vim). I think that only the lineends have been changed, so you will not see the diff between the two commits.

But since other files are in unix format, I think that it is better to keep files in unix format.

Please let me know if this make sense to you.

sorenbs:
What is ^M?

pengyu:
It is related with line end. See here. In some editor, you may not see it. But in vim, you can see it.

http://vim.wikia.com/wiki/File_format

sorenbs:
OK, i merged your pull request. Hope it fixed it.

Venlig Hilsen
Sren

On Tue, Jan 10, 2012 at 1:01 AM, pengyu <
reply@reply.github.com

jraicr:
The "Carriage Return" (CR) show the characters '^M'  at the end of every line if you are using vim.  I think this is what he means. It's a bit uncomfortable for us.  You can read from http://vim.wikia.com/wiki/File_format the next:  

(...) A common problem is that you open a file and see ^M at the end of many lines. Entering :set ff? will probably show that the file was read as unix: the problem is that some lines actually end with CRLF while others end with LF. To fix this, you need to tell Vim to read the file again using dos file format. When reading as dos, all CRLF line endings, and all LF-only line endings, are removed. Then you need to change the file format for the buffer and save the file (...)

sorenbs:
I guess the problem can be alleviated by doing 
git config --global core.autocrlf true
as per http://help.github.com/line-endings/

But all files in the repo should be with unix style line endings, so i hope you will fix it if you see others.

PR #176, opened 2012-01-11
Speed for X and Y Axis

BlackScorp:
Added the Posibility to setup the movement speed for x and y axis. Needed for example for Isometric Map to move player correctly diagonal


BlackScorp:
https://github.com/louisstow/Crafty/pull/177 see here

PR #181, opened 2012-01-16
On destroy of 2D entity, destroy all children

irongaze:
Louis/team, added code to automatically destroy 2D's children when parent is destroyed.  Seems like the right thing to do, fixes some odd behavior.  5 lines changed... let me know if this works for you, or if there's another method you'd like me to pursue for contributing to Crafty.

Thanks!

### DIFF
--- a/src/2D.js
+++ b/src/2D.js
@@ -212,8 +212,15 @@ Crafty.c("2D", {
 			this._cascade(e);
 		});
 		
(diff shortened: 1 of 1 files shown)


sorenbs:
Awesome. This is the way to contribute. 
If you continue to submit good pull requests Louis will probably grant you the commit bit. You have been warned :-)

PR #186, opened 2012-01-20
Uninitialize scene functions.

harrislapiroff:
Added a parameter to Crafty.scene which allows developers to specify a function that will run when a scene uninitializes. This gives developers the option to define an uninitialize function for each scene which gets called in between 2D components being destroyed and the next scene being initialized.

This should not create any backwards incompatibilities. If you only define scenes with one function, the API is exactly the same.


sorenbs:
Thanks! I have merged manually. Please check that i got it all right.

harrislapiroff:
This all looks correct. The generated code is a big part of what's making merges difficult. We should talk about how we handle this.

harrislapiroff:
Also, in the future, here's the ideal way of making these merges, which will retain the commit history.

In your git repository execute these commands:
- `git checkout master` Make sure you're on the master branch. Also make sure you're up to date by pulling from louisstow/Crafty.
- `git checkout -b lapilofu-merge` Create a branch to temporarily work on the merge. Name it whatever you want.
- `git pull https://github.com/lapilofu/Crafty.git scene-uninit` Change these to the address of the repository and the branch name you are trying to merge.
- It will give you a list of conflicting files. Go into these files one by one and you'll see the conflicts are marked off in the text. Git does that for you. Figure out manually how to resolve them. Then use `git add <path_to_file>` to mark each file as resolved. If they're generated files, like crafty.js or crafty-min.js, resolve conflicts by regenerating them after you've resolved all other files. Though I see these files have now been removed from the repository.
- `git commit` will automatically commit the merge, and even provide you with a useful informative merge message.
- `git checkout master`
- `git merge lapilofu-merge`
- `git branch -d lapilofu merge`
- Then you can `git push` as you well know. :)

PR #192, opened 2012-01-31
Isometric demo dev

pengyu:
Changed a few minor things. Remove dependency on jQuery (it is not necessary to have jQuery in order to demonstrate crafty). Right click doesn't work on Mac OS. Fall back to any mouse click.

### DIFF
--- a/demos/isometric/game.js
+++ b/demos/isometric/game.js
@@ -1,5 +1,5 @@
 //pro tip: see also this work in progress by Hex http://jsfiddle.net/hexaust/HV4TX/
-$(document).ready(function() {
+window.onload = function() {
 	Crafty.init();
 
 	Crafty.sprite(128, "images/sprite.png", {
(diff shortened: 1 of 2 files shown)


sorenbs:
I think you should be able to handle right click using the new Crafty.mouseButtons.

pengyu:
Crafty.mouseButtons doesn't work on Mac OS (in chrome), either. See the test code below.

<!DOCTYPE html5>
<html>
  <head>
    <style>
      body, html { margin:0; padding: 0; overflow:hidden; font-family:Arial; font-size:20px }
      #cr-stage { border:2px solid black; margin:5px auto;}
    </style>
  </head>
  <body>
    <script type="text/javascript" src="http://cdn.crafty-modules.com/crafty-DEBUG.js"></script>
    <script>
      window.onload = function () {
        Crafty.init(400, 336);

```
    Crafty.scene("main", function () {
      Crafty.e("2D, DOM, Color, Mouse")
      .attr({x: 100, y: 100, w: 150, h: 200})
      .color("red")
      .bind('Click', function(e) {
        if(e.mouseButton == Crafty.mouseButtons.RIGHT)
        console.log("Clicked right button");
      })
      ;
    });

    Crafty.scene("main");
  };
</script>
```

  </body>
</html>

PR #195, opened 2012-02-01
Added getId

pengyu:


BlackScorp:
var test = Crafty.e("2D,Canvas, test");
console.log(test[0]) and you have the id then

pengyu:
I should have used getDomId instead of getId.

See discussion here.

http://groups.google.com/group/craftyjs/browse_thread/thread/9240f538125c55f2

BlackScorp:
http://jsfiddle.net/X6Urq/  return this._elements.id give you the same value like Crafty(1);

pengyu:
_elements.id is less convenient than getDomId (which is actually implemented by _elements.id). Also _elements should be an internal variable that users should not call directly.

See the section on Visibility here.

https://github.com/louisstow/Crafty/wiki/Style-Guide

BlackScorp:
i just wanted to say, that getDomID method is not needed since you can select an entity by ID this way

var ent = Crafty(2); //Select the entity with ID 2

//display ID
alert(ent[0]); 

//display ID for jQuery selector
alert("ent"+ent[0]);

pengyu:
I didn't know Crafty(2) before.

I can not forseen all the cases where having DOM ID might be useful.

But there might be cases you want to use third party selector (say jQuery), in which case getDomId will be helpful?

Or probably you want to have some debugging information, in which case you want to print DOM ID?

Anyway, having a convenient function doesn't hurt anything, right?

BlackScorp:
var ent = Crafty(2);
$('#ent'+ent[0]).text("test");

thats all, another function will hurt the javascript filesize so the pageload will be slower

harrislapiroff:
This looks great to me. The function is small, and I think that entitity.getDomId() is much more obvious than entity[0].

BlackScorp:
but dom id has a prefix "ent" and not just a number

harrislapiroff:
I've merged this into the 'develop' branch.

PR #196, opened 2012-02-01
Added DraggableHorizontal and DraggableVertical

pengyu:


BlackScorp:
why just not use draggable but ignore mouse.x or mouse.y ?

harrislapiroff:
It does seem like expanding Draggable to take options, rather than introducing two new components that are nearly identical, might keep the code a little DRYer.

Thoughts?

BlackScorp:
that i wanted to say, just Crafty.c("Draggable",{
_axis:null, //setting draggable axis attribute to "v","h" or null to make draggable vertical/horizontal or both
init:function(){
//some stuffs
}
});

somethink similar to this

pengyu:
This will result in several if-statements, which will make code hard to read and change. In general, I think that it is better to avoid if statement if possible. Many C++ template based libraries go very far in terms of eliminating if-statement (as the logic over there is that if statement will cause additional run time as well as readablity).

For JavaScript, the runtime is not that important but for the sake of readablity, it is still worthwhile to eliminate if-statements.

BlackScorp:
with simple maths you can remove the if statments like 
this._axis = {
h:1,
v:1
}

this.x = this._axis.h \* (pos.x - this._startX);
this.y = this._axis.v \* (pos.y - this._startY); //0 \* x is 0

harrislapiroff:
I think in this case, the if statements could be made very obvious and would be worth the trade off in terms of not having three sets of code that are identical except for two lines. In my ideal world, creating a horizontal draggable would work something like this:

```
drag_entity = Crafty.e("2D, Draggable")
drag_entity.draggable(true, false)
```

The if statements in the component would be simple:

```
this._ondrag = function (e) {
    var pos = Crafty.DOM.translate(e.clientX, e.clientY);

    if this._drag_x {
        this.x = pos.x - this._startX;
    }

    if this._drag_y {
        this.y = pos.y - this._startY;
    }

    this.trigger("Dragging", e);
};
```

Seems pretty legible to me.

BlackScorp's approach above also seems legitimate, though, in my opinion a little, a little less legible. Maybe if you called the variable `._drag_multiplier` instead or something.

ahilles107:
Maybe it must be only a component loaded to craftycomponent.com? What are you thing guys?

pengyu:
This is wrong.

this.y = this._axis.v \* (pos.y - this._startY); //0 \* x is 0

When this._axis.v is 0, this.y should not be changed.

harrislapiroff:
Ah, right you are. I only skimmed that solution to get the basic idea. The correct line would probably be something more like:

```
this.y = this.y * (1 - this._axis.v) + this._axis.v * (pos.y - this._startY); // assuming ._axis.v is 0 or 1
```

Makes it less good of a solution, I suppose. I still kinda like mine though.

harrislapiroff:
We could also use a ternary operator, with my approach, which would be compact and legible:

```
this.x = this._drag_x ? pos.x - this._startX : this.x
```

pengyu:
I'm thinking of making draggable allowing arbitrary direction (not just horizontal and vertical).

The ideal solution is to record the position of the mouse when it first pressed (call it O1, the entity is at A). Then move the entity by check the current position of the mouse (call it O2). Then the new position of the entity is A+(O2-O1)project-on-to(v,h), where (O2-O1)project-on-to(v,h) means project the vector O2-O1 onto the vector defined by (v,h).

The coordinate system in crafty is still quite confusing to me. Does anybody know what the code should be for the algorithm that I propose to use?

BlackScorp:
@pengyu i know that my function above was wrong, i just wanted to post that there are more ways just if else.

pengyu:
Removed DraggableHorizontal and DraggableVertical. Modified Draggable to allow drag in any direction.

Zoom and scaling with different ratios in x and y axis are not tested yet.

src/controls.js
```diff
@@ -235,20 +235,30 @@ Crafty.c("Mouse", {
 * @trigger StopDrag - is triggered when dragging ends - MouseEvent
 */
 Crafty.c("Draggable", {
-	_startX: 0,
-	_startY: 0,
+  _origMouseDOMPos: null,
```
BlackScorp, line 238: DOMPos? draggable should work with Canvas as well as DOM

src/controls.js
```diff
@@ -235,20 +235,30 @@ Crafty.c("Mouse", {
...
+	_oldX: null,
```
BlackScorp, line 239: why? _startX mean the X position where the dragging start

src/controls.js
```diff
@@ -277,6 +288,46 @@ Crafty.c("Draggable", {
...
+  * // Note: because of the orientation of x and y axis,
+  * // this is 45 degree clockwise with respect to the x axis.
+	* this.dragDirection({x:1, y:1}) //45 degree.
+	* this.dragDirection(60) //60 degree.
+	* ~~~
+	*/
+	dragDirection: function(dir) {
+		if (typeof dir === 'undefined') {
+			this._dir=null;
+		} else if (("" + parseInt(dir)) == dir) { //dir is a number
```
BlackScorp, line 315: typeof dir == "number" is the JS way to check the type

src/controls.js
```diff
@@ -277,6 +288,46 @@ Crafty.c("Draggable", {
...
+      this._dir={
+        x: Math.cos(dir/180*Math.PI)
```
BlackScorp, line 317: why this? dragging , means drag an element on axis

src/controls.js
```diff
@@ -277,6 +288,46 @@ Crafty.c("Draggable", {
...
+	* Specify the dragging direction.
+	* @example
+	* ~~~
+	* this.dragDirection()
+	* this.dragDirection({x:1, y:0}) //Horizonatal
+	* this.dragDirection({x:0, y:1}) //Vertical
+  * // Note: because of the orientation of x and y axis,
+  * // this is 45 degree clockwise with respect to the x axis.
+	* this.dragDirection({x:1, y:1}) //45 degree.
+	* this.dragDirection(60) //60 degree.
```
BlackScorp, line 309: if you add degree it called rotation

harrislapiroff, line 309: I don't think rotation is quite right. This doesn't rotate the entity, it rotates its direction of draggability.

BlackScorp:
i wish he would make a demo at jsdiffle to see what is the idea behind this

src/controls.js
pengyu, line 238: The original code uses Crafty.DOM.translate(e.clientX, e.clientY). So I name the variable with DOM.

But I tested. The code works for DOM as well.

pengyu:
I personally do not like jsfiddle. It would be more convenient to put the testing code in the repository rather than somewhere else. I have added a few examples in my branch draggable_horizontal_and_vertical_dev.

You can pull it and look at the examples in

testing/Crafty/Input/Draggable/

BlackScorp:
ah i see now, its horizontal dragging. why dont you just add axis function to Draggable component. so i think it would be more clear to do somethink like this

Crafty.e("2D,DOM,Draggable").axis(90); //Drag Horizontal only
Crafty.e("2D,DOM,Draggable").axis(180); //Drag Vertical only
Crafty.e("2D,DOM,Draggable").axis(45); //Drag Diagonal only
Crafty.e("2D,DOM,Draggable"); //Drag All directions

its just, to remove StartX is not Compatible, so ppls with old code would get errors. Try not to touch existing methods or variables

pengyu:
dragDirection is clearer than axis. The keyword "drag" tells that it is related with Draggable. User need not to look at the document to figure this out.

"_startX" was an internal variable, it did not appear anywhere else in crafty.js besides in Draggable. Also, I think "_oldX" is a better name than "_startX" ("start" could mean the very beginning, where "old" means the value when the mouse is clicked).

BlackScorp:
you can edit internal variables with .attr() so it is possible that some games read the values or set them.
axis is used by jQuery UI Draggable http://jqueryui.com/demos/draggable/#option-axis just think that association with existing frameworks could help new developers.

pengyu:
The jQuery doc says "Constrains dragging to either the horizontal (x) or vertical (y) axis. Possible values: 'x', 'y'."

It is clear that jQuery only allows x and y axis. My version of Draggable can be dragged in any direction. Using the word "axis" can not cover directions that are not along x and y axis.

BlackScorp:
the current version of Draggable can drag anyways in any direction, your version rotate the axis and fix the object at that axis you could add as description then "Constrains dragging to axis. Possible values: Number Degree."

Then this example

Crafty.e("2D,DOM,Draggable").axis(90); //Drag Horizontal only
Crafty.e("2D,DOM,Draggable").axis(180); //Drag Vertical only
Crafty.e("2D,DOM,Draggable").axis(45); //Drag Diagonal only
Crafty.e("2D,DOM,Draggable").axis(90).axis(180).axis(45); //Drag Horizontal/Diagonal/Vertical only
Crafty.e("2D,DOM,Draggable"); //Drag All directions

and why noone else is discussing here? Oo

pengyu:
English, you say

"I drag an object along the x axis."

"I drag an object in the 45 degree direction with respect to the x axis."

I don't think that it is necessary to think of rotating axis. It is awkward to say

"I drag an object along the x axis rotated by 45 degree."

Since the drag can be constrained in any direction. "Direction" should be a better keyword than "axis".

Chaining multiple dragDirection seem to be useful to support multiple directions.

Crafty.e("2D,DOM,Draggable").dragDirection(90).dragDirection(180).dragDirection(45); //Drag Horizontal/Diagonal/Vertical only

BlackScorp:
ok then call it directionDrag() since all draggable methods ends with Drag http://craftyjs.com/api/Draggable.html

pengyu:
I think that .stopDrag means "stop dragging". So the word order means something.

In English, you say "dragging direction". So the function name should be dragDirection (to maintain the order in English) rather than directionDrag.

BlackScorp:
do what you prefer;) but if some ppls got trouble with it and ask for help in forums, you have to help them out;)

pengyu:
Sure.

harrislapiroff:
It has been famously said that there are only two hard problems in computer science: cache invalidation and naming things.

harrislapiroff:
Looks great. I've merged it into the new 'develop' branch.

PR #199, opened 2012-02-02
Update src/core.js

BlackScorp:
Quick function to toggle Components

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -208,7 +208,26 @@
 			this.trigger("NewComponent", ul);
 			return this;
 		},
-
+		/**@
+	* @comp Crafty Core
+	* @sign public this. toggleComponent(String componentID,String componentToggle)
+	* @param componentID - Component ID to add or remove.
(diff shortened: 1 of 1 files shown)


harrislapiroff:
I don't think this will work quite as expected. If I do

```
var e = Crafty.e("2D,DOM,Test");
e.toggleComponent("DOM","Canvas");
```

Then I will expect `e` to have components `"2D, Canvas, Test"` but it will only have components `"2D, Test"`, unless I am misunderstanding the code. I would expect `.toggleComponent("DOM", "Canvas")` to switch back and forth between the `DOM` and `Canvas` components. Am I misunderstanding the intention?

harrislapiroff:
Additionally, this is causing a syntax error when I test it. I believe you need a comma on line 230.

BlackScorp:
ah sorry i will edit it, but i did expect that it will change 2D,DOM,Test to 2D,Canvas,Test since i just using craftys add and remove components. ill test it tomorow and add fully working pull request

PR #202, opened 2012-02-03
Update src/DOM.js

BlackScorp:
Dont Create new DOM element if its already exists

toggleComponent("DOM") creates new DOM element instead of Editing the existing ones

### DIFF
--- a/src/DOM.js
+++ b/src/DOM.js
@@ -12,11 +12,13 @@ Crafty.c("DOM", {
 	_element: null,
 
 	init: function () {
-		this._element = document.createElement("div");
-		Crafty.stage.inner.appendChild(this._element);
-		this._element.style.position = "absolute";
-		this._element.id = "ent" + this[0];
-
+		this._element = document.getElementById("ent" + this[0]);
+                if(!this._element){ //Create new DOM element if it does not exists
+			this._element = document.createElement("div");
+			Crafty.stage.inner.appendChild(this._element);
+			this._element.style.position = "absolute";
+			this._element.id = "ent" + this[0];
+                }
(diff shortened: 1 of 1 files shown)


sorenbs:
Would this allow for a situation where an entity is being associated with a dom element that has some old css styles that does not apply to the new entity?

BlackScorp:
uhm i actually dont know what you mean, can you explain more please?

sorenbs:
I just thought that when reusing dom elements, they might have some old css properties. For example, if you create a dom entity with a sprite, removes the dom component, removes the sprite and adds the dom component again the dom element would still have the background image set.
I haven't tested if that is a real problem or just something i'm making up though :-)

BlackScorp:
hm.. that could be, in this case Crafty components need an uninitial event or function which is called on removing the component, then removing the DOM object completly on removing DOM component would allow to swich DOM to Canvas in runtime.

what do you think? add an destructor to core.js in removeComponent function then we can add uninit functions to canvas and dom and stuff

sorenbs:
Yeah, i think adding an uninit function to components would be a neat general solution.

sorenbs:
Okay, i have added the uninit call and corresponding function in DOM. From my brief testing this enables the DOM/Canvas switching. Pretty cool.

https://github.com/louisstow/Crafty/commit/bcc789c30467049f9073a38834084e486730e826

BlackScorp:
can you post a jsfiddle link? i want to see it :D

sorenbs:
hah :-)

The autogenerated snapshot does not pickup changes in the develop branch. We will have to find a solution for that to make it feasible to use jsfiddle again.

PR #204, opened 2012-02-04
Export 2d doc to html

pengyu:
Some attributes of 2D do not appear in the html document, although they are in the comments. Fix the problem, so that they appear in the html document.
# 

Oops. I should have requested louisstow:develop. Is there a way to change the default pull request to  louisstow:develop instead of  louisstow:master?


sorenbs:
Merged to develop

PR #216, opened 2012-02-19
make .color() act as a getter

jayvan:


ahilles107:
Can You add example for usage in docs?

Issue #23, opened 2011-01-22
Mobile devices


louisstow:
Haven't been able to test Crafty on mobile devices but some components to accept mobile inputs would be handy.

ponychicken:
Tried my game in progress on an iphone. It's seriously slow, but everything seems to work. Setting up touchzones is 20 lines or less.

louisstow:
Damn. Is there a way to profile? And do you know which areas lagged?

ponychicken:
Profile: not sure. I'm downloading the ios simulator now, maybe it can shed some light.

ponychicken:
This is nice:
http://stevesouders.com/mobileperf/mobileperfbkm.php

louisstow:
Don't know how possible it will be in pure JS. It's pretty slow on my android.

BossRighteous:
I have a cross device click/touch solution that's pretty easily implemented elsewhere. I'll make it a component, I just need to see how the current custom click events bubble up in Crafty.
An example can be found at http://www.fjhorton.com/spinner.html
This works on Desktop, iOs, and Andriod. The only thing it doesn't include is multi-touch support. But that's also implemented somewhat easily.
An important note is that the move events fire much too fast to deal with directly. The way I handled it was to simply note the last known mouse position and deal with the resulting visual updates in an enterFrame style interval. This is also required because a touchend event doesn't fire a position coord. but this is pretty useful information.

One of the biggest issues with Android devices I have noticed in my own adventures is the slow speed of DOM manipulations. Transitioning to xTransform CSS translations should speed things up fairly nice.
Another speed issue is how the page re-paints and reflows on a css property change. I took a look at the timer loop and I can optimize it a bit I think.
Imagine 20 items, each with a x and y update. Currently, each item processes the left and top properties separately (rightly so.) The issue is this triggers a reflow and repaint of the screen. So 20x2 means that in one update loop the stage is re-adjusted 40 times.
The simple solution is to remove the stage child from the DOM parent right before looping through the redraws, and when the redraw routine completes, simply add the stage back to the parent. That ensures that only one reflow/paint is performed per tick. This would create an issue only where the stage is a non [0] child of its parent. Properly denoting the first-child requirement shouldn't cause many further problems.

Another optimization that I feel is helpful, and certainly in a mobile environment, would be a deltaTime based tick structure.
[...]

PR #223, opened 2012-02-25
Touch events

smindel:
Fix for broken TouchEvents (#23): Don't handle TouchEvents in mouseDispatch but in the new touchDispatch which dispatches a mocked MouseEvent for normalization. Tested with tests/touchevents.html.

P.S.: Kudos for your great work. Just working on a Breakout clone to get my hands dirty. Works like a charm.

### DIFF
--- a/src/controls.js
+++ b/src/controls.js
@@ -25,6 +25,7 @@ Crafty.extend({
 
 	mouseDispatch: function (e) {
 		if (!Crafty.mouseObjs) return;
+		if(window.TouchEvent && e instanceof TouchEvent) { this._mockMouseEvent(e); return; }
 		Crafty.lastEvent = e;
 
 		var maxz = -1,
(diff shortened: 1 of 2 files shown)


ahilles107:
How about safari (ipad) 300ms touch lag?

smindel:
The patch is using TouchStart which fires immediately, good for smooth and responsive dragging of entities around, no delay. A Click is not a TouchEvent but a MouseEvent and that's why it is not affected by the TouchEvent2MouseEvent translation. It gets fired as a legacy event still and it is delayed on touch devices. In mobile projects you would either live with the convenient but delayed legacy ClickEvent or bind to MouseDown/MouseUp which now works equally fast on mouse AND touch devices.

To cut a long story short: click was and still is delayed. With the patch you get a speedy alternative and access to the native TouchEvents which was broken.

PR #224, opened 2012-02-25
Fixes

smindel:
When you remove ANY component form a DOM entity it gets undrawn, resulting in an error when you try to change the appearance of the entity later.


sorenbs:
Good catch. I fixed the problem in this commit already https://github.com/craftyjs/Crafty/commit/7f1295262a0daa0fe11ad802a13bb4b269fe78d0

smindel:
sweet as, cheers

PR #226, opened 2012-02-29
Isometric refactor

ajacksified:
Update to pull request #111 involving @techwraith and @louisstow. Rewrote Crafty.isometric as a module and updated the demo; added a diamond-based isometric module and a new demo to go along with it.


BlackScorp:
your demo link from pull request #111 does not work, maybe you can update it? want to see what did you changed

ajacksified:
@BlackScorp Not sure what state pull #111 is in; the demo here should work. It's pretty much the same code, except a module instead of a rewrite of the isometric component. Does the demo here work for you? (You probably have to rebuild crafty.)

BlackScorp:
@ajacksified i mean that link http://olivinelabs.com/labs/crafty/iso.html it isnt working for me

ajacksified:
Oh, I see. I uploaded the new demo here: http://olivinelabs.com/demos/crafty/diamond-isometric/index.html

BlackScorp:
ok i see the code, but it wont work for other isometric formats like 4:3 or 16:9 it works only if tiles are 2:1 , maybe you can rewrite the code? also this placing on layer seems to be interessting let me know if you edited your demo with sizes, lets say 128x96 (4:3). 
but i still think isometric have to be an extension to crafty , not component, beause it has helper function to place components.

ajacksified:
@BlackScorp That is correct - it assumes 2:1. 

As for leaving it as an extension: check out the discussion with @sorenbs at https://github.com/craftyjs/Crafty/pull/111#issuecomment-4247889 and see what you think.

BlackScorp:
you know , your implementation is that what iam looking for the whole time, i dont want to use place(x,y,z,obj) it could be better to use place(x,y,layer,obj) but your current code assume 2:1 and it is not able to handle nagative x/y values. 

if you could implement this in the current Crafty.isometric that would be really awesome. 

but i dont like the idea using isometric as component, because isometric is like Crafty.math which just has some helper functions. hope you will modifie your code a bit;)

ahilles107:
@louisstow Can You look at this?

mpetrovic:
Gonna have to agree with BlackScorp here. I don't care for the Isometric as a component either. I think its best left as a helper to make placing them less mathy.

I don't agree that place(x,y,layer,obj) is the way to go. Layers imo is for grouping objects that exist in the same 'space'. 

In any case, I'm rewriting a lot of code that intersects with this, so I won't be merging. Keep track of the new work being done in this space in the layers branch, and when camera are ready to have their rendering functions implemented, feel free to jump in.

ajacksified:
@mpetrovic - thanks. I'll keep an eye on the layer update.

PR #227, opened 2012-03-01
Patch 16

BlackScorp:
Fixed Crafty.loader skip loading on not supported audio files. call different event on audio rather than image

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -59,49 +59,58 @@ Crafty.extend({
 	* @see Crafty.assets
 	*/
 	load: function (data, oncomplete, onprogress, onerror) {
-		var i, l = data.length, current, obj, total = l, j = 0, ext;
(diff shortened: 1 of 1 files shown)


src/loader.js
```diff
@@ -59,49 +59,58 @@ Crafty.extend({
...
-			//if there is an error, pass it in the callback (this will be the object that didn't load)
-			obj.onerror = function () {
-				if (onerror) {
-					onerror.call(this, { loaded: j, total: total, percent: (j / total * 100) });
+                            },false);
+                        obj.addEventListener('error',function(){
+                            j--;
+   
+                            if (onerror) {
+					onerror.call(this, {loaded: j, total: total, percent: (j / total * 100),src:this.src});
```
BlackScorp, line 101: you also can see which file could not be loaded or has been loaded

BlackScorp:
addEventListener is not supported by IE

Issue #228, opened 2012-03-01
Starting animation from AnimationEnd


sorenbs:
It should be possible to start an animation from a function bound to AnimationEnd

```
    player = Crafty.e("2D, Canvas, animator, SpriteAnimation")
        .attr({x: 64, y: 64, z: 2})
        .animate("level1", 0, 0, 5)
        .animate("level2", 0, 1, 5)
        .animate("level1", 30, 2)
    player.bind("AnimationEnd", 
        function()
        {
            this.animate("level1",30, 5);
            console.log("AnimationEnd");
        });
```

http://jsfiddle.net/KD8q9/24/

ahilles107:
this still don't work. I works on this but without results. But i fixed #275. Will be cool if anyone can look at this.

BlackScorp:
i think we need AnimationEnd and AnimationStop events AnimationEnd will be triggered after each Reel animation Stop after animation ends

ahilles107:
sounds good for me!

starwed:
Should be fixed by #426

PR #229, opened 2012-03-01
Update src/animation.js

BlackScorp:
Trigger AnimationStop after stoping the animation think issue #228 will be fixed


BlackScorp:
tested it, this.stop unbind AnimationEnd so it wont be triggered..

PR #231, opened 2012-03-07
Update src/core.js

BlackScorp:
enabled console.log in opera

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1073,7 +1073,12 @@
 
 		clone: clone
 	});
-
+/**
+* console.log for Opera
+*/
+ 	if(window.opera){
+                console = {log:opera.postError}
+         }
 	/**
 * Return a unique ID
 */


ahilles107:
Why it shoud be in default Crafty core?

mpetrovic:
I agree, this shouldn't be in Crafty. If you really want to debug in Opera, use include Firebug lite in your testing pages.

I'm actually surprised it doesn't have an implementation of console on its own.

BlackScorp:
its just, on Craftys webpage it says support all browsers, so if anyone is not familar with Opera he would wonder first time why console.log does not work for him on Opera. with this code i just thought it would be easier for them, they dont need to install external stuffs.

i think the best way would be a Crafty debugger like imactjs have one

mpetrovic:
console is not a Crafty feature. The games themselves should work in every
browser, but I don't see any obligation to patch holes in their development
tools.

On Sat, Mar 10, 2012 at 6:46 AM, Witali Mik <
reply@reply.github.com

BlackScorp:
understand... youre right

PR #236, opened 2012-03-13
Properly wrap from the last frame to first frame in SpriteAnimation

### LINKED ISSUE #233: Last frame of SpriteAnimation lasts twice as others
It can be clearly noticed in demos using animations (RPG, Defend the Den). There is also a discussion of this in the forums: https://groups.google.com/forum/#!topic/craftyjs/6QM-o8NDM-g

jayvan:
fixes #233
https://github.com/craftyjs/Crafty/issues/233

### DIFF
--- a/src/animation.js
+++ b/src/animation.js
@@ -129,19 +129,18 @@ Crafty.c("SpriteAnimation", {
 			this.__coord[0] = pos[0];
 			this.__coord[1] = pos[1];
 			this._frame.current = 0;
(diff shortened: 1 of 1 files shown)


jayvan:
Just saw #235. At a glance it looks like just variable names were re-factored but the final frame will still repeat. This branch will conflict with that one so after #235 is merged in I'll branch off and redo this logic fix.

PR #240, opened 2012-03-21
Update src/sound.js

BlackScorp:
Fixed sounds, added volume parameter to play function. 

Can be tested on http://blackscorp.github.com/cron/ channels coming soon

### DIFF
--- a/src/sound.js
+++ b/src/sound.js
@@ -1,298 +1,258 @@
 Crafty.extend({
-/**@
-	* #Crafty.audio
-	* @category Audio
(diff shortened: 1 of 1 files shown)


mpetrovic:
Looks good, but can you mark the members and methods you intend to be private by sticking an underscore at the front of their name? Makes it easier to distinguish them from the rest. 

Or even better, putting them in a closure so they're truly private.

BlackScorp:
will do if i implement channels

PR #241, opened 2012-03-24
Updated src/animation.js

NyanLeek:
Fixed animation for padded sprites, now the coordinates are calculated correctly using own's element padding.

To solve this:
https://groups.google.com/forum/#!topic/craftyjs/SIG1cJXWW1U/discussion

### DIFF
--- a/src/animation.js
+++ b/src/animation.js
@@ -106,8 +106,8 @@ Crafty.c("SpriteAnimation", {
 	  // .animate('PlayerRunning', 0, 0, 3) //setup animation
 		if (typeof fromx === "number") {
       // Defind in Sprite component.
-			tile = this.__tile;
-			tileh = this.__tileh;
+			tile = this.__tile  + parseInt(this.__padding[0] || 0, 10);
+			tileh = this.__tileh + parseInt(this.__padding[1] || 0, 10);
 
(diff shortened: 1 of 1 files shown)


src/animation.js
```diff
@@ -106,8 +106,8 @@ Crafty.c("SpriteAnimation", {
 	  // .animate('PlayerRunning', 0, 0, 3) //setup animation
 		if (typeof fromx === "number") {
       // Defind in Sprite component.
-			tile = this.__tile;
-			tileh = this.__tileh;
+			tile = this.__tile  + parseInt(this.__padding[0] || 0, 10);
+			tileh = this.__tileh + parseInt(this.__padding[1] || 0, 10);
```
BlackScorp, line 110: wont this throw an error? 0, 10 is not an integer so if this.__padding[0] is not set it will not work i suppose

NyanLeek, line 110: Nope, 10 is the second argument to the parseInt function, the 'radix', please refer https://developer.mozilla.org/en/JavaScript/Reference/Global_Objects/parseInt
Also this.__padding is meant to always be set, if not specified by user it becomes 0, this is just a security measure to prevent errors should arise in the future.

Issue #243, opened 2012-03-26
.areaMap polygon overrides .collision(polygon)


NyanLeek:
If they are different methods used for different things (collisions and mouse events) they should have different polygons.

If you set a collision polygon, and then a different areamap for mouse events all the collisions occur within the areamap,
This is due to areaMap and collision storing its polygon in the same variable; this.map, and .isAt reading said map, however if i read the source correctly .isAt is only used for mouse events, so, a good solution would be setting areaMap to store its map in another variable and isAt reading accordingly, and defaulting to collision map if there's no areaMap map.

ahilles107:
This issiue is allredy fixed?

PR #244, opened 2012-03-26
Updated src/2D.js src/controls.js

NyanLeek:
.areaMap polygon overrides .collision(polygon)

issue #243

https://github.com/craftyjs/Crafty/issues/243

Added mapMouse as the new variable for isAt and areaMap to read and set values, respectively, making it possible to handle mouse events separate from collisions, defaults to collision .map if .mapMouse not present.

Patch does not work for canvas, it defaults to the collision polygon, as I don't understand how it gets the data.

### DIFF
--- a/src/2D.js
+++ b/src/2D.js
@@ -478,7 +478,9 @@ Crafty.c("2D", {
 	* an object can't be passed. The arguments require the x and y value
 	*/
 	isAt: function (x, y) {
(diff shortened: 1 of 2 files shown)


mpetrovic:
Please reroll against develop branch.

NyanLeek:
Ok, didn't think of that :)

PR #256, opened 2012-05-01
Added support for ZQSD keys in Fourway and Twoway components (for azerty...

StijnRuts:
... keyboard layouts)


ahilles107:
Anyone see somethin wrong in this PR? If not, then we can merge this.

sorenbs:
I agree with hugeen, but that must be a thing in the future :-)

PR #258, opened 2012-05-11
Update src/sound.js

BlackScorp:
fixed playback repeat on Chrome  here example http://jsfiddle.net/euamd/

### DIFF
--- a/src/sound.js
+++ b/src/sound.js
@@ -191,13 +191,13 @@ Crafty.extend({
             if(s.obj.currentTime) s.obj.currentTime = 0;   
             s.obj.play(); 
             s.played ++;
-            s.obj.onended = function(){
+            s.obj.addEventListener("ended", function(){
                 if(s.played < repeat || repeat == -1){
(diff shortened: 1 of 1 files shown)


src/sound.js
```diff
@@ -191,13 +191,13 @@ Crafty.extend({
             if(s.obj.currentTime) s.obj.currentTime = 0;   
             s.obj.play(); 
             s.played ++;
-            s.obj.onended = function(){
+            s.obj.addEventListener("ended", function(){
```
ahilles107, line 194: i think that will don't work on ie. 

we must use here:

Crafty.addEvent(this, s.obj, 'ended', function(){

BlackScorp, line 194: it will, addEventListener is added since IE9 also audio is since IE9 is supported
http://www.w3schools.com/html5/tag_audio.asp
http://msdn.microsoft.com/en-us/library/ie/ff975245%28v=vs.85%29.aspx

ahilles107, line 194: yep, you're right.

PR #278, opened 2012-05-28
Scale

radzikowski:
Correct zoom operation (remove resize operation canvas and viewport) added separate function for scale to scale all viewport.

Canvas scale is not correct with not integer value (have to do some changes with dirty rectangles)

### DIFF
--- a/src/extensions.js
+++ b/src/extensions.js
@@ -571,6 +571,7 @@ Crafty.extend({
 						height: act.height * zoom
 					};
 					zoom += zoom_tick;
+					this._zoom = zoom;
 					var new_s = {
 						width: act.width * zoom,
 						height: act.height * zoom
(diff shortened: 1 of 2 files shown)


src/extensions.js
```diff
@@ -620,7 +612,45 @@ Crafty.extend({
...
 		})(),
+		/**@
+		 * #Crafty.viewport.scale
+		 * @comp Crafty.viewport
+		 * @sign public void Crafty.viewport.scale(Number amt)
+		 * @param Number amt - amount to zoom/scale in on the element on the viewport by (eg. 2, 4, 0.5)
+		 *
+		 * Zooms/scale the camera. amt > 1 increase scale 
+		 * amt < 1 will scale to small. amt = 0 will reset the zoom/scale.
+		 * Zooming/scaling is multiplicative. To reset the zoom/scale amount, pass 0.
```
ahilles107, line 623: Can you add example?

sorenbs:
nice work

PR #288, opened 2012-06-06
Correct sound mute and unmute opration

radzikowski:
Add "hide/private" function _mute, and function mute, unmute and toggleMute to api.


radzikowski:
Wrong branch.

PR #302, opened 2012-06-18
Sprite Margin

BlackScorp:
Please Accept only extensions.js since crafty sprite has padding, we also need margin for isometric renderer for example, maybe other kind of games need it too.


ahilles107:
can you close this PR and create new - clean?

PR #307, opened 2012-06-27
Separated CSS background assignment into indivudual properties

gvenkat:
The reason for doing this is .css( { 'background-size': 'whatever' } ) will not work as background assignment clobbers all other CSS properties


ahilles107:
Please send this PR to branch develop

PR #310, opened 2012-06-27
Bug Fixes for IE Events and Firefox rAF

austinh:
Fix Internet Explorer Event Propogation and preventDefault errors in KeyboardDispatch, as well as fix request AnimationFrame bug in Firefox when calling pause from inside EnterFrame event.

### DIFF
--- a/src/controls.js
+++ b/src/controls.js
@@ -243,9 +243,13 @@ Crafty.extend({
 		//Among others this prevent the arrow keys from scrolling the parent page
 		//of an iframe hosting the game
 		if(Crafty.selected && !(e.key == 8 || e.key >= 112 && e.key <= 135)) {
-			e.stopPropagation();
+
+            if(e.stopPropagation) e.stopPropagation();
+            else e.cancelBubble = true;
+
 			if(e.preventDefault) e.preventDefault();
 			else e.returnValue = false;
+
 			return false;
 		}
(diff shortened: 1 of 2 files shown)


austinh:
While pausing a game from inside an EnterFrame event, Firefox returns the following error:

NS_ERROR_XPC_BAD_CONVERT_JS: Component returned failure code: 0x80570009 (NS_ERROR_XPC_BAD_CONVERT_JS) [nsIDOMWindow.mozRequestAnimationFrame]

The following psuedo code gives this error.
Crafty.bind("EnterFrame", function()
{
if(GameOver == true)
{
Crafty.pause();
}
});

This is fixed by adding a 0ms setTimeout to the timer.stop() and timer.init() functions, which essentially means the pause/stop will not be run until after the JS Compiler has finished running the last EnterFrame function call.

When using the Keyboard dispatch I get the following errors in IE7 and 8:

Object doesn't support property or method 'preventDefault' in IE8.

Object doesn't support property or method 'stopPropogation'

I created a simple solution in a pull request.

ahilles107:
Please rebase your PR, close this and open new "clean" for branch develop.

sorenbs:
Great work. It is difficult to find your changes in the diff. Could you submit a clean patch against the develop branch?

PR #315, opened 2012-07-17
Fix the startDrag function

ABevier:
I fixed the startDrag function of the Draggable component so that it works when a drag is initiated programatically.  If the function is called, it will use the last known position of the mouse which is stored in Crafty.lastEvent.  I created a private function to share the common code used when starting a drag with a mouse event and starting a drag via startDrag.


ahilles107:
Nice work @ABevier - but this PR must go to develop branch. I will close this, and please open new correct.

PR #332, opened 2012-08-03
Generate list of events

jhcp:
- Generates events.html containing a list of all events triggered by Crafty components
- Create a link to events.html in the index


sorenbs:
Nice work!

PR #338, opened 2012-08-28
Added manual bounds to viewable area

MaxBareiss:
Added variable `bounds` in `Crafty.viewport` which  overrides `Crafty.map.boundaries()` when determining the size of the viewable area.


sorenbs:
Thanks! Could you add a small comment to the viewport documentation about this?

MaxBareiss:
Sure. I started pull request #358 to do that, but I'm not sure whether I added the documentation in the right place.

PR #343, opened 2012-09-08
Enhanced core API documentation

corellian:
I have improved some little chunks of the core API documentation. I hope it will be correct.


sorenbs:
Thank you very much. Your improvements are up on http://craftyengine.com/api/Crafty-Core.html#-removeComponent

PR #347, opened 2012-09-20
Fixes for indexedDB polyfills

talitapagani:
As Internet Explorer 10 will have the "msIndexedDB", I added this object to the polyfill and also made as a single-line statement.

But the main change was the polyfill for IDBTransaction. As IDBTransaction.READ/READ_WRITE and IDBTransaction.readonly/read_write are deprecated on new versions of FF and Chrome (supporting only the strings "readonly" or "read_write"), I added a polyfill to make it compatible with recent and legacy versions of browsers.

### DIFF
--- a/src/storage.js
+++ b/src/storage.js
@@ -135,7 +135,8 @@
 	 */
 
 Crafty.storage = (function () {
-	var db = null, url, gameName, timestamps = {};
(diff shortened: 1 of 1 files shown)


mpetrovic:
Thanks for your help! I really appreciate it.

PR #355, opened 2012-10-01
When autoPause setting is true, window blur handler should check Crafty is not already paused before calling Crafty.pause().

jmwohl:
Because the game might be in a paused state before the blur occurs, the window blur event handler should check that the Crafty is in fact paused before calling pause().  An alternative would be to add the boolean param to the call to pause().


mpetrovic:
same with your last one. can you clear out all these other commits?

PR #357, opened 2012-10-01
Pause audio by id

jmwohl:
Added ability to pause/unpause/togglePause of individual audio objects. Useful for pausing music during a 'pause screen' without pausing other sounds, among other things.


jmwohl:
FYI, the diff shows a lot because my IDE changed the formatting.  The relevant changes are the pause/unpause/togglePause methods.

PR #358, opened 2012-10-07
Added documentation for Crafty.viewport.bounds

MaxBareiss:
This adds documentation about the `Crafty.viewport.bounds` variable added in pull request #338.


mpetrovic:
bounds should be added to the viewport object explicitly and then the documentation would go there.

MaxBareiss:
Is it correct now?

PR #365, opened 2012-10-19
Update src/loader.js

jmwohl:
The extension should be limited to 3 characters in order to allow GET vars to be added to the src paths.  This is useful for cache-busting for browsers that don't fire the onload event when assets are pulled from browser cache.


mpetrovic:
wouldn't a better method be to search for the ? character and substr to that instead of hard-coding to 3? After all, .jpeg is a thing.

PR #369, opened 2012-11-01
SVG support for Crafty.load

koraa:
- Add the ".svg" suffix to the  list of image suffixes
- Use a list instad of a static "IF" to check wether the file is an imager
  => Users can add more suffixes

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -47,7 +47,37 @@ Crafty.extend({
             this.trigger("NewAsset", {key : key, value : value});
         }
     },
-
+        /**@
+	* #Crafty.image_whitelist
+	* @category Assets
+	* 
+    * 
+    * A list of file extensions that can be loaded as images by Crafty.load
+    *
+	* @example
+	* ~~~
+        * Crafty.image_whitelist.push("tif")
+	* Crafty.load(["images/sprite.tif", "sounds/jump.mp3"],
+	*     function() {
+	*         //when loaded
+	*         Crafty.scene("main"); //go to main scene
+	*         Crafty.audio.play("jump.mp3"); //Play the audio file
+	*     },
+	*
+	*     function(e) {
+	*       //progress
+	*     },
+	*
+	*     function(e) {
+	*       //uh oh, error loading
+	*     }
+	* );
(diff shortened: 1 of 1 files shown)


src/loader.js
```diff
@@ -164,7 +195,7 @@ Crafty.extend({
                 }
                    
                  
-            } else if (ext === "jpg" || ext === "jpeg" || ext === "gif" || ext === "png") { 
+            } else if (Crafty.image_whitelist.indexOf(ext) >= 0) {
```
BlackScorp, line 198: isnt it better to check for -1 ? and using lastIndexOf because of filenames like foo.jpg.bar.test

koraa, line 198: > isnt it better to check for -1 ?

Why? I normally check for _any invalid responses_ <=> _x < 0_;
If you want to check for _the specific response that there is no such elem_ you would check for _x == -1_;
This is basically a question of taste (is it not?).

> lastIndexOf because of filenames like foo.jpg.bar.test

No; basically I did this: 

```
ext `elemof` Crafty.image_whitelist
```

because _ext_ should always be the last extension:

```
ext = current.substr(current.lastIndexOf('.') + 1).toLowerCase();
```

Using _lastIndexOf_ forces the programm to check every element, while _indexOf_ may terminate at the first occurence, and therefore might be a little faster.

mpetrovic:
Also, please open the request against the develop branch.

koraa:
(=> Develop Branch)

src/loader.js
BlackScorp, line 198: maybe if(Crafty.image_whitelist[ext]) would be even faster?

koraa, line 198: > maybe if(Crafty.image_whitelist[ext]) would be even faster?

It's an array. Not a dict.

Are you suggesting to use the HashTable mechanisim for speed enhancement?
Normally I would agree, but we are dealing with very few entrys here, so it would not make a difference and if nothing else is not worthy the code-strangeness.

If not: I suggest reading this: http://blog.xkoder.com/2008/07/10/javascript-associative-arrays-demystified/

PR #380, opened 2012-11-21
Workaround webkit bug for image onload event when image is already loaded

starwed:
There's a bug in webkit that means the on load event is never fired (http://code.google.com/p/chromium/issues/detail?id=7731)

This implements the workaround suggested in comment 5 of that bug -- set the image src to "" and back again.


mpetrovic:
Please open the pull request against the develop branch.

Other than that, looks good.

Issue #345, opened 2012-09-13
Elapsed time


tylorr:
In Crafty, how does one measure the amount of time in seconds since the last frame was drawn?

I want to use the elapsed time for movement logic such as:

pos = pos + speed \* elapsedTime

mpetrovic:
The very last post in this thread details a method for doing so:
https://groups.google.com/forum/?fromgroups=#!topic/craftyjs/qtUQaDzVYAM

tylorr:
Sorry, I am looking for the time since the last frame. I want to be able to use it for my movement logic, something like:

pos = pos + speed \* elapsedTime

I updated the issue to reflect this more clearly.

mpetrovic:
You'll probably want to keep track of it directly. Currently Crafty doesn't have that kind of functionality.

starwed:
Fixed.

PR #388, opened 2012-12-15
Better timing measurements:  debugStep() is an alternate game loop with logging info.

starwed:
This adds a function, Crafty.timer.debugStep(), that can replace Crafty.timer.step().

It records the exact timing of events inside the game loop (such as when frames and draw occur) and passes that along by triggering a "TimeLog" event.

This is helpful in designing tests and benchmarks that measure Crafty's performance.

I liked the clarity of outright replacing the game loop to enter debug mode, but I could always refactor it to sit inside the regular step() function, toggled on/off by some conditional.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -37,6 +37,7 @@
 
     GUID, FPS, frame, components, entities, handlers, onloads, tick, requestID,
 	noSetter, loops, milliSecPerFrame, nextGameTick, slice, rlist, rspace,
+    timeLog={postFrameTimes:[]}
 
 	initState = function () {
     	GUID = 1; //GUID for entity IDs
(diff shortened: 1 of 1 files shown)


mpetrovic:
Is there any way to activate this?

starwed:
You activate it by directly replacing the step function:

`Crafty.timer.step = Crafty.timer.debugStep`

This appealed to me because there's no change _at all_ to the game logic if you don't use it.

starwed:
I'm going to close this for now; I think there should be a more systematic approach to adding timing info to the Crafty game loop.  That way it'll be useful not just for benchmarks/debugging but also things like #345

PR #391, opened 2012-12-18
Unbind event for HitBox

jhcp:
Added a RemoveComponent handler to HitBox. It unbinds its EnterFrame
event and clear the HitBox canvas


mpetrovic:
Any idea why the entire file was rewritten?

jhcp:
No idea - I tried to roll back the commit and do the changes again, but with the same result. I added comments to the diff, so that you can see what was changed:
WiredHitBox: lines 33-46 from the original file were replaced for lines 33-56
SolidhitBox: lines 81-93 from the original file were replaced for lines 91-114

louisstow:
Maybe the tabs were converted to spaces?

PR #394, opened 2012-12-20
fix bug in disableDrag()

sbyrnes321:
It is possible and often useful to call disableDrag() while the entity is not being dragged. In that case, there's no reason to call stopDrag(). Actually, there is a good reason NOT to call stopDrag(): It will set off the "StopDrag" trigger which is obviously inappropriate.


mpetrovic:
Please reroll this against the develop branch.

PR #401, opened 2013-01-07
Fixing key press detection for Opera

### LINKED ISSUE #400: Fourway controls (WASD + arrow keys) don't work in Opera
Crafty's Fourway (and possibly Twoway, etc.) component doesn't respond to W, A, S, D, or the arrow keys in Opera.

As an example, see the demo (one of the ones linked from craftyjs.com) for Cron: http://blackscorp.github.com/cron/ 

Here's my full system info:

Version 12.12 
Build 1707 
Platform Mac OS X 
System 10.7.5

darrentorpey:
Resolves #400

I based this on how the jQuery Event object works. This should help protect us from similar issues in the future.

If we need to copy more properties from the original event object, add them to the "props" variable.

### DIFF
--- a/src/controls.js
+++ b/src/controls.js
@@ -250,13 +250,14 @@ Crafty.extend({
 		// Use a Crafty-standard event object to avoid cross-browser issues
 		var original = e,
 			evnt = {},
-			props = "char charCode keyCode type".split(" ");
-		for (i = props.length; i;) {
-			prop = props[--i];
+			props = "char charCode keyCode type shiftKey ctrlKey metaKey timestamp".split(" ");
+		for (var i = props.length; i;) {
+			var prop = props[--i];
 			evnt[prop] = original[prop];
 		}
 		evnt.which = original.charCode != null ? original.charCode : original.keyCode;
 		evnt.key = original.keyCode || original.which;
+		evnt.originalEvent = original;
 		e = evnt;
 
 		if (e.type === "keydown") {
--- a/src/controls.js
+++ b/src/controls.js
@@ -247,7 +247,18 @@ Crafty.extend({
 	* Unicode of the key pressed
 	*/
 	keyboardDispatch: function (e) {
-		e.key = e.keyCode || e.which;
+		// Use a Crafty-standard event object to avoid cross-browser issues
+		var original = e,
+			evnt = {},
+			props = "char charCode keyCode type".split(" ");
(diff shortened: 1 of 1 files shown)


darrentorpey:
My goal here was to start with a minimal working set. I'm not sure how much or what kind of future needs we'd like to anticipate.

I gather the input libraries part of Crafty is fledgling so I wasn't sure what needs I could anticipate. Is there a roadmap for more generalized support of control schemes?

mpetrovic:
Everything is fledgling and no, there's no roadmap past the work I want to
do with Workers right now. I put that into the pile of things to work on,
though.

darrentorpey:
We could also set a "originalEvent" property on our custom event. That will allow transparency into the original event for developers. I think this would be helpful in general even as we develop a more robust set of control utilities/components in Crafty.

mpetrovic:
we should definitely do that, but it should be a last resort option for advanced usage

src/controls.js
```diff
@@ -247,7 +247,18 @@ Crafty.extend({
 	* Unicode of the key pressed
 	*/
 	keyboardDispatch: function (e) {
-		e.key = e.keyCode || e.which;
+		// Use a Crafty-standard event object to avoid cross-browser issues
+		var original = e,
+			evnt = {},
+			props = "char charCode keyCode type".split(" ");
+		for (i = props.length; i;) {
+			prop = props[--i];
```
louisstow: this looks like it should be `var`ed else it be global

src/controls.js
```diff
@@ -247,7 +247,18 @@ Crafty.extend({
 	* Unicode of the key pressed
 	*/
 	keyboardDispatch: function (e) {
-		e.key = e.keyCode || e.which;
+		// Use a Crafty-standard event object to avoid cross-browser issues
+		var original = e,
+			evnt = {},
+			props = "char charCode keyCode type".split(" ");
```
louisstow: Other props I would add to this list is `shiftKey`, `ctrlKey`, `metaKey` and `timestamp`. That would probably do for now. Do this and the var thing and I will pull :)

src/controls.js
darrentorpey: Ah, cool call; I missed that one.

louisstow:
I also agree with storing the `originalEvent` object on this.

louisstow:
Great, thanks a lot!

PR #403, opened 2013-01-08
Change Crafty('*') output to follow style of Crafty(componentname)

sbyrnes321:
Prior behavior: Crafty('_') returns {2:{...entity2...},
3:{...entity3...}, ...} New behavior: Crafty('_') behaves exactly like
Crafty(componentname), i.e. {0:2, 1:3, 2:6, ...}  if the first three
entities are index 2,3,6, or just return the component if there's only
one match. Documented the new behavior.

Crafty('*').each(...) now works, it didn't before.

WARNING: NOT BACK-COMPATIBLE.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -107,10 +107,10 @@
                         i++;
                     }
                     this.length = i;
-			        // if there's only one entity, return the actual entity
-			        if (i === 1) {
-				        return entities[this[0]];
-			        }
+                    // if there's only one entity, return the actual entity
+                    if (i === 1) {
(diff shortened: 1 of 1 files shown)


sbyrnes321:
Since this is not back-compatible, I searched the repository for the strings Crafty('_') or Crafty("_") -- a sign of code that would need to be changed. I didn't find those strings anywhere except right here. (I only searched this repository; I didn't check, e.g., the sample games linked on the Crafty website etc.)

PR #406, opened 2013-01-21
Develop

dobnarr:
Fix for Twoway jumping bug.  Think I've got it right this time.  Thanks for the help.

### DIFF
--- a/src/controls.js
+++ b/src/controls.js
@@ -827,7 +827,7 @@ Crafty.c("Twoway", {
 	* @comp Twoway
 	* @sign public this .twoway(Number speed[, Number jumpSpeed])
 	* @param speed - Amount of pixels to move left or right
-	* @param jumpSpeed - How high the entity should jump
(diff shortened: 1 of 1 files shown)


mpetrovic:
This won't solve the problem you're describing. When jump is explicitly set to 0, !jump will return true.

dobnarr:
I screwed it up worse than that - there's no "this._jump"
I'll redo it.

sbyrnes321:
If you want to test whether the argument jump is missing, just do:

if (jump === undefined) {...}

Don't forget that third equals sign ;-)

It behaves exactly like you expect and the code is extremely easy to read and understand.

dobnarr:
OK, I did it a different way (arguments.length < 2) and resubmitted.  If you all prefer the undefined, I can do it that way instead.  So we can close this request - it's not correct.  The other one should be good.

louisstow:
This has been fixed in another PR

PR #407, opened 2013-01-21
Fix bug in Twoway jumping

dobnarr:
Fixes bug in Twoway.  Original code was:
jump = jump || this._speed \* 2;

If the user explicitly set the jump speed to zero, it would be reset because of the || operator, leading to unexpected jumping.

I also changed some of the comments to match the current variable names.

Discussed here: 
https://groups.google.com/forum/?fromgroups=#!topic/craftyjs/geUU8qnucTQ

### DIFF
--- a/src/controls.js
+++ b/src/controls.js
@@ -825,9 +825,9 @@ Crafty.c("Twoway", {
 	/**@
 	* #.twoway
 	* @comp Twoway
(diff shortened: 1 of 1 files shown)


dobnarr:
I think I finally did it right.  Sorry - please pardon my incompetence while learning the Github system.

PR #409, opened 2013-01-27
Fix sharing of textFont object in Text component

zlandau:
The Text component was creating a this._textFont object outside of init.
This was causing the structure to be shared with all instances of Text.
Whenever one instance changed the font it would be changed for all
instances.

This commit moves the initialization into init() so each instance will
get its own copy.

### DIFF
--- a/src/text.js
+++ b/src/text.js
@@ -7,16 +7,16 @@
 */
 Crafty.c("Text", {
 	_text: "",
-	_textFont: {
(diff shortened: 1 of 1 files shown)


starwed:
This looks good, but should be against develop.

_e:_ Eh, since it seems like we'll need to fix the whole master/develop mess in one pass, might as well accept this anyway.

PR #410, opened 2013-01-31
Unbind event for HitBox (2)

jhcp:
I hope the diff is now seeing only the actual changes, could you please cancel my previous pull request on the same subject?

Description: Added a RemoveComponent handler to HitBox. It unbinds its EnterFrame
event and clear the HitBox canvas

### DIFF
--- a/src/hitbox.js
+++ b/src/hitbox.js
@@ -30,7 +30,7 @@ Crafty.c("WiredHitBox", {
 			}
 			var ctx = c.getContext('2d');
 			var drawed = 0, total = Crafty("WiredHitBox").length;
-			this.requires("Collision").bind("EnterFrame", function () {
+			var drawBoxFunction = function () {
 				if (drawed == total) {
 					ctx.clearRect(0, 0, Crafty.viewport.width, Crafty.viewport.height);
 					drawed = 0;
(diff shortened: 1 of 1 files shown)


starwed:
It doesn't seem like it should be necessary to manually call clearRect here.  I need to read through HitBox a more carefully to figure out what's going on.

starwed:
Ok, I think I understand what it's trying to do -- these components create a whole separate canvas element to draw on.   Each hitbox is drawn on an EnterFrame event.  

I think the intent is that the hitbox canvas is cleared right before the next cycle is started, but... this is determined by counting the number of hitbox frames when the object is created?  And both the `total` and `drawed` variables are local, so tracked separately for each component?  As is, the _existing_ code seems broken, so I'm not sure it makes sense to merge this PR.

If I'm misunderstanding the existing hitbox.js code and it works as intended, set me straight!  :D

starwed:
I'll go ahead and land this, but hitbox.js really needs a complete rewrite.  I opened issue #451 for that.

jhcp:
Hi, sorry I couldn't reply earlier. Indeed, it is flawed, but now it is a bit less flawed. The hitbox is used just for debugging, it doesn't go into production, thus I'm not much concerned with its quality. Nonetheless, I did this small change just because I wanted to be able to turn it on and off at runtime

starwed:
Yeah, that's why I landed the code -- it doesn't help anything to _not_ land it, since you already did the work.

Even for a debug tools it is a big too buggy for my liking -- for instance, the hitboxes often lag behind their actual position, and the way it decides to clear the hitbox canvas is hilariously broken if you examine what its actually doing.  :)  Like I mention elsewhere, having a good story for drawing debug info would make sense, which I plan on doing after the current backlog of patches/major issues has been cleared.

Issue #178, opened 2012-01-11
crop does not work during animation


BlackScorp:
Crafty.c("Player",{
init:function(){
this.requires("SpriteAnimation").crop([10,10,20,20])//just example crop
.animate("walk_left", 6, 3, 8) //During animation the Entitie will have full size and not croped size
.animate("walk_right", 9, 3, 11)
.animate("walk_up", 3, 3, 5)
.animate("walk_down", 0, 3, 2)
}
});

mpetrovic:
and what did you expect to happen?

On Wed, Jan 11, 2012 at 4:54 PM, Witali Mik <
reply@reply.github.com

BlackScorp:
i exect that if i crop an entitie, so it will have the size of crop during his exists..

for example, i have a player sprite that is lets say 200x200 size in px. but each sprite image has a margin to other images, so if i want to make a smaller entite than the spritesize, i have to crop it. at start the entitie is smaller than the sprite image size. and after animation it has its base size again..

starwed:
This is basically the same issue as [described here](https://groups.google.com/forum/#!topic/craftyjs/SIG1cJXWW1U/discussion), except with `__trim` instead of `__padding.`   That was fixed in #241.   I think this could be fixed in a parallel way, assuming that cropping should apply to all sprites.  If not it'll be a little bit more complicated, but still fairly straightforward.

Why do we have two different ways of defining essentially the same thing?  Is there some distinction I'm missing?

starwed:
Should be fixed by PR #665?

PR #426, opened 2013-03-02
Sprite animation enhancements

grozen:
This (rather beefy) change aims to implement the features listed [here](https://groups.google.com/forum/?fromgroups=#!topic/craftyjs/spfDtDN0M60). New capabilities include:
1. Pausing an animation.
2. Resuming an animation.
3. Playing an animation from a specific frame.
4. Getting more information on each frame change (name of animation + new frame).

Includes documentation, tests, and even a small "playground" area on the test page.

I'm pretty sure some of these changes are breaking. While I tried to avoid this as much as possible, I cared more for writing code that was good and made sense. You are welcome to disagree.


starwed:
So this _absolutely_ is not backwards compatible, since the method used to start the animation has changed.  I'm fine with that, since `.animate()` being so overloaded was confusing to me (and obviously a fragile state of affairs), but I imagine it would break a lot of code and demos out there.  

I've read through the patch and it looks mostly good, but a few nits.  (There's a lot of code, so several comments!)
- When a function is called with fewer arguments than it has parameters, the remaining params take on the value undefined, not null.  Since most of the comparisons use non-strict equality (== instead of ===) it _works_, but using strict equality (and nonequality) is probably better.
- It's hard to tell what `cyclesPerFrame` means on first read through; it's more strictly something like `framesPerAnimationFrame` but that would be a bit silly.  Not sure what the best solution is here, but if the name isn't changed the usage of cycle and frame should be clarified in a comment.
- It seems like we're using an awful lot of variables to keep track of state.  When infinite repetition is requested, maybe just set the value of `repeat` to be `Infinity`; that would eliminate the need for one variable and simplify the conditional checks.  (If `x` is `Infinity`, then so is `x - 1`.) 
- If you can see an easy way to avoid creating objects on the various triggered events that would be ++good; objects that are created per entity per frame lead to GC problems.  You might consider just passing the reel object directly, for instance, depending on the sorts of use case you anticipate.
- `Infinitly` should be `Infinitely`

If you want to take a look at issue #178 that would be pretty easy to fix along with this patch; but certainly not necessary.

Finally, I haven't taken the time to look at and run through the tests yet, but since I need to understand how QUnit is being used in Crafty I'll force myself to do that too.  Major kudos for attaching tests, though!

grozen:
Thanks for taking the time to look at this @starwed! I'm a little swamped at work at the moment but I should get around to fixing your comments in ~a week. With that in mind, I'll address your points:
- I actually did that on purpose because I hate equating things to undefined. I also usually align myself with the choices coffeescript makes when generating javascript code. If this is something that's considered heinous I will obviously change these.
- I was worried this will end up confusing. I'll be sure to clarify it somehow.
- Sure thing
- Could you point me to places where I do this? This is obviously an oversight that I will correct.
- I'll exterminate this typo.

I'll be glad to look at #178 as well, but just to make sure, the problem is that the entity size resets after the animation plays out?

starwed:
_re:equalities:_ Not quite sure what you mean here; [Coffescript's](http://coffeescript.org/#operators) existential operator checks against both `undefined` _and_ `null`.   Looking more closely, if you neglect to pass a duration to `.play()` this isn't noticed by the code, because `undefined !== null` is true.  Since duration is a required parameter, this isn't really buggy, but if someone later changed the code to, say, provide a default duration, it might be a bit hard to notice what's going on.  Being consistent by using strict comparison everywhere, and then explicitly checking against undefined, will produce the most maintainable code.

I'll add some notes where objects are being created.  No hurry; because this patch will definitely break code it might make sense to wait until all the other changes this month have been thoroughly tested before landing it.  (Theoretically Crafty does monthly releases, though obviously that hasn't happened lately!)

re: #178: If I understand what's going on, you just need to further modify [this change](https://github.com/craftyjs/Crafty/pull/241/files) to account for `__trim` as well as `__padding`.  I might be missing something, though.

src/animation.js
```diff
@@ -35,175 +47,252 @@ Crafty.c("SpriteAnimation", {
...
+			}
+			else {
+				currentReel.currentFrameNumber = fromFrame;
+				currentReel.cycleNumber = 0;
+			}
 		}
 
-		if (this._frame.frameNumberBetweenSlides++ === data.numberOfFramesBetweenSlides) {
-			var pos = data.currentReel[data.currentSlideNumber++];
+		this.trigger("FrameChange", { reelId: this._currentReelId, frameNumber: currentReel.currentFrameNumber });
```
starwed: Object created.  Maybe just pass currentReel, which should have all the right info?

grozen:
The existential operator is not actually what I had in mind when I thought of the lax comparison, but that is beside the point. I'll be sure to strict all my comparisons up.

src/animation.js
```diff
@@ -35,175 +47,252 @@ Crafty.c("SpriteAnimation", {
...
+			// If we went through the reel, loop the animation or end it
+			if (currentReel.currentFrameNumber >= currentReel.frames.length) {
+				if (currentReel.repeatInfinitly === true || currentReel.repeatsRemaining > 0) {
+					currentReel.repeatsRemaining--;
+					currentReel.currentFrameNumber = 0;
+				}
+				else {
+					currentReel.currentFrameNumber = currentReel.frames.length - 1;
+					this.pause();
+					this.trigger("AnimationEnd", { reelId: this._currentReelId });
```
starwed: Again, maybe just pass currentReel directly?

src/animation.js
```diff
@@ -35,175 +47,252 @@ Crafty.c("SpriteAnimation", {
...
+					return;
+				}
 			}
 
+			this.trigger("FrameChange", { reelId: this._currentReelId, frameNumber: currentReel.currentFrameNumber });
```
starwed: This is where most obj creation would occur; just pass currentReel?

src/animation.js
```diff
@@ -35,175 +47,252 @@ Crafty.c("SpriteAnimation", {
...
+		if (currentReel === undefined) {
+			throw "The supplied reelId, " + reelId + ", is not recognized.";
+		}
+
+		this.pause(); // This will pause the current animation, if one is playing
+
+		this._currentReelId = reelId;
+
+		if (duration !== null) {
+			currentReel.cyclesPerFrame = Math.ceil(duration / currentReel.frames.length);
```
starwed, line 199: this line is executed when duration is `undefined`, such as the case where no duration is passed

starwed:
Ah, [this stackoverflow post](http://stackoverflow.com/questions/13464552/node-js-strange-behavior-of-coffeescript-compiler?lq=1) explains some of the subtleties of checking for existence, and how the coffeescript compiler can do some smart things.

starwed:
I wonder if this will fix #228.

grozen:
It will indeed fix #228. I addressed this specifically [here](https://github.com/grozen/Crafty/commit/e03d67b8f7bed81b59a1acf49e274a94b7ca0432).

grozen:
Alright, @starwed, I believe I've dealt with the issues you pointed out (though I haven't gotten around to including a fix for #178, my apologies). Can't see the commits yet, but I imagine they will pop up soon.

Object creation wise, I now reuse a pair of objects instead of creating new ones. I've done some profiling, and I have to say it doesn't seem to have a serious impact over memory utilization (though it is obviously the right thing to do). Actually, when playing an animation, memory starts shooting up and then getting collected even without firing any events (I disabled those lines for a while). Maybe that's something worth looking at in the future.

starwed:
Seems to work for me.  I wouldn't be shocked if it turned out there were some issues left, but the best way to ferret those out is probably to just merge it.

My only remaining concern is maybe the method names.  Things like `play` and `pause` might be too generic, which can be a problem when you want to combine components.    They could be changed to `playAnimation`, `resetAnimation`, `pauseAnimation` etc, but I don't feel super strongly about it.  Anyone have an opinion here?

_e:_ Oh, and this breaks the rpg demo, since it uses animation.  I can fix that in a new PR if you don't feel like messing with the demos.

grozen:
I tend to be on board with you regarding the method names (is anyone else even watching this pull request?). I'll give it a little while and change it.

Also, I hate to break things, so I'll revive the RPG demo.

grozen:
Alright, It's done. Method names changed, demo revived.

starwed:
Thanks!

PR #427, opened 2013-03-05
Improved dirty rectangles algorithm

statico:
Attached is an improved algorithm for Crafty's dirty rectangle merging which is faster and more correct. It only ever manifests in Crafty applications which use sprites with greater-than-1-bit alpha channel. It was originally written by Mark Logan (@mlogan).

Demo using master: http://statico.github.com/crafty-dirty-rectangles/master.html

Demo using code from this PR: http://statico.github.com/crafty-dirty-rectangles/modified.html

(View developer console to see the benchmarks.)
## The Problem

Here's a screenshot of an example application using the latest version of Crafty. Red arrows point to the artifacts:
Screen Shot 2013-03-04 at 6 17 52 PM
## Benchmarks

Five averages of 100 calls each using the algorithm from master, in seconds:

```
merge: 0.040643565508878175
merge: 0.04171287193709966
merge: 0.03759406384836772
merge: 0.040346535273117594
merge: 0.036881185980458364
```

And the same using the new algorithm:

```
merge: 0.0217425726852057
merge: 0.02055445750529813
merge: 0.02647524913957361
merge: 0.02366336844031616
merge: 0.021237621803094846
```

### DIFF
--- a/src/drawing.js
+++ b/src/drawing.js
@@ -289,6 +289,220 @@ Crafty.extend({
 	}
 });
 
+var DirtyRectangles = (function() {
+
+	function x1(rect) { return rect._x; }
+	function x2(rect) { return rect._x + rect._w; }
+	function y1(rect) { return rect._y; }
+	function y2(rect) { return rect._y + rect._h; }
+
+	function intersects(a, b) {
+		return x1(a) < x2(b) && x2(a) > x1(b) && y1(a) < y2(b) && y2(a) > y1(b);
+	}
+
+	var corner_data = {};
+
+	function reset_corner_data() {
+		corner_data.x1y1 = false;
+		corner_data.x1y2 = false;
+		corner_data.x2y1 = false;
+		corner_data.x2y2 = false;
+		corner_data.count = 0;
+	}
+
+	// Return the number of corners of b that are inside a.
+	// _cornersInside stores its results in _corner_data. This is safe to do
+	// since the only recursive call in this file is in tail position.
+	function corners_inside(a, b) {
+		reset_corner_data();
+
+		// The x1, y1 corner of b.
+		if (x1(b) >= x1(a) && x1(b) <= x2(a)) {
+
+			// The x1, y1 corner of b.
+			if (y1(b) >= y1(a) && y1(b) <= y2(a)) {
+				corner_data.x1y1 = true;
+				corner_data.count++;
+			}
+			// The x1, y2 corner of b
+			if (y2(b) >= y1(a) && y2(b) <= y2(a)) {
+				corner_data.x1y2 = true;
+				corner_data.count++;
+			}
+		}
+
+		if (x2(b) >= x1(a) && x2(b) <= x2(a)) {
+			// The x2, y1 corner of b.
+			if (y1(b) >= y1(a) && y1(b) <= y2(a)) {
+				corner_data.x2y1 = true;
+				corner_data.count++;
+			}
+			// The x2, y2 corner of b
+			if (y2(b) >= y1(a) && y2(b) <= y2(a)) {
+				corner_data.x2y2 = true;
+				corner_data.count++;
+			}
+		}
+
+		return corner_data.count;
+	}
+
+	// Shrink contained so that it no longer overlaps containing.
+	// Requires:
+	//   * Exactly two corners of contained are within containing.
+	//   * _cornersInside called for containing and contained.
+	function shrink_rect(containing, contained) {
+
+		// The x1, y1 and x2, y1 corner of contained.
+		if (corner_data.x1y1 && corner_data.x2y1) {
+			contained._h -= y2(containing) - y1(contained);
+			contained._y = y2(containing);
+			return;
+		}
+
+		// The x1, y1 and x1, y2 corner of contained.
+		if (corner_data.x1y1 && corner_data.x1y2) {
+			contained._w -= x2(containing) - x1(contained);
+			contained._x = x2(containing);
+			return;
+		}
+
+		// The x1, y2 and x2, y2 corner of contained.
+		if (corner_data.x1y2 && corner_data.x2y2) {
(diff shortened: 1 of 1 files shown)


statico:
CC'ing @mlogan

starwed:
Hmm, I just did a different (mostly orthagonal) rewrite of Crafty's dirty rectangle management, but I did touch the code from the old merge function.  There was a straight up math error in how it calculated the newly merged rect, and there was a lot of room for improving just how it ran through the loop as well.  It would probably be worthwhile benchmarking my improved version of the old algorithm to your new version.

Also, I don't see the the buggy behavior in Firefox, only Chrome.  Could you explain what is happening, and why this patch fixes it?  (I'd guess it has something to do with drawing the same alpha sprite twice in the same area when two rectangles are not merged?)

louisstow:
@starwed could those improvements be added to this PR? I would like to use it/merge it soon

mlogan:
@starwed,

The fix done here is that the old dirty rectangle code did not always generate a non-overlapping set of dirty rectangles. This caused areas of the screen to be drawn twice, which is both inefficient, and incorrect in the case of additive alpha blending. The new code generates a final set of dirty rectangles such that no pixel is in more than one rectangle. 

As for firefox, I don't see any behavior at all in FF, as the demo apparently relies on `performance.now()`.

-Mark

starwed:
Sorry I couldn't respond earlier; I was in the middle of teaching.
1. The demo works fine in Firefox Nightly (22.0a).  Superficially from the description it sounds like Chrome is doing the right thing, but are we absolutely certain it's not a bug in Chrome/Webkit?  Why does it look ok in Firefox Nightlies?
2. There's a performance trade-off when deciding how to merge rectangles.  Not merging them means redrawing portions of the screen twice, but merging them means drawing regions that don't need it.  So a change like this could have a direct affect on how long drawing takes.  The old algo was clearly inefficient, but that doesn't automatically mean this one is better.
3. From 2, Crafty really needs some tests and benchmarks that can determine how changes affect performance!   I put together a simple one for full screen redrawing, but a range of tests that look at different usages would be ideal.  Without knowing exactly how a patch alters performance, I'm not sure how anyone is supposed to know whether patches like this or my own should be accepted.

mpetrovic:
For #2, it might be worth looking into drawing polygons rather than
rectangles. If two rectangles intersect, they become a single polygon
that's redrawn once.

This would work better down the line when isometric and 3d cubes are
implemented for canvas.

src/drawing.js
```diff
@@ -289,6 +289,220 @@ Crafty.extend({
...
+								// This case looks like this:
+								// +--------+=========+----------+
+								// |rect    |         |          |
+								// |        |         |          |
+								// +--------+---------+ new_rect |
+								//          +--------------------+
+								// Note how new_rect has 1 corner in rect, while
+								// rect has 2 corners in new_rect.
+								//
+								// Obviously, we shrink rect to not overlap new_rect.
```
starwed, line 458: If rect held a specific entity (which would be a common case) it now must be drawn in two passes.  This is not an obvious win, especially if the bounding rectangle of both only had to be a little bigger.

starwed:
@mlogan 

> The new code generates a final set of dirty rectangles such that no pixel is in more than one rectangle. 

 I don't think this is strictly correct -- when a rectangle is enlarged, it might now cover pixels that overlap with another rectangle earlier in the list.  I don't think this algorithm checks for such effects, but they could still cause a portion of an object to be drawn twice.  If you want to ensure that _every_ overlap has been caught, you'd need to keep looping like the original algorithm did.  (Or come up with something cleverer!)

starwed:
Oh, also, was this supposed to get pulled into develop?  I went to rebase the existing patches I had against this, but then realized that it was in master, not develop.

mpetrovic:
yes, everything goes into develop.

Master should only be currently released code.

mlogan:
@starwed: When rectangles are grown, they are removed from the list and re-processed as though they were new, which handles the case you're describing.

sbyrnes321:
This may be heresy, but have you tried benchmarking what happens if you redraw the whole canvas every time, and just forget about dirty rectangles? It has been claimed that browsers already do dirty rectangle tracking internally: https://twitter.com/gamejs/status/189635929751814144 I don't know if that's true, but it might be worth checking...?

statico:
@sbyrnes321 - Much worse: http://statico.github.com/crafty-dirty-rectangles/drawall.html - I also get a short freeze in animation every few seconds using Chrome.

```
drawall: 0.4351485149527434
drawall: 0.3858118812282141
drawall: 0.4243267327849638
drawall: 0.5091485147546553
drawall: 0.45983168334120395
```

Granted, those benchmarks are for the contrived example above, but when I tried drawing everything every frame with Powderkeg it became unplayable.

starwed:
@mlogan Ah sorry, you're completely right!    Totally missed that even though its right there in the comments.

However, I realized that some of my own changes to how dirty regions are handled [fixes](http://abstractvisitorfactory.com/~tim/AlphaProblems/alternateFix.html) the issue of overlapping alphas in a different way.  So, the question is just what heuristic we should use in merging rectangles; it's ok to have pixels drawn twice if it means the overall rendering is faster.

mlogan:
@starwed I'm pretty sure you don't want to draw pixels with alpha blending more than once.

starwed:
@mlogan In the existing code, all dirty regions are cleared, and then all objects are drawn.  If an object existed in two regions its drawn twice, and that's where the overlapping alpha problem arose.

In [my branch](https://github.com/craftyjs/Crafty/pull/429/files), objects in a dirty region are drawn immediately after it is cleared.  I did this because it reduces the number of clipping operations.  (Just clip once per dirty region, instead of once each time an object is drawn.)   But it has the side effect of preventing overlapping alpha pixels -- the overlapping region is cleared in between draws, instead of being cleared twice at the beginning.

PR #428, opened 2013-03-05
Change to addComponent

starwed:
A change to addComponent, as discussed [here](https://groups.google.com/forum/?fromgroups=#!topic/craftyjs/JMPNyq61KlA).
- Only toggles the __c[id] flag when the component is actually added.
- Doesn't add the component if the flag is already true.

The second change might break some code that was relying on the old behavior, but is clearly the better behavior, and without the first change would break things badly.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -225,7 +225,6 @@
             if (arguments.length > 1) {
                 l = arguments.length;
                 for (; i < l; i++) {
-                    this.__c[arguments[i]] = true;
                     uninit.push(arguments[i]);
                 }
                 //split components if contains comma
@@ -233,21 +232,21 @@
                 comps = id.split(rlist);
                 l = comps.length;
                 for (; i < l; i++) {
-                    this.__c[comps[i]] = true;
                     uninit.push(comps[i]);
                 }
                 //single component passed
             } else {
-                this.__c[id] = true;
                 uninit.push(id);
             }
 
             //extend the components
             ul = uninit.length;
             for (; c < ul; c++) {
+                if (this.__c[uninit[c]] == true)
+                    continue
+                this.__c[uninit[c]] = true
                 comp = components[uninit[c]];
                 this.extend(comp);
-
                 //if constructor, call it
                 if (comp && "init" in comp) {
                     comp.init.call(this);


starwed:
Huh, looking at addComponent I'm pretty sure it could be drastically simplified.   (Currently an array is copied by iteration, directly before iterating through it a second time?!) 

But, merging this fix for now.  It could potentially break existing code as mentioned above, but the current situation is already a bit borked.

jumpcoder:
Now I think both addComponent and requires have the same behavior,I think the requires should not be recommended to use ,because the requires will test if the entity has the component ,if true,it will call addComponent ,then the addComponent will also test if the entity has the component.
I think that the requires can directly call addComponent to compatible with the previous code.

starwed:
@jumpcode Nah, requires doesn't do any separate testing.  That was actually what prompted this bug to be looked into.   It just calls addComponent on every component you hand it.

So they're functionally equivalent, but make the _intent_ clearer.  In the future we could also hook more functionality in to `requires` -- triggering an event when a required component is removed, for instance.

jumpcoder:
Forgive my rush,but I think the second change might confuse the two functions -- `require` and `addComponent`.

Now the `addComponent` has the behavior that beyond his semantic, it result in the two function has some redundant logic.

For example, now the source code of `requires`  function is follow:

```
requires: function (list) {
    var comps = list.split(rlist),
    i = 0, l = comps.length,
    comp;

    //loop over the list of components and add if needed
    for (; i < l; ++i) {
        comp = comps[i];
        if (!this.has(comp)) this.addComponent(comp);
    }

    return this;
}
```

I think it can simplify to this:

```
requires:function(){
    this.addComponent(arguments);
}
```

In other case, according to the semantic of `addComponent`, the end user maybe also make some redundant logic.
for example, we maybe define a component like this:

```
Crafty.c('Player', {
    init:function(){
        if(!this.has('Collision')){
            this.addComponent('Collision');
        }
    }
});
```

It makes the logical repeat.So I think should not add the second change to `addComponent`.

So if the code is:

```
Crafty.c('Player'),{
    init:function(){
        this.addComponent('2D','2D');
    }
})
```

the `addComponent` should add `2D` component for entity twice.because it only care about _adding component_.

starwed:
The problem is that if comp1 requires comp2, then creating `Crafty.c("comp1, comp2")` adds comp2 twice.  This can result in some subtle and hard to find bugs, like [this one](https://groups.google.com/forum/?fromgroups=#!topic/craftyjs/JMPNyq61KlA) that you encountered.  

Since there aren't any reasons you'd _want_ to add a component multiple times, it seemed best for `addComponent` to enforce this.

You're totally right about `requires`, though, there is now redundant logic there!  Thanks for pointing that out.

jumpcoder:
sorry,maybe I didn't make my views clear, I mean that the first change _(Only toggles the __c[id] flag when the component is actually added.)_ has fixed the bug.

But the second change _(Doesn't add the component if the flag is already true.)_ makes the `addComponent`  and `require` equivalent.

So,maybe we should not change the `requires`.I think we should remove the second change of the `addComponent`.
Specifically, I think should remove the two line of the `addComponent`:

```
if (this.__c[uninit[c]] == true)
     continue
```

sbyrnes321:
I agree with @starwed and disagree with @jumpcoder . Adding a component twice is a bad idea that tends to cause hard-to-diagnose bugs. It should not be possible to do it. The fact that `addComponent` and `requires` are redundant suggests that we should think about changing something in the future. For example, one of them could eventually be deprecated, or the behavior of one of them could be changed, etc. But I don't like the change proposed by @jumpcoder, I think it's a change for the worse.

jumpcoder:
@sbyrnes321 I think the  `addComopnent` can be used for somebody that carefully organize their component to avoid adding a component twice.And the `requires` can be used for somebody that don't care about adding a component twice.

After all, if remove the second change, `addComponent` will run slightly faster than `requires`.this is my core reason, for the present, I think the performance is the main factor for HTML5 game.

starwed:
@jumpcoder addComponent is also used when you create a component with `Crafty.e()`.

About speed, let's remember the [Knuth](http://en.wikipedia.org/wiki/Donald_knuth) quote:

> "We should forget about small efficiencies, say about 97% of the time: premature optimization is the root of all evil"

If you have proof that shows there's a significant performance penalty in practice, we can revisit this, but I would be really shocked if it had a real impact.  The amount of time to check whether a component exists should be insignificant compared to the time it takes to add a component to an entity.

jumpcoder:
@starwed @sbyrnes321 Sorry, I think I am little extreme before.

PR #429, opened 2013-03-06
Dirty rectangle rewrite

starwed:
This is a rewrite of the general mechanism that Crafty uses to track dirty areas of the screen.  
1. Only which entities have changed are tracked.  The dirty regions are then calculated right before the scene is drawn.  (The old method was to calculate dirty regions as entities changed, which was inefficient.)
2. The algorithm for merging rectangular regions together was corrected and simplified.
3. This patch also changes the order in which entities are drawn; everything within a particular dirty rectangle is now drawn in one go.

There's an alternative patch for 2 that got merged to master; I'm going to go ahead and submit this pull request anyway.  If that one gets merged to develop, I'll do the work in making them compatible.  However, change 3 would actually _also_ fix the problem the other patch was targeting  ([demo](http://abstractvisitorfactory.com/~tim/AlphaProblems/alternateFix.html)) ; so ideally we'd want to measure how the two algorithms compare.  (As part of change 3, the dirty region is cleared directly before it is drawn to, instead of clearing all regions before drawing any entities.  This prevents the overlapping alpha issue that prompted the other patch)

### DIFF
--- a/src/DOM.js
+++ b/src/DOM.js
@@ -23,7 +23,7 @@ Crafty.c("DOM", {
 		this.bind("Change", function () {
 			if (!this._changed) {
 				this._changed = true;
-				Crafty.DrawManager.add(this);
+				Crafty.DrawManager.addDom(this);
 			}
 		});
 
--- a/src/canvas.js
+++ b/src/canvas.js
@@ -29,21 +29,13 @@ Crafty.c("Canvas", {
 		Crafty.DrawManager.addCanvas(this);
 
 		this.bind("Change", function (e) {
-			//if within screen, add to list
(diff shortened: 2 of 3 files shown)


starwed:
Anyone have comments on these changes?  I'm a little leery of landing something this large without any outside input or testing.    Other than the debug code I added, there should be no outward changes -- strictly performance improvement.

You can set `Crafty.DrawManager.debugDirty = true` if you want to see what parts of the screen are being dirtied.  It's a little messy, but it did let me figure out what was going on with the old merge algo.

starwed:
Ok, since there aren't any comments, merging!

PR #430, opened 2013-03-07
Hashmap revision

starwed:
These patches rewrite portions of Crafty's Hashmap.  Changes:
- Uses the bit pattern `(i << 16)^j`as a coordinate key instead of the string `i + SPACE + j`.  This is faster and creates less garbage, but would break a scene that involved more than 65k cells.  (With the default settings, that would be 2 million pixels.)  I don't this is a real concern, but enlighten me if I'm wrong!  :)
- Add a function to refresh an objects position in the hashmap, rather than removing it and then adding it back.  The old way created a lot of garbage.  (I imagine the refresh function could be further optimized.)
- Where possible, avoid creating objects by reusing existing ones.  Every use of mbr() makes Firefox's GC cry, and we don't want that.

I've used these changes for a long time in a game I wrote, so they seem fairly bug free.  My own benchmarking showed a general performance improvement.   I ran some jsperf tests on [encoding](http://jsperf.com/string-vs-bit-hash/2) and [decoding](http://jsperf.com/hash-reversal-for-bits-and-strings) the hash, and the bitwise version won out with both, but the most noticeable gain was the reduction in GC pauses in Firefox.

### DIFF
--- a/src/HashMap.js
+++ b/src/HashMap.js
@@ -49,7 +49,7 @@
 			for (i = keys.x1; i <= keys.x2; i++) {
 				//insert into all y buckets
 				for (j = keys.y1; j <= keys.y2; j++) {
-					hash = i + SPACE + j;
+					hash = (i << 16)^j;
 					if (!this.map[hash]) this.map[hash] = [];
 					this.map[hash].push(obj);
 				}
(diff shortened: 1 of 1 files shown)


starwed:
I'm going to go ahead and merge this; it shouldn't change the behavior of Hashmap in practice, so if any regressions are spotted I'll back it out.

Just to repeat, I think the refresh function I added could be optimized better, but I feel it's still a direct improvement over the old method.

PR #431, opened 2013-03-19
Refactored Crafty.storage to use pouch.js

chunmun:


mpetrovic:
First of all, please open the pull request against develop.

Second, I think this opens an issue we've been avoiding for a while now.
What is our stance on including other libraries in the engine?

PR #432, opened 2013-03-19
Refactor Crafty.storage to use Pouch.js

chunmun:
Thanks for the tip.

Changed the PR to develop. 

Here's a link to the repo regarding the library I'm using https://github.com/daleharvey/pouchdb.

Basically the guys over at PouchDB are hard at work making it work across browsers and platforms and still pretty lightweight by itself.

I'm trying to write a demo that demonstrates the syncing capabilities of PouchDB syncing from a CouchDB database. That should allow game states to be saved to the local storage while the browser is offline and do the sync with external databases when it gets a connection.


starwed:
@chunmun You'll probably want to rebase/cherry-pick your branch against develop; as it is now, it includes some changes from master that aren't in develop.

sbyrnes321:
Little nitpick: Could you edit crafty-local to pull in the pouch code? That would make it more convenient for developers, because then including crafty-local will still be enough to run all code, as it was before. (e.g. you wouldn't need the 13ee39a change.)

sbyrnes321:
Sorry if these are stupid questions, but I'm not too familiar with this stuff. Can you tell me...
- Does this change help or hurt cross-browser compatibility?

In particular, pouchdb website says IE is not supported, while the Crafty website says IE is supported in general (particular features may be missing but we want them to fail gracefully). If it helps cross-browser compatibility, can you give a specific example, or is that just your guess?
- Is there any improvement or change that someone writing a Crafty game, or playing a Crafty game, would notice? Or is it all under-the-hood? (For example, is it expected that Crafty code would ever call a Pouch function directly?)

starwed:
In addition to everything else, Pouch also uses a different license than Crafty.  (Apache instead of MIT/GPL dual).  I'm going to close this request.

A component that encapsulates Pouch's functionality would be the way to go if you want to make this work available to other people.

PR #434, opened 2013-03-21
Remove entity-factory functionality

sbyrnes321:
The entity-factory functionality is currently broken, and there is
no point in having it in the first place. (Every javascript programmer
already knows how to create reusable code: Create a function!)

Deleting the functionality was discussed and endorsed here:

https://groups.google.com/d/topic/craftyjs/1ESEbI09S7s/discussion

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -45,7 +45,6 @@
 
     	components = {}; //map of components and their functions
     	entities = {}; //map of entities and their data
-        entityFactories = {}; //templates of entities
     	handlers = {}; //global event handlers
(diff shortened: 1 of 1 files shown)


starwed:
> there is no point in having it in the first place.

Well... I guess technically it might be useful for a piece of modular code to expose a helper function.  `Crafty` is the one global object that modules can assume, so I it might make sense in theory to provide a standardized way of accomplishing this.

But since it's broken and no existing code relies on this, I agree -- just axe it.

PR #435, opened 2013-03-25
Update time.js

mucaho:
Simpler iteration

### DIFF
--- a/src/time.js
+++ b/src/time.js
@@ -7,13 +7,13 @@ Crafty.c("Delay", {
 		this._delays = [];
 		this.bind("EnterFrame", function() {
 			var now = new Date().getTime();
-			for (var index = 0; index < this._delays.length; index++) {
+			var index = this._delays.length;
+			while (--index >= 0) {
 				var item = this._delays[index];
 				if(item.start + item.delay + item.pause < now) {
 					item.func.call(this);
 					// remove item from array
 					this._delays.splice(index,1);
-					index--;
 				}
 			}
 		});
--- a/src/time.js
+++ b/src/time.js
@@ -12,8 +12,17 @@ Crafty.c("Delay", {
 				var item = this._delays[index];
 				if(item.start + item.delay + item.pause < now) {
 					item.func.call(this);
-					// remove item from array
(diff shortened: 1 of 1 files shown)


mucaho:
Added changes so that it fixes #396 
The changes do not break existing code.

mucaho:
Easy way to test the new functionality:

``` javascript
<script type="text/javascript">
    window.onload = function() {
        //start crafty
        Crafty.init(400, 336);
        //Crafty.canvas.init();

        Crafty.e("Delay")
            .delay(function() {
                alert("Despicable me!");
                //You can also use Infinity without quotes for repeat count
            }, 3000, 2);
    };
</script>
```

src/time.js
```diff
@@ -7,13 +7,22 @@ Crafty.c("Delay", {
...
 					item.func.call(this);
-					// remove item from array
-					this._delays.splice(index,1);
-					index--;
+					if (item.repeat > 0 ) {
+						// reschedule item
+						item.start = now;
+						item.pause = 0;
+						item.pauseBuffer = 0;
+						if (item.repeat !== Infinity)
```
starwed: Infinity-1 is still Infinity, so no need for a conditional here.

src/time.js
```diff
@@ -7,13 +7,22 @@ Crafty.c("Delay", {
...
+							item.repeat--;
+					} else if (item.repeat === 0) {
```
starwed: If repeat is set to a negative number, it will never be removed.

starwed:
I added a couple of line comments; it needs to be clearer what happens when repeat is set to be negative.  A common convention is for that to repeat infinitely as well.  (I think Crafty.audio.play uses that, for instance.)

starwed:
Thanks for adding the changes mucaho!  I'll need a little bit to make sure I understand how the component work.  (I haven't looked at "Delay" before now.)

starwed:
Ok, looks good to me.  The `if (repeat>0) {} else if (repeat<=0){}`  bit is redundant, but not in a way that makes the code harder to read, so I'll go ahead and merge.

PR #437, opened 2013-03-29
Tiny change to add the ID property to every entity

mucaho:
I have searched the code, there is no apparent way to uniquely identify the entity (unless you extract the id from _entityName or i missed something).

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -24,12 +24,6 @@
     *   Crafty(1)
     * ~~~
     * Passing an integer will select the entity with that `ID`.
-    *
-    * Finding out the `ID` of an entity can be done by returning the property `0`.
-    * ~~~
-    *    var ent = Crafty.e("2D");
-    *    ent[0]; //ID
-    * ~~~
     */
     var Crafty = function (selector) {
         return new Crafty.fn.init(selector);
@@ -371,6 +365,26 @@
 
             this.trigger("RemoveComponent", id);
             return this;
+        },
+
+        /**@
+        * #.getId
+        * @comp Crafty Core
+        * @sign public Number .getId(void)
+        * Returns the ID of this entity.
+        *
+        * For better performance, simply use the this[0] property.
+        * 
+        * @example
+        * Finding out the `ID` of an entity can be done by returning the property `0`.
+        * ~~~
+        *    var ent = Crafty.e("2D");
+        *    ent[0]; //ID
+        *    ent.getId(); //also ID
+        * ~~~
+        */
+        getId: function () {
+        	return this[0];
         },
 
         /**@
--- a/src/core.js
+++ b/src/core.js
@@ -1060,6 +1060,7 @@
 
             entities[id] = null; //register the space
             entities[id] = craft = Crafty(id);
+            craft._id = id;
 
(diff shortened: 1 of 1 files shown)


sbyrnes321:
Actually it is already possible: `some_entity[0]` is its ID.

http://craftyjs.com/api/Crafty.html

If you could tell us the places where you looked (in vain) for that information, it might be helpful for improving the documentation.

Actually, I'm in favor of this pull request and I hope it's accepted, notwithstanding the fact that it's redundant. The reason is, writing `this._id` in your code is much more readable and intuitive than writing `this[0]`.

mucaho:
Ah thanks. I skimmed through that part too quickly. I have seen alot of `this[0]` throughout the source code wondering what that specific property at that index might be.

I don't think that having the redundancy would be good. The "better" alternative would be to refactor all `this[0]` to `this.ID`.

mucaho:
On the other hand refactoring may break existing code :(

mucaho:
What about a wrapper getter method like Crafty Core.has(componentID)? 
That would not break existing code, the property would not be rendundant and most importantly, it would be **better readable**.

``` javascript
/**@
* #.getId
* @comp Crafty Core
* @sign public Number .getId(void)
* Returns the ID of this entity.
*
* For better performance, simply use the this[0] property.
*/
getId: function () {
    return this[0];
},
```

starwed:
Out of curiosity, I checked whether there'd be any performance penalty for using e.id instead of e[0].  [Here is the jsperf link](http://jsperf.com/index-vs-property).  I don't see any real difference in FF or Chrome, so probably no concern there.

A wrapper might make sense, though again, we'd want to make sure there's no performance issues.  Is using a getter faster than calling a function?

_e:_ Actually, the ID is one of the few properties that should never, ever change, right?  Maybe it could just be duplicated at entity creation.

mucaho:
This site is awesome! Thanks for sharing :+1: 
- **Is using a getter faster than calling a function?**
  I created a test comparing the use of a simple `function` vs a `__defineGetter__`. [jsperf link](http://jsperf.com/index-vs-property/2). I ran it on FF19, it was minor win for function. I ran it on chrome it was a bit faster with getter.
  Should i rewrite the code to use `__defineGetter__` instead ? (and the `defineProperty` / normal `function` fallback for browsers that do not support this). How often will `getId()` be called anyway? - existing code uses `this[0]` anyway.
- **Actually, the ID is one of the few properties that should never, ever change, right?**
  AFAIK, id is a simple `Number` thus it is copied by value when returned by a getter method. The user can do with it what he wants, as long as he does not modify `this[0]` directly.

starwed:
I think I'd prefer to just copy the ID into an `.ID` propery on entity creation.  Can anyone see anything wrong with this?  As far as I know the ID will never change, so we don't need to worry about keeping them in sync.

mucaho:
Whats the advantage of having another property over returning it with the help of a function?

starwed:
> Whats the advantage of having another property over returning it with the help of a function?

Just brevity.  After letting the idea simmer for a while, I think a function call is probably less confusing.  So I'll go ahead and merge this as is.

PR #441, opened 2013-04-03
add to documentation of .z and .rotation

sbyrnes321:


starwed:
Silly nitpick; it no longer suggests using mod 360 when assigning.  Is that an intentional change?

sbyrnes321:
I suppose that's intentional. I tried to convey that people can do mod 360 if they want to, but that they don't have to (until they get to millions of degrees). Do you agree? Or is it confusing?

starwed:
I think it's probably fine.  If it ever causes a confused forum post, we can add it back to the docs.

PR #442, opened 2013-04-08
Make the stage element (#cr-stage) more modular

smartboyathome:
I needed to have Crafty use an element other than a div with id 'cr-stage', and I noticed that the support for a modular stage element was already there with Crafty.stage.elem, but there were a couple references to 'cr-stage' still in the code. I updated Crafty.init(w, h) and Crafty.viewport.init(w, h) to take an optional stage_elem parameter that could be either a string or an HTMLElement (known as just an Element in IE8). If no parameter is given, it defaults to the original behavior of finding an element with id cr-stage and, if none exists, creating it under 'body'. In addition to this, I updated the stop function to set the id of the replacement div for stage with the id from the original stage element. Please look it over and let me know what you think.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -753,8 +753,8 @@
         * Uses `requestAnimationFrame` to sync the drawing with the browser but will default to `setInterval` if the browser does not support it.
         * @see Crafty.stop,  Crafty.viewport
         */
-        init: function (w, h) {
-            Crafty.viewport.init(w, h);
+        init: function (w, h, stage_elem) {
+            Crafty.viewport.init(w, h, stage_elem);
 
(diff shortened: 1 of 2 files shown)


mucaho:
Did you test the code? Make a new pull request against the development branch, please

starwed:
This looks mostly ok to me.  It doesn't seem to affect any existing code I have, and it seems that specifying a different stage element ID works as expected.

I'd ask that you add documentation for the new optional parameter, though!  :)

starwed:
I took this commit and added documentation in a new branch.  Closing this one and opening a new PR.

PR #443, opened 2013-04-17
Changes in demo files

lcmartinezdev:
Change the address that you access the src attribute of the script tag


starwed:
Could you explain this one?  (Or link to previous discussion, etc.)

sbyrnes321:
Using crafty-local instead of crafty allows developers to run the demos straight out of the repository directory, without having to run the build script first. And changes in the crafty source files would be immediately reflected in the demos -- again, without having to re-run the build script. As far as I can see, that would be an improvement.

starwed:
Thanks, that makes sense -- merging.

_e:_ Dangit, this was against master.

PR #444, opened 2013-04-24
Clean up core.html unit tests

sbyrnes321:
Cleaning up core.html unit tests...
- Use atomic (i.e. self-contained) tests.
- Fix typos that caused recently-added tests to erroneously fail.
- Consistent whitespace (tabs only).
- Always use strictEqual instead of equal, to avoid false-negatives.

All tests pass right now except the SAT overlap test, which is -13.858192987669298 instead of -18.47759. (The test was already failing before -- I didn't change anything.)

QUnit discovered a global variable creation (from forgetting "var") in collision.js, so I fixed that too.

### DIFF
--- a/src/collision.js
+++ b/src/collision.js
@@ -22,7 +22,7 @@ Crafty.c("Collision", {
         this.requires("2D");
         var area = this._mbr || this;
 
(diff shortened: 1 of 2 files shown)


starwed:
I'm going to trust that you know what you're doing on this one.  (I'm just now getting up to speed on the QUnit stuff.) Merging!

Issue #378, opened 2012-11-08
Cascade Issues


benbarefield:
There appears to be a bug in the cascading of rotation to attached entities:
var this = Crafty.e("2D, DOM, SomeSprite").attr({ x: 5, y : 5 });
var that = Crafty.e("2D, DOM, SomeSprite").attr({ x : 10, y : 5 });
this.attach(that);
this.rotation = 10;

This code results in that.rotation being set to undefined. Looking into Crafty, I found that rotation cascading is done through a rotate(e) function, and it sets the rotation property to e.theta, but e.theta is undefined during the cascade (I don't know if its ever defined)

I don't know if this is a design decision or not, but z values are not being cascaded either.
this.z = -5 does not change that.z

shikolay:
you can not use "this" for variable
https://developer.mozilla.org/en-US/docs/JavaScript/Reference/Reserved_Words
Try to use other identifier

benbarefield:
Sorry - "this" was a poor choice of name. Change "this" in the code to "anythingElse" and you'll see the issue.

benbarefield:
Since this, the undefined rotation seems to have been fixed, but two other issues came up, each with easy fixes:
First, The child object would rotate opposite the parent:
In _rotate, a difference between the previous angle and new angle is calculated, but its done backwards. I had to change:

``` javascript
var difference = this._rotation - v,
```

to

``` javascript
var difference = v - this._rotation,
```

Second, The polygon used for collision detection was also rotating opposite its parent, even with the above change. In polygon.rotate, I changed:

``` javascript
x = e.o.x + (current[0] - e.o.x) * e.cos + (current[1] - e.o.y) * e.sin;
y = e.o.y - (current[0] - e.o.x) * e.sin + (current[1] - e.o.y) * e.cos;
```

to:

``` javascript
x = e.o.x + (current[0] - e.o.x) * e.cos - (current[1] - e.o.y) * e.sin;
y = e.o.y + (current[0] - e.o.x) * e.sin + (current[1] - e.o.y) * e.cos;
```

If you'd prefer I Fork and submit a pull request for this, I can.

starwed:
Fixed by #450

PR #447, opened 2013-04-25
Fix rotation cascading

sbyrnes321:
As discussed in issue #378, rotating a parent does not rotate the child, but instead sets the child's rotation to undefined. (It calls e.theta, which was never defined.) This change seems to fix it, and I added a unit test to confirm.

In #378, @benbarefield also suggests that collision polygons are rotating the wrong way, and proposes a code change to fix it. So I made that change, but I have NOT checked it.


sbyrnes321:
Sorry for various mistakes. Will re-submit a fresh pull request in a bit.

PR #449, opened 2013-04-27
add missing files to crafty-local

sbyrnes321:
It seems that some people have added source files, but forgot to update crafty-local. So I added in the missing files.


starwed:
Pretty straightforward fix; merging.

Issue #175, opened 2012-01-11
IE rotation collision bug 


Antbrooksuk:
Rotating a sprite causes collisions to fail in ie6/7/8

[Google Groups discussion](https://groups.google.com/forum/?fromgroups=#!topic/craftyjs/Lee2AdwwwG8)

sorenbs:
The rotation system needs to be looked at in general.

This simple example behaves strange in all browsers http://jsfiddle.net/qpEGp/

Antbrooksuk:
I've removed rotation from my game for the moment, it would be good to have
a dirty fix or workaround if possible? :)

On Mon, Jan 16, 2012 at 10:05 PM, sorenbs <
reply@reply.github.com

austinh:
I am still having problems with this issue in IE. :( Not sure what to do. Would love it if someone could at least give a dirty fix or lead me on the right track.

starwed:
Could someone who experiences this explain the issue?  The jsfiddle @sorenbs  linked to doesn't load right for me, and the google groups discussion is a bit vague.  (The last links that Anthony Brooks gave seem to be dead?)

sbyrnes321:
I was looking into why that jsfiddle wasn't working. It seems that the collision component makes the draggable component not work! Is this a known issue? Simple case: http://jsfiddle.net/jLzJm/ On my firefox, you can only drag one of the two boxes.

starwed:
It looks like adding "Collision" is blocking other mouse events as well; Maybe the added collision polygon is getting them? ( [jsfiddle](http://jsfiddle.net/jLzJm/3/) )

  I'll open a new issue to investigate, since this seems separate from the rotation/collision issue.

sbyrnes321:
As discussed in #448, the collision box in the above examples is not reset (with .collision()) after the width and height of the component are set to nonzero values.

After taking that into account, here is code that works in firefox and IE10 with the latest crafty. It also works with the last crafty release 0.5.3 (so this is not revealing a problem fixed by #450):

```
Crafty.init(500,500);
Crafty.canvas.init();
var count = 0;
Crafty.e( "2D, Draggable, DOM, Color, Collision" )
    .attr( { w: 20, h: 300 } )
    .collision()
    .color( "red" )
    .attr( { rotation: 30 } );
Crafty.e( "2D, Draggable, DOM, Color, Collision" )
    .attr( { w: 20, h: 300 } )
    .collision()
    .color( "green" )
[...]

PR #450, opened 2013-04-28
fix - rotating parent should rotate child

sbyrnes321:
Previous behavior: Rotating parent caused child rotation to be undefined
Now: Rotating parent rotates the child as expected.

Added unit test that failed before, but passes now.

See issue #378.

Unlike the proposal in #378, this change does not have a side-effect
that messes up collision polygon rotation. (I checked that collision
polygon rotation was correct both before and after this change.)

### DIFF
--- a/src/2D.js
+++ b/src/2D.js
@@ -737,7 +737,7 @@ Crafty.c("2D", {
 		this._origin.y = e.o.y - this._y;
 
 		//modify through the setter method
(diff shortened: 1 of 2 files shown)


starwed:
Looks like everything works to me!  Nice fix.

starwed:
Oh, can anyone check whether this fixes #175?  That issue is about rotation and collision in IE, and I have no windows boxes.

PR #454, opened 2013-05-02
fix -- .each documentation was not copied to website

sbyrnes321:
I don't know exactly how the documentation-generation works, but based
on the model of other similar functions, I bet that the problem is the
missing @comp line.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -620,6 +620,7 @@
 
         /**@
         * #.each
+        * @comp Crafty Core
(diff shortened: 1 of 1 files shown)


starwed:
Yup, this fixes it.  (The build script for the api is `/build/build-api.sh`, though right now it just calls directly into a php script.)  Thanks for the catch! 

I have yet to sit down and go through the build script, but perhaps it is possible to modify it so that it warns about cases like this.

PR #458, opened 2013-05-10
Allow an alternate stage element to be provided

starwed:
This is just #442 but with documentation added.  Copying the original message  by @smartboyathome below:

> I needed to have Crafty use an element other than a div with id 'cr-stage', and I noticed that the support for a modular stage element was already there with Crafty.stage.elem, but there were a couple references to 'cr-stage' still in the code. I updated Crafty.init(w, h) and Crafty.viewport.init(w, h) to take an optional stage_elem parameter that could be either a string or an HTMLElement (known as just an Element in IE8). If no parameter is given, it defaults to the original behavior of finding an element with id cr-stage and, if none exists, creating it under 'body'. In addition to this, I updated the stop function to set the id of the replacement div for stage with the id from the original stage element. Please look it over and let me know what you think.

I'll merge this soon unless someone spots something wrong with the docs, or the actual patch.  The code doesn't allow you to provide the stage element if you don't provide the width or height; does that matter?

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -770,8 +770,8 @@
         * Uses `requestAnimationFrame` to sync the drawing with the browser but will default to `setInterval` if the browser does not support it.
         * @see Crafty.stop,  Crafty.viewport
         */
-        init: function (w, h) {
-            Crafty.viewport.init(w, h);
+        init: function (w, h, stage_elem) {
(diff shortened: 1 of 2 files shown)


starwed:
If it turns out to be a good idea to make this more flexible we can just add that later.

PR #459, opened 2013-05-10
Provide default family and size for text

### LINKED ISSUE #337: "text".textFont : "family" is mandatory
Hi,

From my multiple tests, it seems the property "family" is required on .textFont() call.
Setting a default value for this property would be great.

I suggest a "global" one where we could define a default value for the font family to use.

Tested on Firefox / Canvas - DOM

starwed:
According to the CSS spec, you can't set the font family and size independently.  That means we need to provide defaults.  I chose "10px sans-serif" as the default, since that is apparently what the [canvas uses](https://developer.mozilla.org/en-US/docs/Drawing_text_using_a_canvas).

This mostly fixes #337.  @attonnnn is probably right that being able to set a global default would make sense; any suggestions on the best approach there?  Something as straightforward as just having a `Crafty.defaultTextSize` property?  Or should it be a property of the component somehow?

### DIFF
--- a/src/text.js
+++ b/src/text.js
@@ -20,14 +20,17 @@ Crafty.c("Text", {
 		"size": "",
 		"family": ""
 	},
+	defaultSize: "10px",
+	defaultFamily: "sans-serif",
 	ready: true,
 
 	init: function () {
 		this.requires("2D");
 
 		this.bind("Draw", function (e) {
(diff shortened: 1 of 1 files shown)


starwed:
We could use a better story for changing a component's default settings, so punting on that for now.  I added a mention of the default style to the docs, but left out the specific parameters; I _do_ want to improve how that works in the future, and setting them isn't really any different than setting the properties directly.

PR #463, opened 2013-05-13
Add Crafty.viewport.absoluteScale()

mildmojo:
### Resolution independence: The problem

I need my games to scale to fit the browser window, or at least select something close from a list of known good resolutions. Unfortunately, Crafty's viewport system is working against me. Looking at other open issues, it seems like a common pain point.
### Crafty today

I can't figure out the use case for `Crafty.viewport.scale`. It doesn't work at all for my purposes. Where it applies a cumulative scale factor, I need to apply an absolute scale factor calculated from the window size. Additionally, Crafty resets the canvas transform all by itself when the scene changes or the browser is resized. I need to apply a scale factor that overrides everything else for the lifetime of the game.
### One solution

The existing `Crafty.viewport.scale` method applies the new scale as a multiplier against the old scale. This pull request adds a `Crafty.viewport.absoluteScale` method that sets the scale absolutely, regardless of the old scale. Calling `scale(0.5)` twice results in a scale of 0.25. Calling `absoluteScale(0.5)` twice results in a scale of 0.5. This is useful when scaling the stage in response to browser window resize events. 

Unfortunately, this doesn't fix Crafty's incompatibility between viewport scaling and panning in a mixed Canvas/DOM setting (see [fiddle](http://jsfiddle.net/JUxyE/3/)). It looks like setting `Crafty.viewport.x = 5` (or panning) moves DOM elements by 5 screen pixels and canvas elements by 5 scaled pixels, causing alignment problems. The viewport system may be due for rethinking.
### Sample resizing code

Here's some sample code for a Crafty app that can scale the viewport in response to the browser window size ([fiddle here](http://jsfiddle.net/JUxyE/4/)):

``` javascript
var Screen = { w: 800, h: 480 };

Crafty.init();
Crafty.canvas.init();

var scaler = function() {
  var current = { w: Crafty.viewport.width, 
                  h: Crafty.viewport.height };

  // This will reset the scale when it adjusts the canvas element
  // dimensions, so do it right before setting the scale we want.
  Crafty.viewport.reload();

  if ( current.w > current.h ) {
    // Landscape, scale to fit height
    Crafty.viewport.absoluteScale( current.h / Screen.h );
  } else {
    // Portrait, scale to fit width
    Crafty.viewport.absoluteScale( current.w / Screen.w )
  }
};

// Avoid drawing entities before we're ready to scale.
Crafty.removeEvent( Crafty, window, 'resize', Crafty.viewport.reload );

// Adjust scale to fit inside resized window.
Crafty.addEvent( Crafty, window, 'resize', scaler );

// Run the scaler after scene changes, since they reset the scale.
Crafty.bind( 'SceneChange', scaler );
```
### Discussion

Thoughts? Is there a better way to go about this? :dancer:

### DIFF
--- a/src/extensions.js
+++ b/src/extensions.js
@@ -724,6 +724,66 @@ Crafty.extend({
             }
         })(),
         /**@
+         * #Crafty.viewport.absoluteScale
+         * @comp Crafty.viewport
+         * @sign public void Crafty.viewport.absoluteScale(Number scaleFactor)
+         * @param Number scaleFactor - scale factor relative to original pixels (eg. 2, 4, 0.5)
+         *
+         * Scales the entire canvas by the given factor.
+         *   1.0 = entities render 1:1 between screen pixels and canvas coordinates
+         *   0.5 = entities render at half scale
+         *   2.0 = entities render at 2x scale
+         *
+         * Absolute scale is not cumulative. New scaleFactor will *replace* any previous scaling.
+         *
+         * @example
+         * ~~~
+         * Crafty.init();
+         * Crafty.canvas.init();
+         *
+         * var original = { w: Crafty.viewport.width, h: Crafty.viewport.height };
+         * var scaler = function() {
+         *   var current = { w: Crafty.viewport.width, h: Crafty.viewport.height };
+         *
+         *   // This will reset the scale when it adjusts the canvas element
+         *   // dimensions, so do it right before setting the scale we want.
+         *   Crafty.viewport.reload();
+         *
+         *   if ( current.w > current.h ) {
+         *     // Landscape, scale to fit height
+         *     Crafty.viewport.absoluteScale( current.h / original.h );
+         *   } else {
+         *     // Portrait, scale to fit width
+         *     Crafty.viewport.absoluteScale( current.w / original.w );
+         *   }
+         * };
+         *
+         * // Avoid drawing entities before we're ready to scale.
+         * Crafty.removeEvent( Crafty, window, 'resize', Crafty.viewport.reload );
+         *
+         * // Adjust scale to fit inside resized window.
+         * Crafty.addEvent( Crafty, window, 'resize', scaler );
+         *
+         * // Run the scaler after scene changes, since they reset the scale.
+         * Crafty.bind( 'SceneChange', scaler );
+         * ~~~
+         */
+        absoluteScale: function (scaleFactor) {
+            var ctx = Crafty.canvas.context;
+
+            // Clear entire canvas; scaling invalidates dirty rects and leaves artifacts behind.
+            ctx.save();
+            ctx.setTransform(1,0,0,1,0,0);
+            ctx.clearRect(0, 0, ctx.canvas.width, ctx.canvas.height);
+            ctx.restore();
+
+            // Trick Crafty.viewport.scale into operating in absolute mode.
+            Crafty.stage.inner.style['transform'] =
+                Crafty.stage.inner.style[Crafty.support.prefix + "Transform"] = 'scale(1, 1)';
+            Crafty.viewport._zoom = 1;
+            Crafty.viewport.scale(scaleFactor);
+        },
+        /**@
          * #Crafty.viewport.mouselook
          * @comp Crafty.viewport
          * @sign public void Crafty.viewport.mouselook(Boolean active)


starwed:
Yeah, some of the viewport code is a mess.    (Zoom seems to be completely broken!)

Why do you manually clear the screen in `absoluteScale()`?  Calling `Crafty.viewport.scale` within the function should also trigger a complete redraw.  If that doesn't work there's a bug lurking somewhere.

One problem with using this in response to resize events is that it won't be compatible for zooming/scaling for visual effects. That probably doesn't matter for your particular case, but I wonder if there's a better way to provide that functionality.

mildmojo:
`Why do you manually clear the screen in absoluteScale()?`

I use a full-screen stage (no dimensions passed to `init`). When this code scaled the stage down, bits that were drawn in the area no longer covered by the stage weren't being cleared. So I'd have a smaller, properly-rendered stage and a non-updating giant copy surrounding it. I didn't dig into the Crafty code to figure out exactly why it was happening, but I figured it might be because the dirty rects were scaled down and no longer covered their old locations, so that part of the image wasn't erased. The clear fixed it. I can probably probe deeper here.

Hmm, you're right about this being incompatible with regular zoom/scale. Maybe we can break the absolute scale apart from regular zoom/scale and let the latter figure the absolute scale into their calculations? Basically, let the absolute scale redefine 1.0 when resetting the canvas?

There are also some animation artifacts when scaling down, which may be related to dirty rects and fractional pixels. I haven't looked into that yet, either.

starwed:
When you call scale, it triggers a complete redraw; dirty rectangles are only used for partial redraws.  If there are artifacts left over, that's a bug in the redraw code.  (I have a couple of guesses as to why that might be.)

I think a complete revamp of how the viewport code works is in order.  It would be much more flexible (and not very difficulat) to decouple the implementation of viewport parameters (panning and scale) from the functions that set or animate those parameters.  For one thing, that'll mean we can deal with changing the viewport transformations only during a draw event, rather than (potentially) several times a frame.  It should also make it easier to find and fix the various viewport related bugs.  

So I'll hold off on doing anything with this PR just yet.  It's a good thing to bring up, though!

mildmojo:
I could've written that last bit more clearly. By "animation artifacts when scaling down", I meant that when I tween a sprite from X1 to X2 on a stage that's been scaled down, it leaves a trail around its edges. That's the bit I thought might be related to fractional pixels. It's different from the other problem of original-scale graphics left over outside the now-reduced canvas.

I'm totally in favor of seeing viewport code with way fewer surprises. =) Part of the reason for opening this PR was to see where things lie. We can close it if you like.

I've been looking at Phaser's [test suite](http://gametest.mobi/phaser/), and their [camera scale demo](http://gametest.mobi/phaser/index.php?f=camera%20scale.js&d=cameras) is really slick. If only they were entity/component and had mobile & text rendering figured out. =)

mpetrovic:
If everyone really hates it that much, I can take another crack at it. I
wrote the camera commands in the first place. If they suck, it's my fault.

mildmojo:
Sorry about the dirty commit, too; just realized my text editor introduced a lot of noise when it stripped trailing whitespace. Fixed, if it matters.

mildmojo:
@mpetrovic No worries! I've typically done all my entity positioning by hand with a ton of individual resize callbacks. This was a result of diving into the viewport stuff after the [last Ludum Dare](http://www.ludumdare.com/compo/ludum-dare-26/?action=preview&uid=6748) to see if it could help me out.

Sometime in the future I'll try to make up a few fiddles of the ways the viewport behavior surprises me. It sounds like maybe the viewport features were intended for effects, and I'm trying to use them for utility, like UI alignment.

mpetrovic:
They were designed to mimic camera movements. Scale is actually closer to
the 'dolly' camera movement.

starwed:
@mpetrovic  Had you any thoughts on how to refactor this -- or done any work on it?  The current problem with the viewport isn't really how you call the methods, but bugs in how the backend works. 

My gut feeling is that we should do the following:
- Have the camera movement methods just change some variables representing the position and zoom level of the camera.  Oh, and probably trigger some "ViewportChange" event or the like.
- Have each render method then apply the correct transformations before rendering the scene.

That means the viewport code doesn't need to know about how the render methods work, and means that the transformations aren't applied more often than necessary.  It would let us easily have, say, a background layer that panned at a different rate than the foreground.  It also means that the logic for animating camera motion will be a lot easier to follow and debug.  :)

mpetrovic:
I've already done all of that. It's in the Layers branch in my fork, and
it's been there for months while I've been busy. It still needs some
hammering on and I think a few components need converting away from
EnterFrame, which I renamed to Tick.

starwed:
This PR is pretty out-of-date, and the viewport system has undergone some (if not all) of the changes/enhancements discussed in thread, so closing.

PR #465, opened 2013-05-15
Fix SAT test

starwed:
The SAT collision test in core was expecting the wrong overlap.  It uses `Crafty.circle` to generate two octagonal polygons; their minimal overlap is actually `-15 cos[pi/8] ≈  -13.86`.  The previous incorrect overlap wasn't completely arbitrary: it was `-20 cos[pi/8]`.

For reference, two perfect circles would have an overlap of -15.  But this is testing that SAT works, not how well we approximate circles!

### DIFF
--- a/tests/core.html
+++ b/tests/core.html
@@ -316,13 +316,12 @@ $(document).ready(function() {
 		Crafty("*").destroy();
 	});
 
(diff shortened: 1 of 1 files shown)


sbyrnes321:
Nice, thanks for looking into that!

PR #472, opened 2013-05-19
Grunt

ukko:
Hi, I have created a configuration file for grunt. 

http://gruntjs.com/

Added there a couple of useful rules:
- Concatenation of all the src / *. Js files into one
- Minimize the glued file
- The call syntax checking jshint (as yet no specific rules)
- Running tests qunit

The plans to make a test call and generate documentation in markdown format, and html.

I think that this is not such an important functionality for that would get the new version, but it's utility seriously make life easier for all js developers.

Sorry for my english :-|


starwed:
Looks neat at first glance; we currently use PHP to build docs and concatenate the files.  I'd already been thinking using node might make more sense, given that we're a .js project!  :)

I'll take this out for a spin once we've pushed the next release.

starwed:
Ok, I checked out your branch and got this working.
- Building the regular and minified file seems to work
- The qunit task seems to work, though some of the tests themselves are broken.  `core.html` should be up to date, and when I edit the config file to only run that one, every test passes.
- The one thing that would block us from switching to grunt right now is generating the documentation.  I guess we use a variant of jsdoc, but I'm not clear on whether this is the project's own bespoke creation or something more general.  I can't seem to find any jsdoc projects that use the `@category` tag, for instance, but I've only just looked at this stuff.  Perhaps @louisstow can help?

(Aside: it's probably preferable to rebase against the current branch rather than doing repeated merges.  Unless that would mess up your local workflow.)

ukko:
About documentation generation, I started doing the simplest parser, which is already partly works and can be viewed here https://github.com/ukko/Crafty/blob/md-doc/build/mdoc.js and here its source code on coffee https://github.com/ukko/Crafty/blob/md-doc/build/mdoc.coffee

potomak:
+1 for this PR
-1 for custom docs

what do you think about http://yui.github.io/yuidoc/ and https://github.com/gruntjs/grunt-contrib-yuidoc ?

ukko:
@potomak I think yuidoc can not correctly understand our documentation. 
Because it does not use the format yuidoc, documentation Crafty is a mixture of yuidoc and markdown.

sbyrnes321:
@ukko and @potomak -- To use yuidoc, we would have to go through every entry in the whole documentation and [do XYZ].

Can someone say what exactly is XYZ? If it only takes 10 seconds per entry, we should not rule it out. (I can help.)

starwed:
It wouldn't be too bad to convert between different variants of JSDOC.  However, I think yuidoc might be a bit too class focused?  And AFAICT there is no standard way to define custom tags, and it's missing the type of  `@see` functionality we currently use.  

Also, I really don't like how their own documentation looks & works, which isn't the best sign for a documentation generator!  :)

It's certainly worth considering the various variants of JSDOC, to see if one of them would be well-suited for us.  (We could also simply fork an existing one, if it's a matter of just adding a couple of custom tags or aliases.)

starwed:
@sbyrnes321  So long as all the right _types_ of tags are supported, we could probably do most of XYZ with regexes, since the current format should be easily parseable and in a very similar form to any jsdoc variant.  But I think yuidocs approach to cross-referencing would be trickier to translate.  (From my admittedly not very thorough investigation.)

It would also require the in-code documentation to lie and label components and various objects with `@class`.

The existing parser is actually pretty simple.  Unless there's specific functionality we're missing, I'm ok with simply using a version of it rewritten in js/coffee.

potomak:
@starwed `@see` is not missing, _see_ https://github.com/yui/yuidoc/blob/master/lib/docparser.js#L161

I'm not complaining about the existing parser, but I'm suggesting to embrace _standards_.

starwed:
@potomak If you look further down in the code, the `@see` tag is not actually implemented!  :)

Yuidoc is not a standard.  The big advantage of a standard is that it's commonly accepted by several different tool sets.   Rather, yuidoc is an implementation of the kind of _ad hoc_ standard which is jsdoc.  What we currently use is another variation of that.  

We should look at yuidoc, and jsdoc3, and anything else out there.  If there are features they have in common that our own doc styles needlessly differs from, we should correct that to make our docs more portable.  But from what I can tell, each of those implementations is missing something we currently use.  (There's also just a mismatch in how they name things, since they seem to assume an object-oriented codebase.  Makes sense, since jsdoc emerged from javadoc)

If the only thing missing from yuidoc is the `@see` tag and the ability to alias tag names, maybe we could just fork them and add that functionality.  I'm not sure how much work that would be.

potomak:
https://github.com/yui/yuidoc/blob/master/lib/docparser.js#L611 #todo #fail :)

starwed:
@ukko Did you do any more offline work on the [js based doc generator](https://github.com/ukko/Crafty/blob/md-doc/build/mdoc.coffee)?  If not I'll take a stab at working on it.  :)

ukko:
@starwed I am sorry now absolutely no time to do. I would be happy if you continue to work.

starwed:
@ukko No worries, just wanted to make sure we wouldn't be duplicating our efforts.  And thanks for getting it started!

starwed:
Ok, I have node based doc generation working now.  (I think.)  I wrapped the output in a really, really basic Bootstrap template and [stuck it here](http://ludonomicon.com/~tim/Ludo/Crafty.html).   Should be trivial to use the existing templates instead, just felt like trying a different style out!

There's some things I need to clean up, but I'll probably open a PR soon for this.  Then there shouldn't be anything blocking a switch to grunt.  :)

_e:_ Ah, I forgot about the `@returns` tag.  I should go ahead and document the tag system on the wiki, since it seems to be pretty ad hoc right now.

sbyrnes321:
@starwed (and @ukko) -- Looking good!!! Nice job!

Little mistake I see: .uniqueBind and .one show up in two places each in the left column -- and one of them is a dead link. By comparison, .bind only shows up once.

FYI, I hadn't heard of grunt until this pull request. But lately I've been using it myself and I find it very nice.

starwed:
@sbyrnes321 The `uniqueBind`/`one` mistake is mostly a goof in their comment blocks, not the doc generator.   I'll fix that separately.

I've now set up a grunt task that will generate the docs, and have it producing almost identical output to the current system.

starwed:
I had to rebase this against develop; also went ahead and squashed it down to just a few commits; that became #512. 

Thanks for starting this work!    Releasing and testing crafty should be much easier because of this.

PR #473, opened 2013-05-20
Fixed bug in enableControl()/disableControl()

redbmk:
See bug discussion here:
https://groups.google.com/forum/#!topic/craftyjs/_emWkc7kLDM

### DIFF
--- a/src/controls.js
+++ b/src/controls.js
@@ -765,7 +765,8 @@ Crafty.c("Multiway", {
 	* ~~~
 	*/
 	enableControl: function() {
-		return this.attr('disableControls', false);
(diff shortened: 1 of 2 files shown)


starwed:
Looks good, but maybe not use `attr` just to set a boolean flag.  (In disable/enable). There's some overhead with that (in addition to the function call, it generates an object.)

redbmk:
Good to know. I think I'm probably using .attr() more than I need to in the game I'm working on also. I just added a commit to set it using a simple assignment operator. Tests are still passing - everything seems to be working as expected.

PR #474, opened 2013-05-20
some improvements to text

sbyrnes321:
Where possible, avoid styling text using css, as it makes it harder to
switch DOM <--> canvas (see complaint at
http://buildnewgames.com/game-engine-comparison/), and doesn't always
work anyway. Also, add .unselectable() method

### DIFF
--- a/src/DOM.js
+++ b/src/DOM.js
@@ -239,8 +239,8 @@ Crafty.c("DOM", {
 	*
 	* To return a value, pass the property.
 	* 
-	* Note: For text elements, some css properties are controlled by separate functions
-    * `.textFont()` and `.textColor()`, and ignore `.css()` settings. See Text component for details.
+	* Note: For entities with "Text" component, some css properties are controlled by separate functions
+	* `.textFont()` and `.textColor()`, and ignore `.css()` settings. See Text component for details.
 	* 
 	* @example
 	* ~~~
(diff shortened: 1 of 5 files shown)


src/DOM.js
```diff
@@ -239,9 +239,12 @@ Crafty.c("DOM", {
 	*
 	* To return a value, pass the property.
 	* 
+	* Note: For text elements, some css properties are controlled by separate functions
+    * `.textFont()` and `.textColor()`, and ignore `.css()` settings. See Text component for details.
+	*
```
starwed, line 244: `text elements` should be `"Text" entities` (or something similar).

starwed:
The only qualm I have about this situation is that, in addition to ignoring `.css()`, "Text" entities will now ignore global CSS rules that set family/size.  But you can work around that by using an "HTML" entity -- that makes more sense if you want to use CSS anyway!

One incredibly small nit on the wording; "text element" might mislead someone into thinking those properties would work on such elements in "HTML" entities.  (And I've noticed new users of Crafty have trouble keeping the terms element/component/entity straight, so we  should try to be consistent in the docs.)

sbyrnes321:
Thanks, I changed the wording.

"What is the best approach to text styling" is a separate, larger issue. I'm just trying to make everything consistent with the current approach. Even if the current approach is not _the_ optimal -- I don't know, I haven't evaluated all the possibilities -- it seems like a fine approach to me. Again, it has a nice benefit that the same code will often work for both DOM and canvas text.

starwed:
Agreed.  And thanks for accommodating my incredibly minor request!

PR #477, opened 2013-05-27
HashMap boundries returning wrong coords

### LINKED ISSUE #476: HashMap boundries returning wrong coords
HashMap boundries returning wrong coords

the tile map below is a square but the follow function works funky when x ~> 600px
# i looked into it and it seems as the hashmap boundries was returning the wrong coords

Crafty.init(600, 600);

for(var xxx = 0; xxx < 20; xxx++){
    for(var yyy = 0; yyy < 20; yyy++){
        Crafty.e("2D, DOM, Color")
            .color('rgb('+xxx_10+','+yyy_10+',0)')
            .attr({ x: xxx_64, y: yyy_64, w: 64, h: 64 });
    }
}

Crafty.e("2D, DOM, Color")
    .color('rgb(255,255,0)')
    .attr({ x: 400, y: 400, w: 64, h: 64 });

var player = Crafty.e("2D, DOM, Color, Multiway")
    .color('rgb(255,0,0)')
    .attr({ x: 600, y: 600, w: 64, h: 64 })
    .multiway(3, {UP_ARROW: -90, DOWN_ARROW: 90, RIGHT_ARROW: 0, LEFT_ARROW: 180});

Crafty.viewport.follow(player, 0, 0);

bojank:
fixes #476

### DIFF
--- a/src/HashMap.js
+++ b/src/HashMap.js
@@ -190,7 +190,7 @@
 
         //broad phase coordinate
 				var map_coord = h.split(SPACE),
-					i=map_coord[0],
+					i=map_coord[1],
 					j=map_coord[0];
 				if (i >= hash.max.x) {
 					hash.max.x = i;


starwed:
Nice catch, but this is already fixed in the develop branch.  (As part of a rewrite of the hashmap code.)

In the near future we're going to make master the branch that work occurs on, so less people get caught out by this!

PR #479, opened 2013-06-01
avoid unnecessary function call in _attr

sbyrnes321:
this.pos() is calculated even when it's not used. Actually, it's never
used (because mbr() already has code to call this.pos() if necessary.)
So this is a wasteful function call and object creation.

Also, fixed typos.


starwed:
This is a good fix.  Even with this we still create an object (through `mbr()`) _every time_ an attribute is assigned,so we should probably do some sort of object reuse there.

PR #481, opened 2013-06-01
bump version, update copyright year

sbyrnes321:


starwed:
Hah, just realized I hadn't updated the version, thanks for catching the copyright too!

PR #483, opened 2013-06-06
fix bug where particles can appear in wrong place

sbyrnes321:
In firefox and chrome (at least), the canvases can wind up not aligned on the page. This fixes it in my tests. [Thanks Gizmoa.]

Example:
With old crafty: http://jsfiddle.net/4r6FX/11/
With fixed crafty: http://jsfiddle.net/4r6FX/17/


kevzettler:
:+1:

starwed:
Looks good.  The placement of particles will (I think , haven't tested) get messed up if you scale or shift the viewport, so when the viewport bugs are fixed we'll need to remember to handle that case.

PR #484, opened 2013-06-09
Add a rendering layer for debugging

starwed:
This PR does two major things:
- Refactors the drawing code so that instead of being a hard coded function call, the Canvas and DOM are drawn in response to a "RenderScene" event.  This makes it possible to add additional rendering contexts more easily
- Adds another canvas context for drawing debug info onto, along with several components for common debugging tasks.   An entity with the "DebugCanvas" component will receive a "DebugDraw" event, which it can use to draw to the debug canvas.   
  - I reworked "WiredHitBox" and "SolidHitBox" to use this new context.
  - I added a "VisibleMBR" component for debugging MBR stuff.  (This immediately let me find and solve a related bug!)

The one thing that the first point totally doesn't address is changes to the viewport.  The current viewport code needs some attention, and when that happens viewport functionality can be added here as well.

I should probably add some tests, but since the functionality is basically done I thought I'd open this for people to comment on.  I've tried to document all the new components.


starwed:
I added some pretty basic unit tests.  But some of them rely on a canvas element being present -- is there a good way to handle that so that they would, for instance, run under node or older browsers?

Perhaps tests that need the canvas should be in a special file.

sbyrnes321:
Looks very nice! (Not that I scrutinized every line.)

I think you forgot to update the file list in crafty-local. :-D

Are there any supported browsers that don't have canvas? It seems that IE8 is not supported in practice, even if it is in theory. (The mailing-list thread suggests no one cares about IE8, and there's evidence that IE8 has been broken for a while.) For node.js, does qunit even work in node.js? I've only ever tested in browsers with canvas, so I haven't thought about it. :-D

starwed:
Yeah, if you check out the grunt file in #472, you can run all the current qunit tests in core.html from the commandline.  I haven't actually tried the ones I added here.

There are way too many places that require a hand maintained list of files!   I'll fix crafty-local, but hopefully we can come up with a better solution.

_e:_ Ok, done.  I also squashed the commits down to two basic steps.

starwed:
Ok, merging.  While this is a large patch, most of it is additional functionality and certainly won't break anything.

The exception is the refactoring of the drawing routine -- and I think if that broke anything it would be _really_ broken.

sbyrnes321:
@starwed -- FYI, when you run [qunit tests using grunt in nodejs](https://npmjs.org/package/grunt-contrib-qunit), it DOES have canvas. Nodejs doesn't have canvas by itself, but the tests run in [phanomjs](http://phantomjs.org/) which does have canvas.

starwed:
@sbyrnes321  Thanks, that's good to know!

PR #485, opened 2013-06-10
Add resize event and change hitbox on resize

starwed:
This patch:
- Adds a resize event to "2D" triggered when h or w change
- Allows "Collision" hitboxes to track an entities shape through that event
- Changes collision(polygon) to rotate the polygon before attaching it
- Fixes some errors in calculating the MBR when a rotated entity was resized
- Adds some tests for resizing rotated objects

The changes to "Collision" are not backwards compatible, but they make it act how almost everyone _expects_ it to act.  

Currently, when you resize a rotated entity, the MBR size is adjusted as if it weren't rotated.  To fix this, I factored out the calculation of the MBR into a separate function, and call it on a resize.

The DebugCanvas changes I made in a separate PR were vital to getting this working!  Much easier to catch errors when you can just visually inspect the hitboxes and MBR

_e:_ Quick note: I had to rebased this against develop after opening the PR

### DIFF
--- a/src/2D.js
+++ b/src/2D.js
@@ -330,50 +330,73 @@ Crafty.c("2D", {
 		});
 	},
 
+
 	/**
-	* Calculates the MBR when rotated with an origin point
+	* Calculates the MBR when rotated some number of radians about an origin point o.
+	* Necessary on a rotation, or a resize (when already rotated)
+	*/
+
+	_calculateMBR: function(ox, oy, rad){
+		if (rad == 0){
+			this._mbr =null;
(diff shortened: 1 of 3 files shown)


sbyrnes321:
Looks very nice to me

starwed:
Ok, merging.  This should only break (a) people who accidentally relied on "Collision" having 0 size, and (b) people who used `.collision()` to assign polygons to rotated entities.  (If anyone really did the second one, they can simply rotate the polygon before attaching.)

PR #486, opened 2013-06-11
allow a scene to invoke another scene by returning a string

sbyrnes321:
(in response to discussion at https://groups.google.com/d/msg/craftyjs/PfwKt3owZ5c/w1Jur8d63g0J )

[I haven't tested this yet.]

### DIFF
--- a/src/drawing.js
+++ b/src/drawing.js
@@ -222,6 +222,16 @@ Crafty.extend({
 	* If you want some entities to persist over scenes (as in, not be destroyed)
 	* simply add the component `Persist`.
 	*
+	* Using `Crafty.scene()` is always optional, but some programmers choose
+	* to write games that spend most of the time inside the `init` function of one
+	* scene or another.
+	*
+	* It is bad practice to open a new scene (with `Crafty.scene('foo')`) when you are
+	* inside the `init` function of another scene. (The call stack will grow and grow
+	* if you keep doing that.) Instead, use `return 'foo'`: Whenever an `init` function
+	* returns something with string type, it is interpreted as a new scene to run
+	* immediately.
+	*
 	* @example
 	* ~~~
 	* Crafty.scene("loading", function() {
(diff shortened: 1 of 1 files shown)


starwed:
Hmm, if you use a more complex structure where you call a scene from outside of the initial function, it won't be possible to make use of this approach.

Is there something else we could do?  I've never thought of this issue before, but if you wanted to use lots of scenes it might be significant in long running games.  In other words, is there some way we can return from a scene when it's over from _any_ point within the scene?  Nothing comes to mind!

starwed:
Actually, after thinking about it a bit more, and looking at how my own code works, this isn't too big of a problem.

There shouldn't be any game loop stuff in a scene function.  The typical use of a scene should be to set up a bunch of components, bind any events to things like "EnterFrame", and then return.  All my scene changes except the first ultimately flow from "EnterFrame" or other events, and those flow from the main game loop.  So the nesting typically shouldn't be that bad.

Given that, it's hard for me to understand the use case of this patch.  The example given in the google groups thread was:

```
var function_A = function(){
    ...
    if(something happened){
        Crafty.scene("scene B");
    }
}
```

But no "time" elapses during scene setup, so "something" should never "happen".

Am I missing a common use for scenes?

sbyrnes321:
I agree that "There shouldn't be any game loop stuff in a scene function." But I get the impression from the forum that other people are not following that advice, and treating "scene" as a shell for entire levels. (Maybe that impression is wrong.) Maybe it would be better to just say in the documentation that using scene as a shell for entire levels is a bad idea? I dunno.

starwed:
Hmm, I think the particular post that started this was someone speculatively trying to get their head around how scenes work, rather than asking about an actual problem they had.  To recursively call scenes from other scenes, you'd basically have to write all your own game loop code.  I find it hard to believe that folk are actually doing that!  They might not really understand the control flow of their own code, of course.  :) 

Put it this way: if a scene adds any components, then at least one frame has to pass before they show up.  That'll happen _after_ the scene function has finished running, and so from that point any scene changes will actually be caused by triggered functions.  And you can't just, say, replace regular function calls with scenes all over the place, because a new scene will destroy the existing stage.

I could easily be misunderstanding the issue!  But I'd want to see "proof" that this is an actual problem first, i.e. code in the wild that follows this anti-pattern, or some part of the docs that are confusing.

sbyrnes321:
point well taken

PR #489, opened 2013-06-14
Fix within and contains methods

starwed:
These methods test whether the entity is within or contains an entity.  For a rotated entity, we should test against the MBR.  I mentioned this in the docs, and added some tests.

intersect was considering the MBR already; I tweaked the variable names to match within/contains and added to the docs.

The problem with within was mentioned on the forums: https://groups.google.com/forum/?fromgroups#!topic/craftyjs/8m0NKKRjZ6Q


starwed:
Pretty low risk, so I'll just merge right away.

PR #490, opened 2013-06-14
Timing events

starwed:
This is a pretty simple patch that measures how long the various stages of the game loop take.  It uses this info to
- On "EnterFrame", pass along the time since the last frame occurred.  I was able to use this to have physics run in a time based rather than frame based manner.  
- Trigger some events: "MeasureFrameTime", "MeasureRenderTime", and "MeasureWaitTime" that let you measure and benchmark how long each stage takes.  Waiting is the time between the end of one step and the beginning of the next.

I didn't change the internal logic of the game loop at all; it will still render multiple frames to "catch up" to the desired frame rate.  For the deltaT info to be more useful, it might be nice to have that behavior be customizable.

I was able to tweak one of my games pretty easily to use the dt in its physics; it did seem to run a lot more smoothly.  Long term, if we land something like this we might want to let the various animation components take advantage of it.

I haven't added any docs or tests yet, just wanted to see what people thought.    The changes themselves are really simple.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -32,7 +32,7 @@
     },
 
     GUID, FPS, frame, components, entities, handlers, onloads, tick, requestID,
-	noSetter, loops, milliSecPerFrame, nextGameTick, slice, rlist, rspace,
+	noSetter, loops, milliSecPerFrame, nextGameTick, slice, rlist, rspace, lastFrameTime, endTime, currentTime
 
 	initState = function () {
     	GUID = 1; //GUID for entity IDs
@@ -57,6 +57,9 @@
     	loops = 0;
     	milliSecPerFrame = 1000 / FPS;
     	nextGameTick = (new Date).getTime();
+        currentTime = lastFrameTime = nextGameTick;
+        maxTimestep = 40;  // The goal is about 20ms per time step, but drawing interferes!
+        endTime = 0;        
 
     	slice = Array.prototype.slice;
     	rlist = /\s*,\s*/;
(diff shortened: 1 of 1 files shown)


kevzettler:
:+1:  

Awesome! glad to se `dt` being added

starwed:
I've now updated the docs.  Does anyone have an opinion on better names for "MeasureRenderTime" etc? 

I'll go ahead and merge this soon-ish if no one expresses any doubts about the patch.

kevzettler:
Wait so does this mean `dt` is not available for use in components? like `my_component.bind("EnterFrame", function(dt)})`?

starwed:
It is definitely available for use in components!  :)  The EnterFrame event passes an object which contains the dt value.  So a function bound to "EnterFrame" could be something like:

```
function moveEveryFrame(data){
    this.x += this.vx * data.dt
}
```

This will all be explained in the "EnterFrame" documentation -- you can look at the comments in the patch, or build the docs yourself if you want a preview.

kevzettler:
Alright thats awesome. I think the `Measure` names are fine they are clear and explain what they do.

starwed:
Ok, merging this now.  I'm working on a followup patch that will give you more control over the game loop -- for instance, being able to force one frame per render would be useful if you're going strictly off of the dt.

Another follow-up we should consider is allowing any animation events to work off of a time instead of a number of frames.  Having a component that automatically handled the conversion might be the way to go.

kevzettler:
I agree on the Animation time suggestion. I find it odd that you pass it a number of frames now. I'm working on a multiplayer game that these updates will be huge wins for. Great work @starwed

PR #491, opened 2013-06-14
Fix delay docs

starwed:
Delay was given the wrong component name in the documentation.

Fix that, and clarify the example.


starwed:
Change to the docs only -- merging.

PR #493, opened 2013-06-19
Set _current scene before calling init function

starwed:
Probably makes more sense to set the current scene _before_ running its init function.  This helps prevent surprises if a scene is called directly from another scene's init function!

### DIFF
--- a/src/drawing.js
+++ b/src/drawing.js
@@ -205,7 +205,8 @@ Crafty.extend({
 	/**@
 	* #Crafty.scene
 	* @category Scenes, Stage
-	* @trigger SceneChange - when a scene is played - { oldScene:String, newScene:String }
+	* @trigger SceneChange - just before a new scene is initialized - { oldScene:String, newScene:String }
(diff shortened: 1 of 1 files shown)


sbyrnes321:
I agree 100%. You'll see that reordering it this way [occurred to me too as a good idea](https://github.com/craftyjs/Crafty/pull/486/files). You may want to move the SceneChange trigger up for the same reason, but either way is OK.

starwed:
Yeah, once I figured out what was going on I immediately thought of your earlier patch.  :)

I'm not really sure what the use case is for the "SceneChange" event, so that makes it hard for me to judge whether it should occur before or after the init function.  It's not used anywhere in Crafty internally.

starwed:
Ok, I went ahead and made the "SceneChange" event trigger before the new scene is inited.  I also went ahead and added a "SceneDestroy" event that triggers right before the entities in a scene are destroyed -- might be useful to run cleanup/etc functions while all the entities are still there.

_e:_ Oh, I also changed the module name (for the docs) from "Internal" to "Game Loop", to make it clear what it is.  :)

PR #496, opened 2013-06-27
Adds support for using Crafty as a CommonJS module such as in browserify

00dani:
There was already support for use as an AMD module, and adding CommonJS support isn't a particularly big addition, so why not?


starwed:
Seems fine to me, though I'll wait a day on merging in case someone wants to chime in.

I wonder how the various third-party components play with module systems.

PR #497, opened 2013-06-27
Adds a bower.json package description

00dani:
Crafty is registered as a Bower component but previously lacked such a file, which makes it harder for tools that consume Bower components to work with it.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,7 +1,7 @@
 {
   "name": "crafty",
   "main": "crafty.js",
-  "version": "0.5.3",
+  "version": "0.5.4",
   "repository": {
     "type": "git",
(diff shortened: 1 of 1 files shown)


starwed:
One correction to make: The current version is 0.5.4.

I have never used bower, so I'll just trust you that everything else works.  :)

_e:_ Oops, attached the comment to the wrong bug originally.

00dani:
Whoops. Version fixed.

Aside from that, though, it does indeed work fine (at least how I'm using it!).

starwed:
Ok, merging.  

Aside: There are so many places the version number needs bumping I think we should probably make that part of the build process!  :)

Issue #498, opened 2013-06-28
Text component for Canvas element doesn't redraw correctly without a width or height


kcaze:
When I have an entity with a Sprite component, then calling text() fails to update any Canvas entities with Text components (DOM seems to work fine).

Here's a fiddle illustrating the problem: http://jsfiddle.net/kcaze/x8t2B/1/

starwed:
Ah, I think the issue is actually that the text entity isn't redrawn correctly without a height or width.

When it is the only element, a full redraw is done, so the text changes correctly.  When you add another canvas  element (it doesn't matter what type!) a partial redraw is done instead.  Because the text entity lacks dimensions, it is skipped during this process -- and that's the bug.

I'll work on a fix for that, but a workaround in the meantime might be to give the canvas text a width/height.

kcaze:
As a workaround, in the text() function, I obtain the canvas's context with Crafty.canvas.context and then set the width using that. The height is simply the font size.

Another thing I've noticed is that the draw() method for the text draws the text at (obj._x, obj._y + obj._h). I don't quite understand why the text would be drawn at the y position plus the height, rather than simply at the y position. It seems like the clearing of the dirty rectangles in the DrawManager doesn't clear it properly if the text is drawn at (obj._x, obj._y) though.

EDIT: I realized that adding the height was necessary in order to compensate for the fact that the textBaseline is the bottom of the text by default and not the top. In my pull request, I've fixed that, and had text() update the width and height each time it's called.

starwed:
Should be fixed by #620.

PR #499, opened 2013-06-28
Fixes bug with text component for Canvas not being redrawn

kcaze:


starwed:
This removes the ability to set a font size in something other than pixels, which we should allow if it is possible.

Getting the height of the text can also be more complicated than just getting the font size in any case.  Glyphs with descenders can hang below that distance, if I understand how font sizes work.

Aside: if, as in this PR, we switch how `fillText` is called, we might be able to get rid of the save/restore of the context, which would slightly help performance.

kcaze:
I don't think there's really a good way of measuring the height of text. The simplest method (although ugly) would probably be the Matt Palmerlee's answer here: http://stackoverflow.com/questions/1134586/how-can-you-find-the-height-of-text-on-an-html-canvas.

Also, can this issue and the other one be merged somehow?

starwed:
Creating a canvas element and measuring the text every time it changes would have way too big a performance impact!

I'm sure we can [estimate the height](http://reeddesign.co.uk/test/points-pixels.html) for several different types of units without resorting to anything so drastic.  :)

kcaze:
As you mentioned earlier though, getting the height isn't as simple as the font size, since glyphs can descend arbitrarily low below the baseline (http://stackoverflow.com/questions/5718299/font-size-what-can-i-be-sure-of).

Also, I'm not creating a canvas element each time, I'm just creating a span. I'm not sure if this really has a significant impact on performance, but if it does, an alternative could be to create a single span to be used for measuring text height and position it off screen.

starwed:
@kcaze Ah, sorry, somehow misread your code!  I think creating and destroying hundreds of spans per second would probably still be bad for performance, though.

I think the best solution would be to, by default, estimate the height based on the font size, but to also allow the user to override this by calling a `setHeight(height)` method.  Without arguments it would attempt to directly measure the height of the text using the technique you suggest.  (Or maybe with some special argument, with no arguments reverting to the default height?)

kcaze:
That seems like a reasonable solution. Should the estimated height be based on the page you linked (<a href= "http://reeddesign.co.uk/test/points-pixels.html">http://reeddesign.co.uk/test/points-pixels.htm</a>l)?

A bit of leeway should be given though so that the estimated height is slightly greater than what's listed on that site.

kevinsimper:
@starwed Do you still think that what you mention is the best solution. If yes, i will be glad to submit a pull request on that.

starwed:
I had some trouble figuring out the best heuristic to use.  Probably best to be conservative with that?  We just  need something that works the majority of the time with common fonts, and is manually adjustable to account for outliers like international scripts.

starwed:
Here's a sample function we could use to determine the pixel size from the font-size string.  I haven't in any way tested it, just presenting an approach we could use:

``` javascript
    // takes a CSS font-size string and gets the height of the resulting font in px
var getFontHeight = (function(){
        // regex for grabbing the first string of letters
    var re = /\b(.+?)\b)/;
    // From the CSS spec.  "em" and "ex" are undefined on a canvas.
    var multipliers = {
        "px":1,
        "pt": 4/3,
        "pc": 16,
        "cm": 96/2.54,
        "mm": 96/25.4,
        "in": 96,
        "em": undefined, 
        "ex": undefined
    };
    var fudge = 1.1;
    return function (font){
        var number = parseFloat(font);  
        var match = re.exec(font);
        var unit = match[1];
        if (multipliers[unit] !== undefined)
            return Math.ceil(number * fudge * multipliers[unit]);
        else
            return Math.ceil(number * fudge);
    }
})();
```

toastino:
Hi! Has there been made any further progress on this issue? Because i've experienced the same behavior. Using kcaze's attempt on solving this problem, i noticed the following behavior: It fixes the problem illustrated in the fiddle in https://github.com/craftyjs/Crafty/issues/498 jet there still seems to be a problem with the redrawing of larger fonts. Here http://jsfiddle.net/Lg3Bx/1/ is a test similar to the old one but this crops the redrawn text strangely.(crafty compiled from kcaze changes) So the calculated width seems to be off by a bit as well.

starwed:
@toastino  For now I would suggest just using DOM text.  I'll start working on a patch for Canvas text, but I think no matter what, the DOM approach will always have advantages.

starwed:
Ok, got something working.  Modified fiddle [here](http://jsfiddle.net/Lg3Bx/5/).

Crafty was adjusting the width of text _after_ drawing it.

starwed:
Closing this due to #620.

toastino:
@starwed Ahh ok. I was working on a fix myself! That saved me a lot of time, thank you!

PR #500, opened 2013-06-30
Add unique and onetime binds

starwed:
This adds two types of special binding methods.  I don't like the current names, though!  :)
- `. uniqueBind()` prevents the same event from being bound more than once
- `.one()` binds an event, and removes it once triggered.

Just as we have `Crafty.bind()` and `entity.bind()`, there are two versions of these.  I also added some basic qunit tests for binding/triggering them.

TODO: The current implementation is very straightforward (just exactly what I found myself doing in my own code) but could certainly be optimized.   Also, there are probably places in the base code where we should switch to using these.  I just wanted to get the basic functionality landed.

**_About the names**_
- `.one()` is what jQuery uses for the same functionality, so I copied them.  It's short, but I don't think it's a very intuitive name.
- `uniqueBind()` is ungainly, but at least describes what it does.

If you can come up with some better names, please let me know.  :)

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -553,6 +553,46 @@
         },
 
         /**@
+        * #.uniqueBind
+        * @category Crafty Core
+        * @sign public Number ..uniqueBind(String eventName, Function callback)
+        * @param eventName - Name of the event to bind to
+        * @param callback - Method to execute upon event triggered
+        * @returns ID of the current callback used to unbind
+        * 
+        * Works like Crafty.bind, but prevents a callback from being bound multiple times.
+        * 
+        * @see .bind
+        */
+        uniqueBind: function (event, callback){
+            this.unbind(event, callback)
+            this.bind(event, callback)
+
+        },
+
+        /**@
+        * #.one
+        * @category Crafty Core
+        * @sign public Number one(String eventName, Function callback)
+        * @param eventName - Name of the event to bind to
+        * @param callback - Method to execute upon event triggered
+        * @returns ID of the current callback used to unbind
+        * 
+        * Works like Crafty.bind, but will be unbound once the event triggers.
+        * 
+        * @see .bind
+        */
+        one: function (event, callback){
+            var self = this;
+            var oneHandler = function(){
+                callback();
+                self.unbind(event, oneHandler);
+            }
+            return self.bind(event, oneHandler);
+
+        },
+
+        /**@
         * #.unbind
         * @comp Crafty Core
         * @sign public this .unbind(String eventName[, Function callback])
@@ -1156,6 +1196,47 @@
 
             if (!hdl.global) hdl.global = [];
             return hdl.global.push(callback) - 1;
+        },
+
+
+        /**@
+        * #Crafty.uniqueBind
+        * @category Core, Events
+        * @sign public Number uniqueBind(String eventName, Function callback)
+        * @param eventName - Name of the event to bind to
+        * @param callback - Method to execute upon event triggered
+        * @returns ID of the current callback used to unbind
+        * 
+        * Works like Crafty.bind, but prevents a callback from being bound multiple times.
+        * 
+        * @see Crafty.bind
+        */
+        uniqueBind: function (event, callback){
+            this.unbind(event, callback)
+            this.bind(event, callback)
+
+        },
+
+        /**@
+        * #Crafty.one
+        * @category Core, Events
+        * @sign public Number one(String eventName, Function callback)
+        * @param eventName - Name of the event to bind to
+        * @param callback - Method to execute upon event triggered
+        * @returns ID of the current callback used to unbind
+        * 
+        * Works like Crafty.bind, but will be unbound once the event triggers.
+        * 
+        * @see Crafty.bind
+        */
+        one: function (event, callback){
+            var self = this;
+            var oneHandler = function(){
+                callback();
+                self.unbind(event, oneHandler);
+            }
+            return self.bind(event, oneHandler);
+
         },
 
         /**@
--- a/src/core.js
+++ b/src/core.js
@@ -584,8 +584,8 @@
         */
         one: function (event, callback){
             var self = this;
-            var oneHandler = function(){
-                callback();
+            var oneHandler = function(data){
+                callback.call(self, data);
                 self.unbind(event, oneHandler);
             }
             return self.bind(event, oneHandler);
@@ -622,8 +622,7 @@
                 for (; i < l; i++) {
                     current = hdl[this[0]];
                     if (current[i] == callback) {
-                        current.splice(i, 1);
-                        i--;
+                        delete current[i]
                     }
                 }
             });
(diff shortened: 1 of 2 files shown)


jumpcoder:
I think we can also extend `.bind` to specify the event will be triggered how many times.
For example:

```
.bind('EventName', callback, times);
```

And the `.one` will be the shortcuts of `.bind('EventName', callback, 1)` .
Is there any plan to implement this?

starwed:
I've often wanted the equivalent of `one` while using Crafty -- I've never needed the variant you propose.  Is this something that has come up for you?  

While it would be pretty trivial to add, it might constrain _other_ enhancements to `bind` in the future.  For that reason I'd be hesitant to add this ability unless it comes up often enough to justify the increase in complexity.

sbyrnes321:
I definitely will be using these!

Is it OK to call `callback()` instead of `callback.call(self)` or something like that? (It might be a stupid question, I didn't look carefully.)

mpetrovic:
No. There is a difference between them.

callback() will use 'window' as 'this', while callback.call(self) will use
'self' as 'this.

sbyrnes321:
@mpetrovic -- Yes, I understand there's a difference. I was suggesting that @starwed might have made a mistake by putting `callback()` in the proposed code (lines 588 and 1235), rather than `callback.call(self)` or something. I'm not 100% sure. Maybe the test code should have another test like:

```
var e = Crafty.e("Triggerable")
e.xyz = 0;
e.one("Increment", function () {
   this.xyz++;
});
e.trigger("Increment");
e.trigger("Increment");
strictEqual(e.xyz, 1, ".one fired once");
```

This matches with a typical use and would resolve all doubts about whether the code is treating `this` correctly :-D

sbyrnes321:
Another test to add:

```
var temp = Crafty.e('blah');
temp.xyz = 0;
temp.abc = 0;
temp.def = 0;
temp.one('Event A', function () {this.xyz++; });
temp.bind('Event A', function () {this.abc++; });
temp.one('Event A', function () {this.def++; });
temp.trigger('Event A');
temp.trigger('Event A');
strictEqual(temp.xyz, 1);
strictEqual(temp.abc, 2);
strictEqual(temp.def, 1);
temp.destroy();
```

I think it will fail, because the iteration in `trigger` gets screwed up by the fact that the `handlers` object is being modified in the middle of the iteration.

I could be wrong. And if I'm correct, I'm not sure what the solution is...

starwed:
@sbyrnes321 You're right on both counts -- callback should be invoked with the self context, and removing a callback mid-trigger messes things up.  Luckily it's not too hard to fix the trigger problem: just go through the array of handlers in reverse order.  (I notice that this is a fix we've had to do in several different contexts!  I guess if order doesn't matter, you should always iterate through arrays "backwards" to avoid these problems.)

Now, this might break code that relies on the _order_ of callbacks, since they'd no happen in exactly the opposite order.  But I don't think the core of crafty relies on that.

_e:_ Also, I somehow totally forgot to have `one` pass in the data from the event...

starwed:
Ok, I added a commit that fixes those issues.  

A little nervous that changing the order of triggers might mess someone up, though.  In particular, user bound events would now trigger _before_ the core crafty handlers.  I never needed to rely on this in my own code, but I wouldn't be surprised if it's the sort of change that would cause folk hard-to-find bugs.

Any ideas on how to fix this without messing with trigger order?

sbyrnes321:
Do some people have trigger-order-sensitive code? Yes. The code will break, it's true. On the other hand, such code was quite problematic in the first place. So breaking it is not necessarily an awful problem.

Is there an alternative? Sure...

(1) `unbind` could delete an entry from `handlers` without splicing, e.g. `delete handlers[trigger][entity][i]`. It would leave a hole in the array (an entry which is `undefined`).
(2) `bind` could loop through the array looking for the first open spot (where the entry is `undefined`) instead of always appending at the end. That would keep the array mostly full in most cases, which is helpful for the `trigger` iteration.
(3) `trigger` would check each entry before running it to see if it's `undefined`.

In terms of speed and code clarity, I doubt there will be a big difference either way. Maybe this strategy is slightly inferior insofar as triggering occurs much more frequently than binding and unbinding so even a few holes in the handlers array is too many. Nah, it's probably no difference. An advantage of this strategy is that we would once again have the option of having `bind` return an index rather than returning `this`. (I'm not sure if we want to do this, but it's an option.)

Personally, I would vote for backwards-trigger, but either way is OK with me...

starwed:
Oh, the obvious way would be to just add new handlers to the beginning of the array instead of to the end (instead of  `push` use `unshift`,) thus preserving the order they're processed in.

While I'm tempted to **not** do that, since people really shouldn't rely on the order of bound events, it would probably be for the best to not change something so fundamental without good reason.  `unshift` would be a bit slower when there are lots of handlers in an array, but I can't imagine that would have a meaningful impact in most cases.

sbyrnes321:
@starwed -- Yes, using backwards-iteration in `trigger`, and add-to-beginning in `bind`, and splice in `unbind`, seems to be a good solution.

It keeps all triggering orders exactly the same as status quo. It's compatible with deleting an event binding in the middle of a trigger iteration.

OTOH, it's not compatible with _creating_ an event binding in the middle of a trigger iteration. (It would skip a possibly-unrelated callback.) But I think all of the proposals would have problems in this situation, and the "proper" behavior is undefined in this situation anyway (i.e. whether the new callback should run immediately or wait until the next trigger).

So your idea gets my vote. :-D

starwed:
> OTOH, it's not compatible with creating an event binding in the middle of a trigger iteration. (It would skip a possibly-unrelated callback.) But I think all of the proposals would have problems in this situation, and the "proper" behavior is undefined in this situation anyway (i.e. whether the new callback should run immediately or wait until the next trigger)

Hmm, even if there's no "best" way, I think not running an unrelated call-back probably is definitely wrong.  And it could actually be somewhat common to want to kind of "swap out" a handler on a triggered event.  I'll think about it a little more.  Grr. :(

starwed:
Ok, new suggestion that should be more consistent and still fairly simple.
- `unbind` deletes the array entry
- `trigger` checks whether `callbacks[i]` is undefined, and splices it out if so.
- Binding/triggering work in the order that they always have.

This avoids issues because arrays are mutated by the same code that iterates over them.

I can think of one remaining corner case: nested events.  If one "Move" event both unbinds a "Move" callback and triggers a _second_ "Move" event, things would not be guaranteed to work correctly.  (Because the _second_ trigger would remove the callback, messing up the first loop.)  That seems sufficiently perverse I'm not sure we need to handle it.

sbyrnes321:
Wow, really great idea!! :-D

starwed:
Hmm, _or_ we could have the callback arrays "cleaned up" once per game step.  That means the array entries would never shuffle around while they were being iterated over, because splice would only be called when no callbacks were in progress.

Possibly I'm over thinking this!  :)  But actually, this might be the sanest way.  It's pretty similar to your earlier suggestion, except we would never fill in holes.

sbyrnes321:
Scanning through the entire `handlers` object each frame is a potential drag on performance. Especially if you have a lot of entities that respond to a variety of rare events like "End of level 4".

So I prefer your previous idea of cleaning up during the course of triggering.

starwed:
Yeah, there are probably enough entries in the handler objects that it might have a real performance impact.

If it turns out that the corner case I mention causes real problems, I can think of a couple ways to work around it.  So I'll go ahead and implement that.  (Without the work-arounds for now.)

starwed:
So, we know that this will break in some odd corner cases, but we believe that this shouldn't matter in practice.

Best way to find out is to just merge it!

PR #501, opened 2013-07-02
Add the ability for components to run a special function on removal/destruction

starwed:
Allows components to define a `remove` method that work similarly to `init`.  It will be called when the entity is destroyed or the component is removed.  It's passed a boolean flag that is true on destruction, false on simple removal.

This patch also moves the trigger of "RemoveComponent" so that it occurs before the component is removed, instead of after.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -363,15 +363,19 @@
         * ~~~
         */
         removeComponent: function (id, soft) {
-            if (soft === false) {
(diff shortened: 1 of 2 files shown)


starwed:
Merging.  At some later point, might be worth going through the code and change existing components to use this new functionality.

PR #503, opened 2013-07-04
Changes to how _attr passes change information

starwed:
The changes are that it:
- Uses the actual object position instead of the MBR.  If you want to know how an object has moved, you want to compare its _actual_ position, not its MBR.
- Create a rectPool to improve performance by recycling rectangle objects.  Previously, a new object was created every time you set an object's position, which hurts performance.  I didn't want to simply store the info in an object attached to the entity, because it's conceivable a call to `_attr` could trigger a _second_ such call on the same entity.  So the `Crafty.rectPool` object creates a little pool of objects to recycle. 

I did a fair amount of benchmarking the rectPool optimization to make sure that it (a) didn't hurt average performance, and (b) really did reduce stutter.  I found the best metric was to look at the number of "EnterFrame" events per call to `step()`.  With a lot of objects moving around, it reduced this number across the board.

(This PR also has a commit to remove an accidental global in DeubgLayer.js.)

### DIFF
--- a/src/2D.js
+++ b/src/2D.js
@@ -1,3 +1,47 @@
+
+// Crafty._rectPool 
+//
+// This is a private object used internally by 2D methods
+// Cascade and _attr need to keep track of an entity's old position,
+// but we want to avoid creating temp objects every time an attribute is set.
+// The solution is to have a pool of objects that can be reused.
+//
+// The current implementation makes a BIG ASSUMPTION:  that if multiple rectangles are requested, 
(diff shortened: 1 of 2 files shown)


sbyrnes321:
Looks great to me, keep up the good work!!

starwed:
Merging.  This should not break anything.  No one should be depending on the currently broken behavior -- which in any case only occurs when you both rotate and rescale an entity.

PR #505, opened 2013-07-07
Make 2d comparison functions consistently use _x style properties

### LINKED ISSUE #504: Bug due to the fact that this._mbr uses x sometimes and _x other times
I have a bug where I can create an element where `this._mbr` has the properties `_x`, `_y`, `_w`, `_h`. But then it's used in `isAt`, which expects `x`,`y`,`w`,`h` (no underscores). This is a problem! Mouse events on this element do not get detected.

So far I can't reproduce this in a simple code snippet, nor can I give a more specific chain-of-events that causes this. It happens in my complicated game, and I'll post again if I figure it out more precisely. (Sorry I don't have time for now.) It might have something to do with the fact that my element is rotated. Also, I might have done something wrong in my code! But anyway, what I wrote in the previous paragraph sure looks like a bug in Crafty. :-D

FYI, I'm working off the very latest development branch (f9a66b047117baa2a9d5167f51a7a7bfdd3562f0)

starwed:
Modifies `isAt`, `contains`, and `within` to accept objects with `_x` style properties, such as MBRs.  (Fixes #504).  Would break code that passed manually constructed objects with old-style properties.

Adds a couple tests to better catch issues with MBRs and within/contains.

### DIFF
--- a/src/2D.js
+++ b/src/2D.js
@@ -444,7 +444,7 @@ Crafty.c("2D", {
 	* @param w - Width of the rect
 	* @param h - Height of the rect
 	* @sign public Boolean .within(Object rect)
-	* @param rect - An object that must have the `x, y, w, h` values as properties
+	* @param rect - An object that must have the `_x, _y, _w, _h` values as properties
 	* Determines if this current entity is within another rectangle.
 	*/
 	within: function (x, y, w, h) {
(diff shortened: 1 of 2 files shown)


sbyrnes321:
I can confirm that this fixes the bug I saw at #504. Looking good, thanks!

starwed:
Ok, thanks for confirming!

PR #506, opened 2013-07-08
Fixed comment error of Crafty.getVersion

jumpcoder:


starwed:
Looks good, thanks for the fix!

PR #508, opened 2013-07-10
Passing scene parameters

mikepaulson:
Modified the Crafty.scene to pass the second parameter received to the scene's init function when calling the init function (not when creating the scene).

### DIFF
--- a/src/drawing.js
+++ b/src/drawing.js
@@ -203,93 +203,93 @@ Crafty.extend({
 	_current: null,
 
 	/**@
-    * #Crafty.scene
-    * @category Scenes, Stage
-    * @trigger SceneChange - just before a new scene is initialized - { oldScene:String, newScene:String }
-    * @trigger SceneDestroy - just before the current scene is destroyed - { newScene:String  }
-    * @sign public void Crafty.scene(String sceneName, [Function|Object] obj[, Function uninit])
-    * @param sceneName - Name of the scene to add
-    * @param obj - Function to execute when scene is played or Object parameter to pass when initializing scene
-    * @param uninit - Function to execute before next scene is played, after entities with `2D` are destroyed
-    * @sign public void Crafty.scene(String sceneName, Object obj)
-    * @param sceneName - Name of scene to play
-    * 
-    * Method to create scenes on the stage. Pass an ID and function to register a scene.
-    *
-    * To play a scene, just pass the ID and an optional parameter. When a scene is played, all
-    * previously-created entities with the `2D` component are destroyed. The
-    * viewport is also reset.
-    *
-    * If you want some entities to persist over scenes (as in, not be destroyed)
-    * simply add the component `Persist`.
-    *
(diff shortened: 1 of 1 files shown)


sbyrnes321:
Looks good! Little suggestion about the documentation ... Here it is as in your commit:

```
* @sign public void Crafty.scene(String sceneName, (Function|Object) obj[, Function uninit])
* @param sceneName - Name of the scene to add
* @param init - Function to execute when scene is played
* @param uninit - Function to execute before next scene is played, after entities with `2D` are destroyed
* @sign public void Crafty.scene(String sceneName)
* @param sceneName - Name of scene to play
```

Here's what I think it should be:

```
* @sign public void Crafty.scene(String sceneName, Function init, [Function uninit])
* @param sceneName - Name of the scene to add
* @param init - Function to execute when scene is played
* @param uninit - Function to execute before next scene is played, after entities with `2D` are destroyed
* @sign public void Crafty.scene(String sceneName, [Parameter obj])
* @param sceneName - Name of scene to play
* @param obj - An optional parameter (number or string or object) that is passed to the init function
```

The idea is that there are two different ways to call the function (defining a scene vs running a scene), and the first four lines describe the first way to call it, and the last three lines describe the second way to call it.

Anyway, that's just a nitpick. This is a nice feature!

starwed:
Agree with sbyrnes321 about the docs & code!  Thanks for submitting the PR.

If possible, could you rebase your branch to squash it down to one commit?  Looking at old pull requests I've realized that removing unnecessary commits will make it easier to read through the history of the code.  If you're not familiar with git don't worry about it! 

 I'd be happy to walk you through it if you want, but an extra commit or two doesn't really matter.  :)

starwed:
@mikepaulson  Were you going to revisit this, or should I go ahead and make the doc changes myself? 

Thanks again for the code!

kevinsimper:
I tried to do it, but I can not seem to get it right.

I started with forking his repo, then `git rebase -i HEAD~3", then

```
pick 838c993 Added support for parameter passing to scene init.
s f2b83d7 Revert "Added support for parameter passing to scene init."
s 959f5a0 Added optional parameter to scene init.
```

But then i get this error:
Could not apply f2b83d7... Revert "Added support for parameter passing to scene init."

Why do i get that?

kevinsimper:
The documentation on this pull request should be improved. In the documentation he has changed, that it says it is a object, but he is passing a string.

Also i think the parameter should not be called obj, when it can be a function also. 

There should also be made a test for it.

starwed:
Also, all the small changes caused by fixing the lint issues make this require a fully manual merge, which makes me keep putting it off.  :)

I should probably just do a new PR from scratch.

potomak:
Ok, nice to know, I could make something about it after this deadline period!
Thanks @kevinsimper and @starwed

starwed:
I submitted a new PR, #619, that implements this idea.

PR #510, opened 2013-07-15
Develop

ghost:
Current bower.json causes a warning with bower install since the version in the file doesn't match the git tag.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,7 +1,7 @@
 {
   "name": "crafty",
   "main": "crafty.js",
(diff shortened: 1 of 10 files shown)


starwed:
Hi, looks like you were working off of the wrong branch -- you've included many extraneous commits in this PR. 

Also, we haven't actually upgraded the version info in Crafty yet; the `0.6.0-alpha` tag is just for a testing release.  I'm not really clear on what bower _does_, so please explain why this change matters.

PR #511, opened 2013-07-17
fix PHP warnings while running build-api.php

### LINKED ISSUE #475: PHP warnings while running build-api.php
Whenever I run build-api.php, I get the following notification:

"PHP Notice:  Undefined offset: 1 in /home/.../build/build-api.php on line 61"

The same message is printed ~40 times consecutively.

But the overall operation seems successful, insofar as I wind up with documentation that looks OK. Can the warning be ignored? Or is it a sign of a problem?

(I have no experience with PHP, so I can't help here.)

BradDenver:
this commit fixes #475

### DIFF
--- a/build/build-api.php
+++ b/build/build-api.php
@@ -58,7 +58,8 @@ function docs($files, $path, $save) {
 							break;
 						case PARAM:
 							$split = preg_split("/\s+-\s+/", $value);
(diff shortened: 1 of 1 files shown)


starwed:
Thanks for fixing this!

BradDenver:
Your welcome.

I saw this was an easy fix while looking at bountysource.

PR #512, opened 2013-07-17
Grunt integration

starwed:
This PR lets Crafty be built and tested with the grunt framework.  This work was started by @ukko in #472 (thanks!); I finished up the doc generation and tweaked some of the configuration.

If you have node installed, using this should  be pretty simple -- just checkout the branch and run `npm install`.  That should install all the node modules you need.   Then run `grunt <command>`.

The available commands are
- `concat` builds crafty.js.  (This is also the default)
- `api` builds the documentation
- `release` builds crafty.js and crafty-min.js, as well as the documentation
- `jshint` lints the project; currently this reports hundreds of issues
- `qunit` runs the core.html tests.
- `check` builds crafty.js, and then lints and tests it.

Would be nice if someone could test these before we land the patch.  (Though, I guess since the old build system is still there it won't matter too much.)
### api-gen

The new doc generator is written in coffeescript, which is compiled to js.  (I suspect there's a way to avoid including the compiled file.)  It produces nearly identical output to the old generator.  Some small changes like how method returns are displayed, and the actual HTML probably isn't identical.  

Some easy checks I included reveal several  places where documentation was broken (missing data necessary to be included in the docs.)  I'll submit a later PR request to fix all those.  :)


starwed:
Err, a stray mouse click merged this before I realized I wasn't updating the pull request after all.  Well, like I said, this at least won't break any existing functionality!  :)

sbyrnes321:
I compared the crafty.js built by grunt with the crafty.js built by php. They're close, but not identical...
- You forgot `fps.js` in package.json.
- build-crafty.php includes BOTH `animate.js` AND `animation.js` . package.json leaves one out. Is that on purpose? If so, we should delete the unused file altogether. If it's a mistake, can you (again) put it in the list in the same order as it is in build-crafty.js? That makes it easy to discover any other omissions.
- In the opposite direction, you remembered diamondiso.js but it's been missing from build-crafty.php! :-D

For all three of these, can we try to keep the same order of the file-list between build-crafty.php and package.json? That will make it easy to discover any other changes or omissions.

Incidentally, I hope package.json can be the master file-list from now on. We can even make crafty-local.js read the file list from package.json instead of having its own hard-coded file list. I'll make that change when I get a chance.
- `package.json` should use `craftyjs.com`, not `craftyengine.com` right?
- It's redundant to put in both `license.txt` and the banner. Here's the start of the generated `crafty.js`:

```
/**
 * crafty 0.5.4
 * http://craftyengine.com/
 *
 * Copyright 2013, Louis Stowasser
 * Dual licensed under the MIT or GPL licenses.
 */

/*!
* Crafty v0.5.4
* http://craftyjs.com
*
* Copyright 2010-2013, Louis Stowasser
* Dual licensed under the MIT or GPL licenses.
*/
```

Either of them could equally well be dropped.

I'm happy to do all of these changes if you don't. :-D

sbyrnes321:
Some other items:

`grunt qunit` does not pass when I run it. Does it pass for you? For me, it dies in DebugLayer - Hitbox debugging:

`TypeError: 'undefined' is not a function (evaluating 'e.polygon.points.toSource()')`

Another thing: A better way to handle the coffeescript might be to compile the `.coffee` file to js with grunt, run the js file, and then delete the js file with grunt. It could all be a single grunt command

https://npmjs.org/package/grunt-contrib-coffee
https://npmjs.org/package/grunt-contrib-clean

Then only the `.coffee` file needs to be saved in the repository, not the `.js` file. And there is no additional software required to build the docs, besides node / grunt. (Again I'm happy to make this change if you like. :-D )

[The compiled js file doesn't necessarily need to be deleted, it could also be .gitignored.]

Sorry if I'm misunderstanding anything.

starwed:
- Oops, I think I got confused because I had been _thinking_ about removing those, and so when I updated ukko's file list I was a bit too eager!   ("Animation" seems to be completely unused and undocumented, and "FPS" also seems kind of vestigial.)  I'll put them back for now.  Diamondiso was actually present all along in package.json; guess we should now check whether it works or not.  :)
- Yeah, using package.json as the master list for everything seems best.  I made it so all the grunt tasks use it for files, and  `crafty-local` could probably be modified to read from it.    Was there some place else that needed to know the files?  (It should also be possible to have the build script check that everything in `/src` is actually being used and log a warning if not.)
  - There are also several places in the code where we need to bump the version number -- it should be easy to automatically insert that when the files are built.  Using grunt really opens up a lot of possibilities, but I wanted to land the basic functionality first.
- Agree on the coffeescript; with the right nodules I think you can just point node at a coffee file and we don't even need to worry about cleaning up the .js file.
- The very last test fails for me when run from grunt, because of a silly use of the Gecko-only `toSource()` in one of the tests.  (It was just in there for debugging; I thought I'd already fixed it, but obviously not.)

I'll do a separate maintenance PR that fixes some of these issues: fixing some doc bugs, package.json, and the unit tests.  (Either tonight or tomorrow morning.)  If you want to jump on anything, go for it!

Oh, I also need to actually try cloning the repo and setting a dev environment up from a clean slate.

kevinsimper:
Cool that you implented Grunt! Thumbs up!

Is there a reason why you choose to write it in coffeescript?

starwed:
@Ukko started writing the parser in coffeescript, so it made sense to stick with that. 

I do tend to write all my own stuff in CS, though; the indentation style feels nicer to me.

PR #513, opened 2013-07-18
Misc cleanup

starwed:
This fixes numerous small issues either revealed or caused by the grunt integration.
- Several typos or missing tags in doc blocks
  - I didn't feel like fixing it at this time, but I noticed that the storage component needs to use the @trigger tag to document its events.
- Coffee modules can be loaded directly (No intermediate .js file needed)
- Removes silly bug in core.html test
- Deletes animate.js and fps.js, since they were unused/undocumented.
- Fixes a bug in the api generator that caused @return tags to not be processed.
- Added useful links and a quickstart for grunt to the README

If anyone thinks we should keep animate/fps, speak up now!  :)


starwed:
I added some instructions for grunt.  I guess it's not clear to me whether you still need `grunt-cli` after doing the local install.  (I'm also not sure what the state of things is on windows -- do we need separate instructions there?)

_e:_ Rebased to consolidate some commits.  Pretty sure you need `grunt-cli` installed globally, so added a wiki page that explains things.  The README now just outlines how the build system works and links to the main build page.

grozen:
@starwed I see that we now have some coffeescript in the repo? I imagine this is not to be used liberally?

PR #514, opened 2013-07-18
crafty-local should read file-list from package.json

sbyrnes321:
(maintaining separate file-lists is prone to errors when the file list is updated.)

### DIFF
--- a/crafty-local.js
+++ b/crafty-local.js
@@ -4,51 +4,45 @@
  */
 
 (function (window) {
-	var include = [
-		'core',
-		'intro',
-		'HashMap',
-		'2D',
-		'collision',
-		'DebugLayer',
-		'DOM',
-		'fps',
-		'html',
-		'storage',
-		'extensions',
-		'device',
-		'sprite',
-		'canvas',
-		'controls',
-		'animate',
-		'animation',
-		'drawing',
-		'isometric',
-		'particles',
-		'sound',
-		'text',
-		'loader',
-		'math',
-		'time',
-		'outro'
-	],
-	l = include.length, i, tr = new XMLHttpRequest(), output = '', url, base = '', scripts = document.getElementsByTagName('script');
+	var i, l, url, pkg,
+		tr = new XMLHttpRequest(),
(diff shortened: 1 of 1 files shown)


mpetrovic:
does (function () {}()) work? If so, do you need the outermost parens?

sbyrnes321:
Yes it works - I tested in firefox and chrome at least. I switched the order because jslint told me to, I didn't think it through too carefully. I found a discussion at

http://benalman.com/news/2010/11/immediately-invoked-function-expression/

which suggests that both of them work equally well, so it seems to me that jslint (Crockford) has an arbitrary preference. Oh well, whatever.

As for the outermost parentheses, the blog says they're required to avoid a syntax error.

starwed:
What is the correct way to test this?  

Any use of `crafty-local` always gives the permission error for me -- unless I spin up a local server.  Is this the intended use?  (I remember having problem with Chrome accessing the local file system, but Firefox always seemed willing.)

mpetrovic:
Yes, it's the intended use.

I run it with the file:// and turn off strict-origin-for-file-uris, or
whatever, in firefox when doing core work. If you can install a web server
locally and run it from there as well, but I don't want a web server on the
same machine I game on.

sbyrnes321:
I tested by running core.html. Actually if you open `core.html` you'll see the instructions for using crafty-local. Maybe I should copy them here too...

It works in chrome if you open chrome with the flag `--allow-file-access-from-files`.

It works in firefox if crafty-local is in the same directory or a subdirectory as the HTML file calling it. But not if it's in a parent directory, cousin directory, etc. [Unless you change a setting in about:config, `security.fileuri.strict_origin_policy` --> false.]

I don't know about other browsers.

starwed:
Ok, looks good to me.  (I just tested with `python -m SimpleHTTPServer 8080`).

The error message should probably mention the intended usage, but I'm not too worried about that.

mpetrovic:
The intended usage is for us. No one not doing work on the engine itself
should be using Crafty-local.js.

PR #516, opened 2013-07-27
Fix `multiway`'s behavior when dealing with zero-valued axis of `speed`.

mnemotic:
This fixes the behavior of `multiway` when `speed` is as an object
with 0 value for either axis. Passing `{x: 0, y: 0}`, or variations
of, is now safe.


starwed:
Looks good, thanks!

PR #518, opened 2013-08-05
storage fixes and change to trigger

chadams:
Found a few issues relating to storage. We getting an error in Chrome because it deprecated setVersion. Also, has to change core to support multiple parameters for data.

Check the commit comments to see details.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -653,8 +653,10 @@
         *
         * Unlike DOM events, Crafty events are exectued synchronously.
         */
-        trigger: function (event, data) {
+        trigger: function (event, data_n) {
             // (To learn how the handlers object works, see inline comment at Crafty.bind)
+            var args = Array.prototype.slice.call(arguments);
+            args.shift();
             if (this.length === 1) {
                 //find the handlers assigned to the event and entity
                 if (handlers[event] && handlers[event][this[0]]) {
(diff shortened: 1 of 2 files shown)


starwed:
Thanks for the PR!  

I haven't looked at the storage changes yet, but I'm uncertain whether trigger should change (there are other ways of fixing up save data.)  The current Crafty convention is to always pass an object.  This does have some nice features: the order of parameters doesn't matter, and it makes it very obvious what options are being passed in.  And anything you can do with multiple parameters, you can also do with a single object.

On the other hand, there are times when you want to avoid creating temporary objects just to pass to an event (performance!), and in these cases having multiple parameters would be nice.   Looking at jQuery, their trigger allows multiple parameters.

I'd really appreciate hearing what others think of the proposed change to trigger.

Aside: If we add extra parameters to an event, we need to ensure that the document generator supports it.

mpetrovic:
Storage is gonna be slow anyway, because it's writing to a database (i.e.
hard disk) every time. People shouldn't be using it often enough that
objects from that event will account for much garbage collection.

starwed:
Yeah, the change for trigger absolutely isn't need here.  And it's better if the way we pass data isn't tied to a particular implementation.

But allowing multiple parameters is certainly something that's passed through my mind before in other contexts.

kevinsimper:
I think this looks good, it is smarter that you can pass more parameters, instead of creating a temp object as you say starwed.

PR #519, opened 2013-08-05
fix for building in linux/unix

jjrh:
An extra dos character is included which causes npm install to fail.

ERR! Couldn't read dependencies.
ERR! Failed to parse json
ERR! Unexpected token ﻿

I simply ran dos2unix which removed this character.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,4 +1,4 @@
-﻿{
+{
     "name": "crafty",
     "version": "0.5.4",
     "title": "Crafty game framework",


starwed:
Seems like a pretty safe change, even if there's no apparent difference on the diff.  :)

I had tested it on OSX, so it apparently wasn't broken on all UNIX-like things.

sbyrnes321:
For the record, I believe the change was to remove the [UTF-8 BOM](http://en.wikipedia.org/wiki/UTF-8#Byte_order_mark). It is a relatively common bug for linux programs to not be able to parse the UTF-8 BOM. I [recently learned](https://github.com/sbyrnes321/numericalunits/issues/1) that this bug exists in some python versions for example. Anyway, removing the BOM is a good idea.

PR #520, opened 2013-08-07
fix broken qunit links in tests

sbyrnes321:
Replace the broken links by local copies of jquery and qunit. This allows developers to easily run the tests while they are off-line, or while jquery.com is off-line.


starwed:
Ok, looks good.  I spent a fruitless 15 minutes trying to figure out how I'd broken the unit tests, until I figured out that the files weren't loading.  Glad you have it already in hand!  :)

PR #522, opened 2013-08-07
Fixing a bug in centerOn for non-zero origins

ericathegreat:
When the viewport is not already at zero and .centerOn
is called, it was not correctly centering. The further
away from the origin the viewport was, the further away
from the center target it would end up.


starwed:
Looks good.

PR #523, opened 2013-08-07
Making other tests run as part of the Grunt task.

ericathegreat:
- A few tests still pointing to the jquery repo instead of the local copy
- Making sure that grunt qunit tests more than just core.


sbyrnes321:
This is great! (In case it's not clear: Right now, these tests are all passing. I tried it myself.)

starwed:
Thanks for the PR!

> Right now, these tests are all passing.

Good to hear!  I'd been keeping a close eye on core, but some of the others I hadn't ran through in a while.

PR #525, opened 2013-08-11
Viewport refactoring (round 1)

starwed:
This refactors the scroll and scale viewport functions.  They now simply update the appropriate viewport variables, and trigger an "InvalidateViewport" event.  (I went ahead and moved all the viewport stuff into its own file, since it was taking up almost all of "extensions.js".)

The rendering functions then make any adjustment necessary to their render target.  This should drastically improve perforamance, because I got rid of all unneeded calls to `drawAll()`.
- I fixed various and sundry bugs I noticed while implementing this, but I think there are probably many more.  Several of the viewport functions don't seem to consider the possibility that the viewport might be scaled.
- I also fixed a small bug that meant drawAll wasn't catching rotated objects if their MBR was onscreen but their base coordinates weren't.
- I tested scale and scroll, as well as follow, and all seemed to work fine.  I _think_ the other's should still work, since in most places they implicitly rely on scroll.  But no promises.  

This is just round one; several of the other viewport functions need refactoring, but this is the base work that should make that easier!


starwed:
Oh, I forgot to document the new events (and maybe some other stuff?)   I'll get to that tomorrow.

starwed:
Ok, added the docs and removed an extraneous copy of a line that had crept in there somewhere.  Anyone have any further thoughts before I land this for testing?  Easiest to see the changes by looking at commit 82a84c, since a file was added.

starwed:
I'm going to go ahead and merge this, and create a new alpha for testing.

PR #527, opened 2013-08-12
small grunt fixes

sbyrnes321:
The grunt-generated banner makes it unnecessary to include license.txt.

(Before this change, the non-minified file has essentially the same banner twice.)

Also, changed craftyengine.com to craftyjs.com, which seems to be the canonical address.

(license.txt could be deleted altogether, but only after the PHP build process is officially deprecated.)


starwed:
Both seem pretty reasonable.

PR #528, opened 2013-08-16
Customizable game loop

starwed:
This patch reworks the game loop to allow control over its behavior.  It can now be set to three modes, which control what `dt` is passed to each frame, as well as how often frames are called.  You set these by calling `Crafty.timer.steptype()`
- "fixed", which is the existing behavior.  Each frame processes the same amount of time, and as many frames as necessary are processed to reach the target rate.
- "variable", which processes a single frame per render.  This frame is handed the total time since the last frame, up to some maximum value.
- "semifixed", which takes the total time since the last step, and splits it evenly between as many frames are necessary to avoid exceeding some maximum timestep.

It also cleans up how the game loop and Crafty.timer are actually implemented, moving a bunch of internal variables into the scope of Crafty.timer.

While Crafty _works_ with this patch, there are a bunch of components that assume a constant time step per frame.  Animation, tweeing, and multiway are some of them.  I'm almost done with fixes for animation and tween, but figured the base changes should be reviewed separately.

The "fixed" mode should basically replicate the current behavior (with less tolerance for very long visual pauses), but since it's basically reimplemented there might be some subtle mistakes somewhere!

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -30,36 +30,22 @@
     var Crafty = function (selector) {
         return new Crafty.fn.init(selector);
     },
-
-    GUID, FPS, frame, components, entities, handlers, onloads, tick, requestID,
-	noSetter, loops, milliSecPerFrame, nextGameTick, slice, rlist, rspace, lastFrameTime, endTime, currentTime
+    // Internal variables
+    GUID, frame, components, entities, handlers, onloads,
+	noSetter,  slice, rlist, rspace, milliSecPerFrame
+    
 
 	initState = function () {
     	GUID = 1; //GUID for entity IDs
-    	FPS = 50;
-    	frame = 1;
 
     	components = {}; //map of components and their functions
     	entities = {}; //map of entities and their data
     	handlers = {}; //global event handlers
     	onloads = []; //temporary storage of onload handlers
-    	tick;
-
-    	/*
-		* `window.requestAnimationFrame` or its variants is called for animation.
-		* `.requestID` keeps a record of the return value previous `window.requestAnimationFrame` call.
-		* This is an internal variable. Used to stop frame.
-		*/
-    	requestID;
 
     	noSetter;
 
-    	loops = 0;
-    	milliSecPerFrame = 1000 / FPS;
-    	nextGameTick = (new Date).getTime();
-        currentTime = lastFrameTime = nextGameTick;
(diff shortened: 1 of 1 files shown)


kevinsimper:
Looks cool, good idea! I understand the code and that some times a frame can either take longer or shorter to process.

So "variable" would allow you to a variable framerate, cool. Then i do not fully understand the "semifixed" option, how that would affect crafty? What would that be good for?

starwed:
I chose the mode names based off of [this article](http://gafferongames.com/game-physics/fix-your-timestep/), though they don't quite match up.  (So maybe I should come up with new names for fixed, semifixed)

The essential problem is this: given a chunk of time since the last game step, we need to progress the game world by that amount of time and then render it. 
- "Variable" is exactly as you'd probably expect:  there is one "EnterFrame" event per "RenderScene" event.  "EnterFrame" gets passed a value for `dt` that indicates the actual time that has passed.  The problem is when `dt` gets too large, the physics might break down -- you can set a maximum value, but that would then lead to game slowdown.  The advantage of this mode is that you have exactly one "EnterFrame" per step through the game loop, and so you can avoid getting hit by slowdown if you're willing to have the physics get sloppy.
- "Fixed" is the way we used to do things.  (We didn't pass a `dt` value before, but the assumption was there.)  Each "EnterFrame" corresponds to a fixed amount of time, so it's always passed the same value of `dt`.   The time since the last step is split between however many frames it takes to consume the time.  The advantage of this mode is that game physics is 100% stable; you get the same results no matter the frame rate.  The disadvantage is it can require more total "EnterFrame" events, and motion won't appear as smooth due to the game clock alternately being ahead/behind of the real time.  (A consequence of consuming time in fixed chunks.)
- "semifixed" is kind of a hybrid between the two.  Given a maximum `dt`, it processes the time that's passed in however many "EnterFrame" events it will take to consume that time without surpassing the maximum `dt`.  That lets you get a lot of the advantages of "variable" and "fixed" while avoiding the very worst behaviors, but physics will not be completely deterministic, and a "death spiral" when the game gets processor intensive is still possible.

kevinsimper:
Okay, that explained it better, thanks.

The name could be "VariableMax"?

starwed:
I'm going to go ahead and merge this.  Still very open to changing the names in alter patch before release, but not a fan of 'variablemax'.  :)

Unfortunately I'm not sure how to write good unit tests for this, since the expected behavior is variable!

Issue #532, opened 2013-08-28
Buildsteps for api documentation with Grunt


kevinsimper:
What are the buildsteps for the `grunt api` command?

I know that you have to have installed grunt-cli:
Then you can do:
`npm install´ 
to get all the grunt tasks installed.

Then you need to compile the coffee-script file to js.

Then when you try running the command `grunt api` you get a lot of errors like these:
{ [Error: ENOENT, open 'build/api/Crafty-e.html'] errno: 34, code: 'ENOENT', path: 'build/api/Crafty-e.html' }

sbyrnes321:
Compiling the coffee-script file is not a separate step. If I understand correctly, grunt sees that the extension is .coffee and automatically interprets it as a coffee-script file.

Try deleting your compiled file (`api-gen.js` or whatever you called it) and then try `grunt api` again. Does it help?

kevinsimper:
I found the error.

It was because the folder /build/api/ did not exists, so when i created it all was fine.

How can that be avoided?

sbyrnes321:
You're right! That's a bug. Pull request #533 fixes it on my computer

kevinsimper:
Cool, that was nice solved.

PR #533, opened 2013-08-28
grunt should create API directory if it doesn't already exist

sbyrnes321:
Fixes issue #532


starwed:
Ah, I think I encountered this bug and then totally forgot about it!  Works for me on OSX in both cases (directory exists/doesn't exist), so merging.

PR #535, opened 2013-09-01
refactor .requires, documentation for .requires and .addComponent

sbyrnes321:
Now that addComponent does not re-add components that are already added (see https://github.com/craftyjs/Crafty/pull/428), it has become exactly the same as requires. Therefore I deleted the redundant code and clarified the documentation.

This does not preclude the possibility that requires and addComponent could be re-differentiated in the future (some vague proposals were mentioned in discussion at https://github.com/craftyjs/Crafty/pull/428 ). That's a different issue. In the code-base as it exists today, the two functions are synonyms, and we might as well make it easier for people to understand that.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -319,7 +319,11 @@
         * Makes sure the entity has the components listed. If the entity does not
         * have the component, it will add it.
         * 
-        * (This function is the exactly same as `.addComponent`.)
+        * (In the current version of Crafty, this function behaves exactly the same
+        * as `addComponent`. By convention, developers have used `requires` for
+        * component dependencies -- i.e. to indicate specifically that one component
+        * will only work properly if another component is present -- and used
+        * `addComponent` in all other situations.)
         * 
         * @see .addComponent
         */
--- a/src/core.js
+++ b/src/core.js
@@ -197,11 +197,13 @@
         * This means it will copy properties and assign methods to
         * augment the functionality of the entity.
         *
-        * There are multiple methods of adding components. Passing a
-        * string with a list of component names or passing multiple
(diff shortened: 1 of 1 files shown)


kevinsimper:
Looks like a good idea.

Now I am just asking, but shouldn't require be deprecated then some time in the future?

I will wait and see if somebody else have something to add, else I will merge it.

starwed:
> Now I am just asking, but shouldn't require be deprecated then some time in the future?

I'd say no, because they (should) have different semantic implications.  For instance, if component **A** requires component **B**, then probably **A** should be alerted when **B** is removed.

Along those lines, I'd actually prefer the documentation to _not_ claim requires and addComponent are the same.  Because they probably won't be in the future!

sbyrnes321:
@starwed -- The documentation can be changed in the future, just like the code can. So I don't see why you're worried. But if it helps, I could replace

"(This function is the exactly same as `.addComponent`.)"

with a more nuanced statement like

"(In the current version of Crafty, this function behaves exactly the same as `.addComponent`.)"

or

"(In the current version of Crafty, this function behaves exactly the same as `.addComponent`. By convention, developers have used `.requires` for component dependencies -- i.e. to indicate specifically that one component will only work properly if another component is present -- and used `addComponent` in all other situations.)"

I am open-minded to proposals to alter `requires`, and also open-minded to proposals to deprecate `requires`. But I think those should be separate discussions, for another day. :-D

starwed:
I like either of those replacements!

sbyrnes321:
I added a commit that edits the wording.

PR #537, opened 2013-09-02
Update core.js

matheusazzi:
Improve the condition to triple equal.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -912,7 +912,7 @@
         * ~~~
         */
         pause: function (toggle) {
-            if (arguments.length == 1 ? toggle : !this._paused) {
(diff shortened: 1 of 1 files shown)


sbyrnes321:
Well, the change doesn't make any difference, but I still like it and I hope it's accepted.

I am in the school of thought that `==` and `!=` are the "evil twins" of `===` and `!==` and should never be used (except in very rare specific circumstances). Every time I see `==` in javascript code, I have to stop and think "Could this cause strange behaviors due to type coercion? No? Good!" Whereas when I see `===`, I can understand the code immediately, without having to stop and think it through. :-D

matheusazzi:
Yes, I think the same way :D
Is always better use `===` against `==`, because make it more clearly.

PR #538, opened 2013-09-04
Fixing timing test based on game loop refactor.

ericathegreat:
This test was failing because the game loop refactor removed access to the timer's "currentTime" variable.
The current time can be accessed from the enterframe event though, so the test now uses these values and asserts that multiple frames have been entered in the duration.


kevinsimper:
I have tested it and it's perfect :+1:

PR #545, opened 2013-09-09
switch from getter to attribute-lookup

sbyrnes321:
attribute-lookups are faster than getters. We use attribute-lookups when possible everywhere in the code. This is a rare exception, presumably an accidental oversight.

I tested in a game with some rotating elements - it still works. The qunit tests also pass.


starwed:
Yeah, there is no reason to use .x and .y there.  I think I wrote that line, probably just put into the wrong mindset by the attributes of origin.

Issue #548, opened 2013-09-14
In some circumstances, canvas not being cleared properly


starwed:
As [reported by sulter](https://mail.google.com/mail/u/0/?shva=1#inbox/1410d711a25eace4) in the forum.

[This fiddle](http://jsfiddle.net/UmEn4/1/) reproduces the bug for me in both Firefox and Chrome.  

Probably expanding the cleared area by a pixel in each direction would fix this, but I'd like to understand _why_ it's happening.

starwed:
After looking through the drawing code, I notice that the coordinates of dirty areas are converted to integers when regions overlap, but aren't in other cases.  This would explain why the behavior appears when the entity moves quickly -- when moving slowly, the old and new position would overlap, but when moving quickly they are disjoint.  

_e:_ Yes, coercing all coordinates to integers seems to work; just need to figure out the best/fastest way to do this.  A workaround in existing versions of crafty would be to make sure `x` and `y` are always integers.  (I was doing that as a matter of habit when using crafty, which is why I never noticed this issue!)

starwed:
I think the solution will be to "round up" the dirty areas at the point that we render that region.  Currently there's some duplication of effort when areas get merged, and it's a bit messy to catch every case.  Far simpler to only round right before the point where it matters.

Sulter:
Thanks for taking it up starwed. I realized myself that the issue was trying to set the x and y to float points. For some reason I just assumed that would be handled by the engine. 
I think it would be reasonable to just warn the users in the API, like it's done with the rotation. At least for now.

starwed:
It certainly _should_ be handled by the engine!  :)  I'll try to land a fix for this tomorrow -- it's not very complicated, but I've had a vicious cold all week.

starwed:
Well, I fixed the issue of the canvas not clearing.

I haven't landed it due to a lingering issue with the MBR, though -- unfortunately when rotating by exactly 90 degrees we get a little bit of slop due to trig functions not quite being zero.  (`Cos(Math.PI/2)` !== 0, for instance -- it's something like 10^-17 instead.)  So a square rotated by 90 degrees ends up increasing in size by 1 pixel.  This then causes a couple of tests to fail.

Obviously I can just alter the tests, but thought I'd bring it up here.  It's annoying that a rotation by 90/180/etc degrees changes things it shouldn't, but is this a serious problem that's worth fixing?  Or just something to live with?

sbyrnes321:
[...]

PR #550, opened 2013-09-25
Fixes rounding of dirty rects and MBRs

starwed:
We want the areas of the screen that we clear and clip to have integer coordinates.  Prior to this patch there were some cases that were not caught.   The solution is to do this "rounding up" of regions directly before we use them, rather than as they are created.

Similarly, the MBRs used should have integer coordinates.  This fixes a regression that caused them to be "rounded down" in some cases, and also special cases rotations of 90 degrees.   Previously, taking (say) the cosine of PI/2 resulted in a very small nonzero value, which due to rounding caused some deviations of the MBR from an ideal 90 degree rotation.

This seems worth special-casing due to the privileged nature of 90 degree rotations.

### DIFF
--- a/src/2D.js
+++ b/src/2D.js
@@ -475,6 +475,9 @@ Crafty.c("2D", {
 
         var ct = Math.cos(rad),
             st = Math.sin(rad);
+        // Special case 90 degree rotations to prevent rounding problems
+        ct = (ct < 1e-10 && ct > -1e-10) ? 0 : ct;
+        st = (st < 1e-10 && st > -1e-10) ? 0 : st;
         var x0 = ox + (this._x - ox) * ct + (this._y - oy) * st,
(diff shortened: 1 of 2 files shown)


starwed:
This would fix issue #548.  

I almost introduced a nasty bug that redrew too much of the screen.  I've previously fixed a similar bug; they really hurt performance, but don't affect correctness, so they're hard to catch.  I previously added a debug flag[1] that lets you see which areas of the screen are being redrawn, but because everything happens internally, I'm not sure how to best unit test this.
1.  Set `Crafty.DrawManager.debugDirty = true` to view the regions rendered

sbyrnes321:
I'm not familiar with how this section of code works ... but I did fetch your changes, and my game is still displaying correctly and running smoothly.

The `debugDirty` thing is neat!

PR #551, opened 2013-09-27
corrected formula for calculating x and y in sprite function

motatoes:
As I understand it we need to include the padding for each cell when computing the x and y coordinates in a sprite. This is also what is done in the [sprite function in extensions.js](https://github.com/craftyjs/Crafty/blob/develop/src/extensions.js#L240) , but there are missing parenthesis in the sprite.js sprite() function

I Hope I'm not missing something here.

### DIFF
--- a/src/sprite.js
+++ b/src/sprite.js
@@ -93,8 +93,8 @@ Crafty.c("Sprite", {
      * The coordinate of the slide within the sprite in the format of [x, y, w, h].
      */
     sprite: function (x, y, w, h) {
-        this.__coord = [x * this.__tile + this.__padding[0] + this.__trim[0],
-            y * this.__tileh + this.__padding[1] + this.__trim[1],
+        this.__coord = [x * (this.__tile + this.__padding[0]) + this.__trim[0],
+            y * (this.__tileh + this.__padding[1]) + this.__trim[1],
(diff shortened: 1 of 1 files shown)


starwed:
Ok, that does look right.  The docs say that padding is used to specify the spacing between tiles.  The current code treats padding as if it were spacing inside a tile -- but that's what trim is for.

I'm curious if there's anywhere we rely on the current broken behavior, though.  There also might be some incorrect examples lurking somewhere.  Although I suspect that most people simply use the sprite function from extension.js, which is why this bug has been lurking for so long.

motatoes:
I'm not sure about if there is a use case for this but I think both of the formulas should be consistent to avoid broken behavior. I noticed this when I tried to modify the [getting started](http://buildnewgames.com/introduction-to-crafty) guide to add a default position for the sprite. So I modified [this line](https://gist.github.com/darrentorpey/3d848cc275e5a4ed6c4e#file-components-js-L32) to:

```
else {
    this.sprite(2, 2);
    this.pauseAnimation();
}
```

And then I saw that the sprite was not split correctly.

starwed:
I'll definitely land this patch -- just wanted to give folks an opportunity to mention other, related bits of the codebase if they knew of any.

starwed:
Merging.

Issue #456, opened 2013-05-06
Unit tests reveal many problems with IE8 (or at least, IE10 in "IE8 mode")


sbyrnes321:
Well, I have Windows 7, so I can't install IE8. but I can use IE10 with browser mode set to IE8 mode. Maybe someone else will check this with the real IE8.

When I take the latest crafty with the latest "core.html" unit-tests, they pass in firefox, chrome, IE10, and "IE10 in IE9 mode". (Except for the SAT test at the bottom, which always fails, probably the test is out-of-date.)

But for "IE10 in IE8 mode", many of the tests fail.
- The "All components - universal selector" test: "Expected: 7  Result: 8"
- The "CORE: setter" set of tests: "Died on test #1 undefined: Unable to get property 'push' of undefined or null reference"
- The "2D: position" tests: All of them fail except #5.
- The "2D: Intersect" tests 1 and 2 are failing ("Expected: true  Result: false")
- The "2D: child" tests 1-3 are failing (1. "Expected: 51 Result: 0". 2. "Expected: 52 Result: 0" "3. Expected: 50 Result: 0."
- The "2D: child_rotate" tests 2 and 4 are failing (2. "Expected: 35 Result: 15" 4. "Expected: 57 Result: 37")

Many of these, but not all, seem to come from the use of "+=" setters like entity.x += 5, which all seem not to work in IE8 mode. (I could be wrong.)

mpetrovic:
if += doesn't work in IE8, that is a huuuuuge problem.

sbyrnes321:
I further investigated core.html in IE8. So far, I haven't found "real" problems, only problems with the tests.
- `Crafty('*').length` is different in IE8 because there's a special entity (search for `ViewportSetter`) which only exists to update the viewport each frame. (This is a workaround to the lack of getter/setters.) For the same reason `Crafty('*').destroy()` should maybe be avoided. (There's nothing wrong with this, although in my mind it would be simpler to create a global event binding instead of a special entity.)
- Any test like `this.x = 5; console.log(this._x === 5)` will fail in IE8 because of the different getter/setter behavior. It should be modified to `this.x = 5; Crafty.trigger('EnterFrame'); console.log(this._x === 5);` This problem explains most or maybe all of the `2D` tests failing.
[...]

PR #553, opened 2013-10-03
fix unit-tests to avoid false-positives in IE8

sbyrnes321:
As discussed at https://github.com/craftyjs/Crafty/issues/456 , this pull-request fixes the core.html unit tests to reduce false-positives in IE8 (and also one in IE10). Some of the unit tests still fail in IE8, but I think the remaining ones are actual problems in the crafty code, not just bad tests. I ran the tests and they all pass in IE10, IE9 mode (in IE10), chromium, firefox, and grunt (phantomjs).

### DIFF
--- a/tests/core.html
+++ b/tests/core.html
@@ -42,11 +42,15 @@ $(document).ready(function() {
 		strictEqual( Crafty("test test2").length, 2, "Two components ANDed" );
 		strictEqual( Crafty("test, test2").length, 7, "Two components ORed" );
 		
-		strictEqual( Crafty("*").length, 7, "All components - universal selector" );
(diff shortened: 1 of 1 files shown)


starwed:
I don't really like the replacement of `Crafty("*").destroy()` with more specific versions -- it makes unit tests more fragile to modify.   (In fact, if possible, it should probably be run automatically after every module.)

Perhaps it would be better to modify the nature of ViewportSetter, rather than distorting how we run unit tests to deal with a special case.  Is there a compelling reason why it's a component?  Seems like it would be trivial to just bind the adjustment function to "EnterFrame" directly.

sbyrnes321:
OK, I'll change the viewport code and try again...

PR #556, opened 2013-10-07
Fix jshint warnings

potomak:
I run `grunt check` and I saw a lot of jshint warnings, I'd like to put fixes for them into this pull request.

Note: merge this pull request only after every warning has been removed.

### DIFF
--- a/src/controls.js
+++ b/src/controls.js
@@ -75,7 +75,7 @@ Crafty.extend({
             type = e.type;
 
         //Normalize button according to http://unixpapa.com/js/mouse.html
-        if (e.which == null) {
+        if (typeof e.which === 'undefined') {
             e.mouseButton = (e.button < 2) ? Crafty.mouseButtons.LEFT : ((e.button == 4) ? Crafty.mouseButtons.MIDDLE : Crafty.mouseButtons.RIGHT);
         } else {
             e.mouseButton = (e.which < 2) ? Crafty.mouseButtons.LEFT : ((e.which == 2) ? Crafty.mouseButtons.MIDDLE : Crafty.mouseButtons.RIGHT);
@@ -89,7 +89,7 @@ Crafty.extend({
             while (typeof (tar.id) != 'string' && tar.id.indexOf('ent') == -1) {
                 tar = tar.parentNode;
             }
-            ent = Crafty(parseInt(tar.id.replace('ent', '')));
+            ent = Crafty(parseInt(tar.id.replace('ent', ''), 10));
             if (ent.has('Mouse') && ent.isAt(x, y))
                 closest = ent;
         }
@@ -211,10 +211,10 @@ Crafty.extend({
         first.target.dispatchEvent(simulatedEvent);
 
         // trigger click when it should be triggered
-        if (lastEvent != null && lastEvent.type == 'mousedown' && type == 'mouseup') {
+        if (lastEvent !== null && lastEvent.type == 'mousedown' && type == 'mouseup') {
             type = 'click';
 
-            var simulatedEvent = document.createEvent("MouseEvent");
+            simulatedEvent = document.createEvent("MouseEvent");
             simulatedEvent.initMouseEvent(type, true, true, window, 1,
                 first.screenX,
                 first.screenY,
@@ -280,7 +280,7 @@ Crafty.extend({
             var prop = props[--i];
             evnt[prop] = original[prop];
         }
-        evnt.which = original.charCode != null ? original.charCode : original.keyCode;
+        evnt.which = original.charCode !== null ? original.charCode : original.keyCode;
         evnt.key = original.keyCode || original.which;
         evnt.originalEvent = original;
         e = evnt;
@@ -462,7 +462,7 @@ Crafty.c("Draggable", {
             var pos = Crafty.DOM.translate(e.clientX, e.clientY);
 
             // ignore invalid 0 0 position - strange problem on ipad
-            if (pos.x == 0 || pos.y == 0) {
+            if (pos.x === 0 || pos.y === 0) {
                 return false;
             }
 
@@ -484,7 +484,7 @@ Crafty.c("Draggable", {
         };
 
         this._onup = function upper(e) {
-            if (this._dragging == true) {
+            if (this._dragging === true) {
                 Crafty.removeEvent(this, Crafty.stage.elem, "mousemove", this._ondrag);
                 Crafty.removeEvent(this, Crafty.stage.elem, "mouseup", this._onup);
                 this._dragging = false;
@@ -522,7 +522,7 @@ Crafty.c("Draggable", {
     dragDirection: function (dir) {
         if (typeof dir === 'undefined') {
             this._dir = null;
-        } else if (("" + parseInt(dir)) == dir) { //dir is a number
+        } else if (("" + parseInt(dir, 10)) == dir) { //dir is a number
             this._dir = {
                 x: Math.cos(dir / 180 * Math.PI),
                 y: Math.sin(dir / 180 * Math.PI)
--- a/src/core.js
+++ b/src/core.js
@@ -43,10 +43,6 @@
             handlers = {}; //global event handlers
             onloads = []; //temporary storage of onload handlers
 
-            noSetter;
-
-
-
             slice = Array.prototype.slice;
             rlist = /\s*,\s*/;
             rspace = /\s+/;
@@ -240,7 +236,7 @@
             //extend the components
             ul = uninit.length;
             for (; c < ul; c++) {
-                if (this.__c[uninit[c]] == true)
+                if (this.__c[uninit[c]] === true)
                     continue;
                 this.__c[uninit[c]] = true;
                 comp = components[uninit[c]];
@@ -1296,7 +1292,7 @@
                 // Check whether h needs to be processed
                 if (!hdl.hasOwnProperty(h)) continue;
                 callbacks = hdl[h];
-                if (!callbacks || callbacks.length == 0) continue;
+                if (!callbacks || callbacks.length === 0) continue;
 
                 //if an entity, call with that context; else the global context
                 if (entities[h])
--- a/src/drawing.js
+++ b/src/drawing.js
@@ -315,8 +315,8 @@ Crafty.extend({
      * @see Text.textColor
      */
     toRGB: function (hex, alpha) {
-        var hex = (hex.charAt(0) === '#') ? hex.substr(1) : hex,
-            c = [],
+        hex = (hex.charAt(0) === '#') ? hex.substr(1) : hex;
+        var c = [],
             result;
 
         c[0] = parseInt(hex.substr(0, 2), 16);
@@ -339,10 +339,10 @@ Crafty.extend({
  */
 Crafty.DrawManager = (function () {
     /** Helper function to sort by globalZ */
-
     function zsort(a, b) {
         return a._globalZ - b._globalZ;
-    };
+    }
+
     /** array of dirty rects on screen */
     var dirty_rects = [],
         changed_objs = [],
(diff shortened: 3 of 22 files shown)


starwed:
We'll definitely want to squash these commits before merging.  :)

potomak:
@starwed the PR is complete now. I missed some pieces of code on purpose, see https://github.com/potomak/Crafty/commit/0a7bc58a211b28d1a8eaf77b194deebdae4dc8db

Note: I think we need a different loading strategy, what do you think about the [ES6 Module Transpiler](https://github.com/square/es6-module-transpiler)?

starwed:
@potomak  Thanks for the work!  But could you squash these commits together?  

Also, maybe open a new issue on the loading strategy, to discuss pros/cons of current and proposed approaches.

sbyrnes321:
Have you tested the code?

I checked out your branch and ran it in my game, and it doesn't work. I confirmed that this specific pull-request is to blame, because it works when I compile using the version immediately before those commits.

The error I got was: `TypeError: obj.staleRect is undefined` in the line `obj.staleRect._x = rect._x;` in crafty.js line 8464.

starwed:
That would be because the code was relying on the fact that `undefined == null` returns true.  While the old code is fine in the sense that it does exactly what we want, it's probably harder to reason about and so should be changed.  Probably should check directly if it's an object: `if (typeof obj.staleRect !== "object")`

There are a few other places where `== null` checks were replaced with `===null`, which does change the semantics and isn't always what was intended.  Sometimes it's used to check that an object exists, or that an argument was passed.  The linter can't read our minds, so we need to be careful in "fixing" those.

(Compiled coffeescript will use such `==null` comparisons when it can prove that they're safe, but people are a bit worse at thinking these things through than compilers...)

potomak:
@starwed ok about everything (stash, new issue about loading strategy).

@sbyrnes321 I run tests using the task `grunt check` and they seemed ok.

About the `==` vs `===` topic, it's a good practice to always use `===` to avoid unexpected type castings, so I think we can accept to break things now during the alpha stage, do you agree?

potomak:
@louisstow can you please help me understand the Crafty loading process?

See #557

src/drawing.js
```diff
@@ -375,7 +375,7 @@ Crafty.DrawManager = (function () {
                 for (i = 0, l = changed_objs.length; i < l; i++) {
                     obj = changed_objs[i];
                     rect = obj._mbr || obj;
-                    if (obj.staleRect == null)
+                    if (obj.staleRect === null)
```
potomak: @sbyrnes321 this is the change that breaks things, I think we can fix it in a new commit maybe also adding a test case.

starwed:
> About the == vs === topic, it's a good practice to always use === to avoid unexpected type castings, so I think we can accept to break things now during the alpha stage, do you agree?

The point is we were _deliberately_ using type casting there.  That should probably be avoided, but there's no point in having an intermediate version that breaks things.  :)  Pretty much everywhere that `==null` was changed to `===null` should actually be something else; I'll go through and figure out what those should be.

sbyrnes321:
Great! When you think you've solved the problem, I'm happy to re-test.

src/controls.js
```diff
@@ -75,7 +75,7 @@ Crafty.extend({
             type = e.type;
 
         //Normalize button according to http://unixpapa.com/js/mouse.html
-        if (e.which == null) {
+        if (e.which === null) {
```
starwed: I [believe](http://stackoverflow.com/questions/10315260/internet-explorer-showing-e-which-as-undefined) this should check for `undefined`, not `null`.  It's an IE specific hack, so I can't check directly.

src/controls.js
```diff
@@ -211,10 +211,10 @@ Crafty.extend({
         first.target.dispatchEvent(simulatedEvent);
 
         // trigger click when it should be triggered
-        if (lastEvent != null && lastEvent.type == 'mousedown' && type == 'mouseup') {
+        if (lastEvent !== null && lastEvent.type == 'mousedown' && type == 'mouseup') {
```
starwed, line 214: This is a Crafty set variable that is either `null` or an object, so this check looks ok!

src/controls.js
```diff
@@ -484,7 +484,7 @@ Crafty.c("Draggable", {
         };
 
         this._onup = function upper(e) {
-            if (this._dragging == true) {
+            if (this._dragging) {
```
starwed: `._dragging` is an actual boolean, so better to explicitly test whether `this._dragging === true`.

src/core.js
```diff
@@ -240,7 +236,7 @@
             //extend the components
             ul = uninit.length;
             for (; c < ul; c++) {
-                if (this.__c[uninit[c]] == true)
+                if (this.__c[uninit[c]])
```
starwed: The properties of `.__c` should be either `true` or `undefined`, so again use the `=== true` check if we're cleaning these up.

src/drawing.js
```diff
@@ -356,7 +356,7 @@ Crafty.DrawManager = (function () {
         rectManager = {
             /** Finds smallest rectangles that overlaps a and b, merges them into target */
             merge: function (a, b, target) {
-                if (target == null)
+                if (target === null)
```
starwed: If `target` isn't passed to `merge`, it will be `undefined` rather than `null`.  Instead, check `if (typeof target === "undefined")`.

Although, this is pretty deep in an inner loop -- this might be a place where checking `==null` is ok for performance reasons.  I can't find any info on which is more performant in a cursory search.

src/drawing.js
starwed: Looking at this now, it's probably better to define staleRect when each `"Canvas"` object is created, to simply avoid the check here completely.  But for now check against `undefined` rather than `null`.

src/loader.js
```diff
@@ -287,119 +287,123 @@ Crafty.extend({
...
+                                    delay[dset] = [];
                             }
                         }
                     }
                 setTimeout(function () {
                     each(paths, function (path) {
                         if (scripts[path]) {
-                            id && (ids[id] = 1)
-                            return scripts[path] == 2 && callback()
+                            if (id)
```
starwed, line 337: Heh, whoever wrote the original code here _really_ valued terseness...

src/storage.js
```diff
@@ -336,7 +336,7 @@ Crafty.storage = (function () {
             },
 
             save: function (key, type, data, callback) {
-                if (db == null) {
+                if (db === null) {
```
starwed, line 339: We explicitly set `db` to `null` earlier (line 132), so this check is correct.

src/storage.js
```diff
@@ -460,7 +460,7 @@ Crafty.storage = (function () {
             },
 
             load: function (key, type, callback) {
-                if (db[type] == null) {
+                if (db[type] === null) {
```
starwed: I suspect this should be checking against `undefined` rather than `null`. The `type` parameter is either `"save"` or `"cache"` according to the docs, and we don't explicitly set those properties to `null` anywhere that I can see.

Since in the case that it's not missing it simply attempts to call again in a little bit, I'd assume that we're waiting for some other code to init the database, so `undefined` is the most likely value; if in doubt, we could simply leave it as is.

src/storage.js
```diff
@@ -477,7 +477,7 @@ Crafty.storage = (function () {
             },
 
             getAllKeys: function (type, callback) {
-                if (db[type] == null) {
+                if (db[type] === null) {
```
starwed: See comment on l463.

src/time.js
```diff
@@ -61,7 +61,7 @@ Crafty.c("Delay", {
 	* ~~~
 	* console.log("start");
 	* Crafty.e("Delay").delay(function() {
-		 console.log("100ms later");
+    *   console.log("100ms later");
```
starwed: This is still misaligned.

src/storage.js
potomak: https://github.com/craftyjs/Crafty/pull/556/files#r6943168

src/viewport.js
```diff
@@ -372,6 +372,7 @@ Crafty.extend({
                     Crafty.viewport.x += diff.x;
                     Crafty.viewport.y += diff.y;
                     Crafty.viewport._clamp();
+                    break;
```
starwed, line 378: Is this correct?  Or was the code relying on it falling through to the next case?

I think it's the latter, since the diff is added to the viewport each time, and taken as the difference between arg and `lastMouse`.  If it didn't fall through, it would appear that `lastMouse` would never be modified.

This type of confusion is exactly why it's bad style, of course.  :)  So if I'm right, the fix would be to copy those lines in the `start:` case.

starwed:
I went through and checked everywhere the actual semantics changed; added line notes for any issues I saw.

src/viewport.js
potomak, line 378: Sorry about that, I checked this but I didn't understand the code well!

As you suggested, just adding those line to the `case` would have unaltered the behaviour.

potomak:
I will work on them tonight!

Great code review!

src/controls.js
potomak: I think that the best way to test boolean values is without `===` or `==`.

If you know in advance that `this._dragging` is only ever going to have a `true` or `false` value, then comparing it explicitly is just extra code and unnecessary and `if (this._dragging)` is more compact and arguably cleaner/better.

src/drawing.js
potomak: On my browser performances are better using `typeof value === 'undefined'`, see http://jsperf.com/value-null-vs-typeof-value-undefined

src/controls.js
sbyrnes321: The reason that I prefer `=== true` is because it makes it easier to read and understand the code, which in turn makes it easier to maintain and debug it. If I see `if (this._dragging)`, I have to figure out what it means ... Does it mean: "I have a method called `this._dragging()`, but I'm not sure if it's been defined yet!" Or does it mean: "I have an string stored in `this._dragging`, but I'm not sure whether it's empty or not!" Or does it mean: "I have a boolean stored in `this._dragging`, but I'm not sure whether it's true or false!" You cannot tell, you have to spend time and mental energy trying to figure it out.

I don't deny that `=== true` is "less concise", by eight characters. But the purpose of writing concisely is not to conserve pixels, but to allow readers to understand the code at a glance. Writing concisely is not a goal in itself, but a means to an end (bug-free and easily-maintainable code). If deleting eight characters makes the code more ambiguous, then we shouldn't delete the characters. I think that your argument is just as silly as saying "We should replace `var overlap` by `var o` because it's more concise".

potomak: Again, I partially agree. You have to figure out what it means only if you're _not sure_ about your code quility.

> I have a method called `this._dragging()`, but I'm not sure if it's been defined yet!

No, because you would have wrote `typeof this._dragging === 'undefined'`.

> I have an string stored in this._dragging, but I'm not sure whether it's empty or not!

No, because you would have wrote `this._dragging.length === 0`.

Anyway I agree with you that using `=== true` banishes all doubts.

Note: I'm not trying to remove `=== true` to save eight characters but only for a matter of readability.

sbyrnes321:
I checked out the new changes (97502e3) -- this time my game worked with no problem.

potomak:
Nice to know! Thanks @sbyrnes321

starwed:
Ok, it looks good glancing over it!  I'll want to take a second look later, just because there are so many changes, so I'll hold off on merging just for a little bit.

I would mildly prefer it to be one commit, since there's not much point in having an intermediate version with bugs in the repo, but if you don't have time I'll just go ahead and merge both once I've double-checked it.

potomak:
@starwed I pushed a unique commit squashing the previous one.

/cc @sbyrnes321

starwed:
Ok, looks good!  I'm going to go ahead and merge.

There are still a few issues raised by jshint, at least some of which are probably valid.  (Creating lots of identical anonymous functions inside loops, for instance.)  But we can patch those separately, since they involve more intrusive changes.

The use of `eval` to parse JSON should only trigger in IE8 and below, where there's probably no good way around it.

Issues with intro/outro.js, of course, are expected; either we should set jshint to ignore them, or use a different strategy for building crafty.js.  (See #557)

potomak:
Ok, thanks.

PR #558, opened 2013-10-11
IE8 viewport code: Use a global event instead of a listener entity

sbyrnes321:
The advantage of this approach is: IE8 code has the same number of
entities as other browsers, so Crafty('*') behaves predictably etc.
Also, this change removes a layer of unnecessary abstraction /
complexity.


louisstow:
Thanks!

PR #559, opened 2013-10-12
fix unit-tests to avoid false-positives in IE8

sbyrnes321:
As discussed at #456 , this pull-request fixes the core.html unit tests to reduce false-positives in IE8 (and also one in IE10). Some of the unit tests still fail in IE8, but I think the remaining ones are actual problems in the crafty code, not just bad tests. I ran the tests and they all pass in IE10, IE9 mode (in IE10), chromium, firefox, and grunt (phantomjs).

(This is a re-do of https://github.com/craftyjs/Crafty/pull/553 after incorporating the suggestions.)

### DIFF
--- a/tests/core.html
+++ b/tests/core.html
@@ -111,6 +111,11 @@ $(document).ready(function() {
 	});
 
 	test("setter", function() {
+		if (!(Crafty.support.setter || Crafty.support.defineProperty)) {
+			// IE8 has a setter() function but it behaves differently. No test is currently written for IE8.
+			expect(0);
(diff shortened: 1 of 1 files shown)


starwed:
In a very real sense IE8 _does_ fail these unit tests, though -- other parts of the game engine assume that the changes happen instantly, without "EnterFrame" needing to be called.

I kind of suspect that in the long run, we'll root out all the IE8 specific code, but I guess we should at least land this for now, to see just how broken other parts are.

PR #560, opened 2013-10-16
Fixing issue with loader.js not triggering oncomplete() for empty data array

sgmeyer:


starwed:
Have you tested on the develop branch?  There should already be a fix for that issue (see [line 217](https://github.com/sgmeyer/Crafty/blob/27becb7fe682b7c04292e5bbfbf46e9c9064d460/src/loader.js#L217) )

sgmeyer:
Closing.  I verified the fix against dev branch.

Issue #557, opened 2013-10-08
Modules and concatenation strategy


potomak:
I'd like to remove intro.js, outro.js, and to _fix_ core.js, see #556.

/cc @louisstow @starwed

starwed:
Ah, I probably wouldn't use the term "loading" to describe this.  We concatenate all the files together to produce crafty.js, and a release is of that single file.    Someone just writing a game would never see intro.js.

It would be great if each file could also stand alone (since then we could run e.g. jsvalidate against them each separately) but only if it's strictly better than what we do right now.

sbyrnes321:
I agree with @starwed -- I don't think cancatenating is a bad solution. We can just remember that those jshints should be ignored...

potomak:
I partially agree, I think we're also speaking about loading as right now if you want to develop or test the library you must do it against the compiled version.

Using modules we can test them separately and use them separately during development.

Moreover right now compiling is order dependent, I think it's not a reliable solution.

Note: I think that

> We can just remember that those jshints should be ignored...

it's not the best solution.

sbyrnes321:
jquery has an intro and outro -- https://github.com/jquery/jquery/blob/master/src/outro.js -- it can't be _such_ a bad idea.

If you don't like my idea of ignoring those jshint warnings, here's another idea: The grunt jshint task can be edited to avoid jshint'ing the very simple files `intro.js` and `outro.js`, and to ignore the specific error "E006: Unexpected early end of program." in the specific file `core.js`. Then the jshint task can run perfectly with no warnings displayed.

I disagree with your comment "compiling is order dependent, I think it's not a reliable solution." Concatenation has to be done in the order core, intro, (everything else), outro. What is "unreliable"? It's true that someone could change the order of the list in package.json. But then they would immediately notice that this change makes crafty stop working! So they would undo that change, and put the order back to the way it was. If you're really concerned about this, you can add a comment in package.json:

```
// The order of the files in this list needs to be core.js, intro.js, (everything else), outro.js,
// because the files are concatenated in this order to create crafty.js
```

Then you can be 100% confident that the files will always be concatenated in the right order. Very "reliable", right?

Again, I don't see any problem with the status quo. If you have another idea, can you describe more specifically how it would work?

potomak:
[...]

PR #561, opened 2013-10-18
Fixing Core.js broken javascript

sgmeyer:
This appears to have changed  at tree:cb43fc1f40.  This was changed
back to the previous commit tree:ee796c7671

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1605,4 +1605,4 @@
     } else { // browser global
         window.Crafty = Crafty;
     }
-})(window,
+})(window);


sbyrnes321:
It's not broken -- the file is not supposed to be run on its own. It's supposed to be built (along with the rest of the `src`) into the `crafty.js` file before being run.

In the build process, the files of `src/` are concatenated in the order `core.js`, then `intro.js`, then everything else, then `outro.js`. If you look at the files you'll see how they concatenate into a single sensible code.

Sorry if you wasted your time trying to understand it! That's why @potomak doesn't like the idea of concatenating, and has proposed to rewrite it in a different way -- https://github.com/craftyjs/Crafty/issues/557

sgmeyer:
yeah, I can see that now.  Interesting approach.  Thanks for the information.  I was building these locally using Grunt and Uglify.  I didn't have the order correct so this is good to know.

Issue #549, opened 2013-09-18
Sprites visualization error on IE9


potomak:
I'm rendering a 2D component but it doesn't show in IE9.

The component is named `Monster3D`, this is the source code:

``` javascript
Crafty.c('Monster3D', {
  init: function() {
    this.requires('Actor3D, spriteMonster, SpriteAnimation')
      .attr({three: {x: 0, y: 0, z: -100, w: 40, h: 40}})
      .animate('MonsterMoving', 0, 0, 18);

    this.animate('MonsterMoving', 100, -1);
  }
});
```

The `Actor3D` component is a custom component composed by `Canvas` and `3D`, the latest is a simple custom component to render canvas components in a 3D space.

``` javascript
Crafty.c('Actor3D', {
  init: function() {
    this.requires('3D, Canvas');
  },

  at: function(x, y, z) {
    this.three.x = x;
    this.three.y = y;
    this.three.z = z;

    return this;
  }
});
```

I put 3x5 `Monster3D` components by running:

``` javascript
Crafty.e('Monster3D').at(-200, 0, 300);
Crafty.e('Monster3D').at(-100, 0, 300);
Crafty.e('Monster3D').at(   0, 0, 300);
Crafty.e('Monster3D').at( 100, 0, 300);
Crafty.e('Monster3D').at( 200, 0, 300);
Crafty.e('Monster3D').at(-200, 0, 200);
Crafty.e('Monster3D').at(-100, 0, 200);
Crafty.e('Monster3D').at(   0, 0, 200);
Crafty.e('Monster3D').at( 100, 0, 200);
Crafty.e('Monster3D').at( 200, 0, 200);
Crafty.e('Monster3D').at(-200, 0, 100);
Crafty.e('Monster3D').at(-100, 0, 100);
Crafty.e('Monster3D').at(   0, 0, 100);
Crafty.e('Monster3D').at( 100, 0, 100);
Crafty.e('Monster3D').at( 200, 0, 100);
```

At http://popinator.fungostudios.com/error/ there's the test script.

Chrome (Ubuntu 13.04):
![screenshot from 2013-09-18 15 21 05](https://f.cloud.github.com/assets/153842/1165099/a9b6221c-2066-11e3-9a76-3519ff6800ed.png)

IE9 (Windows 7):
![popinator_error](https://f.cloud.github.com/assets/153842/1165100/aff0ebda-2066-11e3-8599-d754c161f517.png)

[...]

PR #562, opened 2013-10-21
Fix image loader on IE; add loader test

potomak:
`obj.src = ""` triggers the error event on IE while it doesn't on webkit
so adding a check for the type of browser fixes the error.

See #549 for more details.


kevinsimper:
Looks good :+1: 

Really cool that you found the solution for the problem!

starwed:
The webkit workaround is discussed in #382.  The original bug in webkit is [7731](http://code.google.com/p/chromium/issues/detail?id=7731).

There might also be a better way around it; the webkit bug only occurs when you try to load an already loaded image, so perhaps we could directly detect that.

kevinsimper:
@starwed How do you think that best could be done? Should there be stored a list of already downloaded files in array?

potomak:
@starwed @kevinsimper I think we could merge this PR to make it work now and open a new issue about that to resolve it in a better way in the future.

Moreover this is a **workaround** to mitigate a webkit bug, so it should be solved _as a workaround_.

Note: this bug is hanging since **11 months** ago, this means that either very few people are using IE or very few people are using Crafty to develop HTML5 games.

kevinsimper:
@potomak I will merge this then. Will you open that new issue?

PR #563, opened 2013-10-21
refactor Draggable code

sbyrnes321:
This is more memory-efficient -- All helper functions only exist once, not a separate copy for each draggable entity. It's also easier to read (in my opinion).

### DIFF
--- a/src/controls.js
+++ b/src/controls.js
@@ -450,49 +450,47 @@ Crafty.c("Draggable", {
     _dragging: false,
     _dir: null,
 
-    _ondrag: null,
-    _ondown: null,
-    _onup: null,
-
-    //Note: the code is note tested with zoom, etc., that may distort the direction between the viewport and the coordinate on the canvas.
+    //Note: the code is not tested with zoom, etc., that may distort the direction between the viewport and the coordinate on the canvas.
     init: function () {
         this.requires("Mouse");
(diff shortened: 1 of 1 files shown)


kevinsimper:
Looks good, it is more readable now :+1:

potomak:
@sbyrnes321 yes code is more readable but I don't think that this method saves memory as

``` javascript
{
  init: function() {
    this.foo = 'foo';
  }
}
```

is the same as

``` javascript
{
  init: function() {},
  foo: 'foo'
}
```

**Update:** I don't think it's matter of memory but it's faster to do as you're doing, maybe because you're removing a lot of code from the `init()` method, see http://jsperf.com/property-vs-this

:+1:

kevinsimper:
Cool that it is 4 times faster! Amazing that there is such a difference.

starwed:
@potomak  You're right if you're talking about literals, but not if you're talking about objects or functions. 

``` javascript
init: { this.foo = function(){} }
```

The above way creates a new function every time a component is called

``` javascript
init: {},
foo: function(){}
```

This way only creates the function once, and all components reference the same function.

So it would save memory.

potomak:
@starwed I'm not sure about that, but I'll think about it, thanks for your explanation!

PR #565, opened 2013-10-22
Delete empty file - JavaScript.js

kevinsimper:
Does anybody know why it is there?


starwed:
It was added without comment in [this commit](https://github.com/craftyjs/Crafty/commit/0f25676be01ef416db7c4b25e7d7740743285a65), and has always been just an empty file.

I'd assume it was added by accident?

PR #568, opened 2013-10-23
Browserify Crafty.js

kevinsimper:
Now instead of concatenate the files, Crafty.js now use the CommonJS approach with Browserify 2.

The intro.js and outro.js files is removed because they is not needed anymore.

Hashmap is exported a module.

All the test are passing.


src/2D.js
```diff
@@ -1,3 +1,6 @@
+var Crafty = require('./core.js')
+,   document = window.document
```
potomak: This syntax is ugly in my opinion and more important it is inconsistent.

kevinsimper: How would you suggest it looked like?

potomak: I like repeating the `var` token (see http://nodeguide.com/style.html#variable-declarations), but to be consistent you should put the comma at the end of each line, see some lines below, for example at https://github.com/kevinsimper/Crafty/blob/6c8835a650b972874cc22e238f1960f16466e3c0/src/2D.js#L15-16

potomak:
I think you can now remove also `crafty-local.js`.

src/2D.js
kevinsimper: I will do that when i get home :)

potomak: :+1:

sbyrnes321:
I like it! I tried it on my game and it still worked. The final crafty.js file is the same size and more-or-less equally readable.

I agree with @potomak that commas should go at the end of the line not the beginning -- jshint prefers that too.

This change would break crafty-local, so we should delete it. (Or could it be edited somehow?) The purpose of crafty-local was to allow you to change the source code after editing it, without needing to install grunt. That's nice, but we don't _really_ need it. I assume that most people who are making a javascript game would be willing to install node and grunt.

Also, this change will break the old PHP build process. So those files might as well be deleted.

kevinsimper:
I made the chances about the commas

I also delete the php files.

starwed:
There are still some references  to `crafty-local` in the test suite.  (In the comments, where there are instructions for running the tests without compiling.)

Presumably this new approach should make it _easier_ to load the files locally, right?

mpetrovic:
The purpose of crafty-local was to 'build' Crafty at runtime, so work on
the engine could be done without invoking a formal build process. Just make
your changes to the files, refresh the browser and examine results.

I'm also an old fart when it comes to programming. Never touched Grunt in
my life.

kevinsimper:
This would make it easier to load the files locally, because all the files is now valid javascript documents.

This would also make it easier to split up the code and only referencing it where it is needed.
It will also be possible now to use code from the npm community.

I have setup grunt now with the task watch, which will with the terminal command `grunt watch`, continues build the crafty.js file. So if any of the files in `src` is changed, it will build it again.

kevinsimper:
This also makes it possible to make source maps, which will in the dev version of crafty.js map errors, directly to the individual files, like they were the 24 individual javascript files.
Then there will never be an javascript error on line 5000 anymore :)

potomak:
This is not the purpose of this PR, but as you're changing `package.json` I think you could also move all `dependencies` to `devDependencies` as they are required to build, test, and manage Crafty code but they're not required by the Cratfy engine itself.

kevinsimper:
You are right, that should be done.

starwed:
I'm not so certain about moving `dependencies` to `devDependencies`.  The reason to checkout the module is to build crafty -- that's all it does!

If you disagree, let's talk about it in a separate issue.  :)

@kevinsimper As far as I can tell, this is ready to be merged.  Was there anything else you wanted to push before we do so?

kevinsimper:
No, that would be all.

PR #571, opened 2013-10-24
Feature/fix version method

potomak:
Note: this PR is based on @kevinsimper's https://github.com/kevinsimper/Crafty/tree/commonjs branch, so if accepted it should be merged only after #568


kevinsimper:
Great!

starwed:
I'm curious what this actually looks like in the generated Crafty code.

Also, this means that we need to update the test whenever we update the version number --  I guess there's no way round that?

kevinsimper:
This changes was not thought through, mainly because as @starwed points at, the whole package.json will be put into the crafty.js file which is not that useful.

I think it has to stay as it was before, even through it would be smart to have the version number only one place in the package.json file.

potomak:
When you `require` a file or a module, browserify includes it in the build script. I think this is the right way to handle version handling. It's better to include `package.js` inside `crafty.js` than update another version number at every release.

I agree about the chicken-and-egg problem of the test, that's why I think it should be just removed.

starwed:
> It's better to include package.js inside crafty.js than update another version number at every release.

What about pulling the version number from package.json when crafty.js is built?

This must be an issue that other projects have had.  :)

potomak:
Well, we could read the package.js file and parse it, but it wouldn't work without using crafty as a npm package (example https://github.com/isaacs/npm/blob/master/lib/npm.js#L58-L60).

Note: I was searching through popular node projects and I found the latest express commit https://github.com/visionmedia/express/commit/ba5c48aa864202ae9962a4bb7d5b570efa9caa44 :smile:

kevinsimper:
@starwed 

> What about pulling the version number from package.json when crafty.js is built?

If you want to do that, which could be cool, you would have to make a new file which contains the version and it then exports the version.

version.js

```
exports.version = "0.5.4";
```

@potomak I do not think you can look at express, because that does not need to be loaded in the browser. The solution by requiring the package.json file is a good solution, but not when it needs to be loaded over the wire.

starwed:
> If you want to do that, which could be cool, you would have to make a new file which contains the version and it then exports the version.

I was suggesting adding an additional step to the build process that reads in both crafty.js and package.json, and then writes out a file with the correct version.  I'm not especially in love with the idea, but it wouldn't need to use browserify in any way.

potomak:
See also #587.

Issue #570, opened 2013-10-24
Move npm development dependencies to devDependencies


potomak:
I think all `dependencies` inside `package.json` should be moved to `devDependencies` as they are required to build, test, and manage Crafty code but they're not required by the Cratfy engine itself.

See https://github.com/craftyjs/Crafty/pull/568#issuecomment-26970762

starwed:
I disagree, because the module is _for_ developing crafty.    The npm docs say:

> If someone is planning on downloading and using your module in their program, then they probably don't want or need to download and build the external test or documentation framework that you use.

But Crafty is not (currently) intended to be used as a standalone module, or to be used from other modules.  The _only_ reason to set it up right now is to work on development.  This might change in the future!  But for now I think we should leave things as they are.

potomak:
@starwed even if Crafty is not currently intended to be used as a standalone module it's **always wrong** to list development dependencies as package dependencies.

From the spec (https://npmjs.org/doc/json.html#dependencies):

> Please do not put test harnesses or transpilers in your dependencies hash.

Note: this update wouldn't change the development process on the contrary it could only enhance its overall correctness.

Moreover npm could be used also as a _frontend package manager_, see http://dontkry.com/posts/code/using-npm-on-the-client-side.html

Anyway, this is a really low relevance issue so feel free to close it if you completely disagree with me.

kevinsimper:
I would also think that the current dependencies should be moved to devDependencies, because they are used for development, and are not used in the sourcecode of Crafty.js.

Also it does not make any difference in the workflow. When you do: `npm install` it will also install devDependencies. Only when you env is set to production, then it will not install devDependencies.

starwed:
AFAICT this was resolved at some point -- everything is now in devDependencies.  So closing!

PR #572, opened 2013-10-25
Move dependencies to devDependencies

kevinsimper:
Discussed in #570

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "src/time.js",
     "src/DebugLayer.js"
   ],
-  "dependencies": {
+  "devDependencies": {
     "grunt": "~0.4.1",
     "grunt-contrib-uglify": "0.2.0",
     "grunt-contrib-concat": "0.3.0",


potomak:
I think it's wrong to close this PR, I still think those are `devDependencies`.

starwed:
> I think it's wrong to close this PR, I still think those are devDependencies.

As far as I can tell, that field exists for projects where you could benefit from installing the project without installing the tools needed to build it.

I don't think that makes sense with crafty -- installing the package without the build tools is never what you want.  The two different fields exist to be useful, not to satisfy some abstract classification scheme.

On the other hand, if we want to move towards the different .js files of Crafty being useful without being built, it _would_ make sense to split off devDependencies.  Using browserify seems to be a step in that direction.  Is it currently possible to use `require` to import Crafty code without building it first?  Or is that something we could easily switch to?  If so, moving things to devDependencies makes sense.

kevinsimper:
As already discussed devDependencies does only matter when your env is set to production, so I do not think it really matters. But as of fact Crafty does not work without these dependencies and therefore I think it is fine as it is now. If it were that Crafty could function without is dependencies, they should be moved to devDependencies.

@starwed Not as it is now. You can require each file, but since we are not `export`ing anything, you do not get access to any functions. It would be easy to smack a lot of `export` object different places, but it would not makes sense the way Crafty depends on a global object and the way does dependency management by listing components and extending new objects.

Like Crafty's core is infact very small, and all the other is a lot of components, because if you in the 2d.js exported the functions, you would not need to require it anywhere else in the Crafty project, because the `.c()` infact does the work by kind of `require()` the component and make the function available in the entities where you need it.

If we wanted to make it a possible to require the individual files, you would make each component as they are now export a module, but i do not think it is useful for anything right now, because it is components and a lot of components depends on other crafty components and require (or browserify) does not know that, eg. `this.requires("2D");`.

Hope that makes sense, it could be cool to make individual files able to be required, but it would make it another project. You would be forced to use Browserify to compile your game to be playable in the browser (which could be cool because of the new possibilities from NPM, which would make components distribution very easy!)

Issue #569, opened 2013-10-23
Text textFont breaks with font-families containing isolated numbers


jameshoward:
E.g. new `Crafty.e('Text').textFont({ family : 'Times New Roman', size : '20px' }).text('Testing');`

Workaround is to switch to double quotes and wrap the family in single quotes.

E.g. family : `"'Times New Roman'"`

Might be nice to include this in the code instead, automatically wrapping families with spaces in single quotes or alternatively just wrapping them all in singlequotes as I don't believe it matters in CSS.

potomak:
@jameshoward maybe you're using an old version of Crafty. I can't find the `setFont` method anywhere.
If you're talking about the `textFont` method (see http://craftyjs.com/api/Text.html) it works for me on Chromium even without quotes.
Can you please setup a working example of your problem? You can use http://jsfiddle.net/

jameshoward:
Yeah, sorry, I corrected the title to textFont but not the text.

On closer inspection it seems the issue wasn't the spaces but the fact the font-family name has a number in it. 'Museo 300'. My example wasn't a good one.

See http://jsfiddle.net/fF6SU/ - the CSS for 'font' isn't applied, presumably as the string that is built isn't valid because it would be '20px Museo 300' while '20px Times New Roman' is fine.

potomak:
I updated the code: http://jsfiddle.net/fF6SU/2/
I can reproduce the error now.

Let's see what @starwed, @kevinsimper or @sbyrnes321 think about this issue.

starwed:
Yeah, the solution should be to just add quotes around the family when creating the font string.

(We can't set the properties individually because that's not an option when styling Canvas text.)

kevinsimper:
I have made a pull request where I solved it.

PR #574, opened 2013-10-25
Text property family should have singlequotes around the string

kevinsimper:
Fixes bug reported in #569

Also made a test for it, which checks that the family has singlequotes around the text.


potomak:
:+1: for the test! Every PR fixing some bug should include a test.

Issue #534, opened 2013-08-29
Make the documentation follow the common template


kevinsimper:
Instead of applying the template, just inserting this instead:

```
--- 
layout: default

---
```

This will make the craftyjs.com website easier to maintain, since one edit in the template will reflect the whole website, and not like now were the API documentation is very static.

starwed:
I might be confused, but I think this removes the ability to build the docs locally.

Can we hook grunt up to a jekyl processor, so that it also performs that step of generating the documentation?

kevinsimper:
I can see what you mean.

It is not possible to do with Github Pages, sadly. Then we have to move the website to a other host.

I think the best solution would be to create two different grunt process, one for local api documentation and one for the craftyjs.com website.

starwed:
> It is not possible to do with Github Pages, sadly. Then we have to move the website to a other host.

I was wondering if we could locally use a similar processor to the one Pages uses, to generate the files locally.

Alternatively, we could just not use Jekyl in favor of some other approach -- nothing about Github Pages prevents this.  :)

kevinsimper:
We could something like https://github.com/visionmedia/dox which makes the documentation into a json file, and then we could use whatever template system.

I would really like that it would be a one-page documentation, because it is that much faster to jump between and you can search the whole documentation with the native browser search.

starwed:
> We could something like https://github.com/visionmedia/dox which makes the documentation into a json file, and then we could use whatever template system.

There are two separate issues here: how we parse documentation from the code, and how we generate a website that hosts not just auto-generated content, but also other resources.  AFAICT dox is the former, and doesn't help with the latter.

We already had an existing solution that generated HTML; that was changed in this bug to integrate with the Pages template system.  I agree that we should use some sort of more formal templating system, so that headers/etc are consistent across all the documentation.  But I think we need to switch to a node based system we can use locally, rather than relying on Github's stuff.

> I would really like that it would be a one-page documentation, because it is that much faster to jump between and you can search the whole documentation with the native browser search.

I tend to think that would be far too unwieldy.   But maybe if I saw it in practice it would be ok?

starwed:
Some searching turned up [assemble](http://assemble.io/), which is a node/grunt-based static site generator.  [This blurb](http://stackoverflow.com/a/16254540/460138) makes it sound ideal for our purposes, but I haven't actually looked at it in depth yet!

kevinsimper:
[...]

PR #580, opened 2013-10-25
Remove old build system

potomak:
This PR removes old PHP build system and API documentation generation. It removes also a lot of static assets (css, images) hosted at https://github.com/craftyjs/craftyjs.github.com.

If merged the [build page](https://github.com/craftyjs/Crafty/wiki/Building) should also be changed.

See also bc7ce76


starwed:
Just to be clear, the static assets aren't used anywhere in the actual API docs, right?

kevinsimper:
As @starwed says, i do not think that all of the assets should be deleted. Mainly because it should be possible to create the documentation on your own computer.

potomak:
I removed images as they are present at https://github.com/craftyjs/craftyjs.github.com/tree/master/images

The current workflow is:
1. generate api documentation
2. move generated docs to https://github.com/craftyjs/craftyjs.github.com

I'm trying to find a better way to make api documentation readable also from outside of the site repo, it should be also a solution for #534.

kevinsimper:
I will go ahead an merge this, because as @potomak point out is the workflow like that now.

But you should be able to build the documentation offline, but that would need to be built from the start, so it is fine to remove the old one now.

starwed:
This did delete some resources needed for building local documentation; most importantly `global.css`, but also a couple images used in the header and footer.

PR #582, opened 2013-10-25
Remove browserify dependency

kenany:
It's not used anywhere besides in grunt-browserify, which already defines it as a peer dependency. It will still be installed to `node_modules` upon `npm install` anyways.


kevinsimper:
Your right, thanks :+1:

PR #583, opened 2013-10-26
Refactor animation.js to use a time based rather than frame based approach.

starwed:
There are still some issues with this, but the base functionality is there!

Before this lands, we need to check that
- [x] the documentation builds correctly, and
- [ ] the new interface actually works for a few simple games.

_Old issues:_
- How durations are handled should be discussed
- Exactly when "SpriteAnimation" should trigger events is unclear to me.  When should "FrameChange" fire -- literally any time the frame changes, or only when it does so as a result of the animation?
- More comprehensive tests are needed, and, probably, some additional examples in the docs
- It just generally needs some more eyes on the new interface for SpriteAnimation!  :)

So this isn't ready to land, but is at least ready to get feedback on.

_e:_ Rebased against the most recent develop, then fixed linting issues detected by jshint.

_Commit notes below:_
## Crafty.easing
- A new object `Crafty.easing`; it abstracts the concept of a state that changes over time.
- Used in the new "Tween" and "SpriteAnimation" components; can also be used when we fix the viewport animation functions.
- Currently treats durations as a number of frames when given as a number, or as a number of ms when given as a string.  Time based is probably a better default, but would break backwards compatibility.
- No public documentation right now, since it's for internal use.
- Added tests in core.html
## Tween
- Uses the new easing object
- Now treats a set of properties to tween as a group, only sending one "TweenEnd" event per group
- Removes some special handling for mouse events that I'm pretty sure is already handled by how the "2D" component works.
- Updated the docs and added tests in core.html
## SpriteAnimation
- Uses the new easing object
- Refactored the interface completely.  Uses `reel()` to set and change reels, `animate()` to play them.  Favors composable single purpose methods over tons of optional arguments.  See documentation/source for full details.
- The duration of an animation is now set when creating a reel, instead of when playing it.
- Updated the docs and tests in animation.html

### DIFF
--- a/src/animation.js
+++ b/src/animation.js
@@ -1,431 +1,558 @@
 var Crafty = require('./core.js'),
-    document = window.document;
+	document = window.document;
+
+Crafty.easing = function(duration) {
+	this.timePerFrame = 1000 / Crafty.timer.FPS();
+	this.duration = duration;   //default duration given in ms
+	this.reset();
+};
+
+
+Crafty.easing.prototype = {
+	duration: 0,
+	clock:0,
+	steps: null,
+	complete: false,
+	paused: false,
+	
+	// init values 
+	reset: function(){		
+		this.loops = 1;
+		this.clock = 0;
+		this.complete = false;
+		this.paused = false;  
+	},
+
+	repeat: function(loopCount){
+		this.loops = loopCount;
+	},
+
+	setProgress: function(progress, loopCount){
+		this.clock = this.duration * progress;
+		if (typeof loopCount !== "undefined")
+			this.loops = loopCount;
+		
+	},
+
+	pause: function(){
+		this.paused = true;
+	},
+
+	resume: function(){
+		this.paused = false;
+		this.complete = false;
+	},
+
+	// Increment the clock by some amount dt
+	// Handles looping and sets a flag on completion
+	tick: function(dt){
+		if (this.paused || this.complete) return;
+		this.clock += dt;
+		this.frames = Math.floor(this.clock/this.timePerFrame);
+		while (this.clock >= this.duration && this.complete === false){
+			this.loops--;
+			if (this.loops > 0)
+				this.clock -= this.duration;
+			else
+				this.complete = true;
+		}
+	},
+
+	// same as value for now; with other time value functions would be more useful
+	time: function(){
+		return ( Math.min(this.clock/this.duration, 1) );
+		
+	},
+
+	// Value is where along the tweening curve we are
+	// For now it's simply linear; but we can easily add new types
+	value: function(){
+		return this.time();
+	}
+
+};
+
+
+
+
+
 
 /**@
- * #SpriteAnimation
- * @category Animation
- * @trigger AnimationEnd - When the animation finishes - { reelId: <reelID> }
- * @trigger FrameChange - Each frame change - { reelId: <reelID>, frameNumber: <New frame's number> }
- *
- * Used to animate sprites by treating a sprite map as a set of animation frames.
- * Must be applied to an entity that has a sprite-map component.
- *
- * Note: All data recieved from events is only valid until the next event of that
- * type takes place. If you wish to preserve the data, make a copy of it.
- *
- * @see crafty.sprite
- */
+* #SpriteAnimation
+* @category Animation
+* @trigger StartAnimation - When an animation starts playing, or is resumed from the paused state - {Reel}
+* @trigger AnimationEnd - When the animation finishes - { Reel }
+* @trigger FrameChange - Each time the frame of the current reel changes - { Reel }
+* @trigger ReelChange - When the reel changes - { Reel }
+*
+* Used to animate sprites by treating a sprite map as a set of animation frames.
+* Must be applied to an entity that has a sprite-map component.
+*
+* To define an animation, see the `reel` method.  To play an animation, see the `animate` method.
+*   
+* A reel is an object that contains the animation frames and current state for an animation.  The reel object has the following properties:
+* @param id: (String) - the name of the reel
+* @param frames: (Array) - A list of frames in the format [xpos, ypos]
+* @param currentFrame: (Number) - The index of the current frame
+* @param easing: (Crafty.easing object) - The object that handles the internal progress of the animation.  
+* @param duration: (Number) - The duration in milliseconds.
+* 
+* Many animation related events pass a reel object as data.  As typical with events, this should be treated as read only data that might be later altered by the entity.  If you wish to preserve the data, make a copy of it.
+*
+* @see crafty.sprite
+*/
 Crafty.c("SpriteAnimation", {
-    /**@
-     * #._reels
-     * @comp SpriteAnimation
-     *
-     * A map in which the keys are the names assigned to animations defined using
-     * the component (also known as reelIDs), and the values are objects describing
-     * the animation and its state.
-     */
-    _reels: null,
+	/*
+	*
+	* A map in which the keys are the names assigned to animations defined using
+	* the component (also known as reelIDs), and the values are objects describing
+	* the animation and its state.
+	*/
+	_reels: null,
 
-    /**@
-     * #._currentReelId
-     * @comp SpriteAnimation
-     *
-     * The reelID of the currently active reel (which is one of the elements in `this._reels`).
-     * This value is `null` if no reel is active. Some of the component's actions can be invoked
-     * without specifying a reel, in which case they will work on the active reel.
-     */
-    _currentReelId: null,
+	/*
+	* The reelID of the currently active reel (which is one of the elements in `this._reels`).
+	* This value is `null` if no reel is active. Some of the component's actions can be invoked
+	* without specifying a reel, in which case they will work on the active reel.
+	*/
+	_currentReelId: null,
 
-    /**@
-     * #._isPlaying
-     * @comp SpriteAnimation
-     *
-     * Whether or not an animation is currently playing.
-     */
-    _isPlaying: false,
+	/*
+	* The currently active reel.
+	* This value is `null` if no reel is active. 
+	*/
+	_currentReel: null,
 
-    /**@
-     * #._frameChangeInfo
-     * @comp SpriteAnimation
-     *
-     * Contains information about the latest frame change event.
-     */
-    _frameChangeInfo: {
-        reelId: undefined,
-        frameNumber: undefined
-    },
+	/*
+	* Whether or not an animation is currently playing.
+	*/
+	_isPlaying: false,
 
-    /**@
-     * #._animationEndInfo
-     * @comp SpriteAnimation
-     *
-     * Contains information about the latest animation end event.
-     */
-    _animationEndInfo: {
-        reelId: undefined
-    },
 
-    init: function () {
-        this._reels = {};
-    },
+	init: function () {
+		this._reels = {};
+	},
 
-    /**@
-     * #.animate
-     * @comp SpriteAnimation
-     * @sign public this .animate(String reelId, Number fromX, Number y, Number toX)
-     * @param reelId - ID of the animation reel being created
-     * @param fromX - Starting `x` position on the sprite map (x's unit is the horizontal size of the sprite in the sprite map).
-     * @param y - `y` position on the sprite map (y's unit is the horizontal size of the sprite in the sprite map). Remains constant through the animation.
-     * @param toX - End `x` position on the sprite map. This can be smaller than `fromX`, in which case the frames will play in descending order.
-     * @sign public this .animate(String reelId, Array frames)
-     * @param reelId - ID of the animation reel being created
-     * @param frames - Array of arrays containing the `x` and `y` values of successive frames: [[x1,y1],[x2,y2],...] (the values are in the unit of the sprite map's width/height respectively).
-     *
-     * Method to setup animation reels. Animation works by changing the sprites over
-     * a duration. Only works for sprites built with the Crafty.sprite methods.
-     * See the Tween component for animation of 2D properties.
-     *
-     * To setup an animation reel, pass the name of the reel (used to identify the reel and play it later), and either an
-     * array of absolute sprite positions or the start x on the sprite map, the y on the sprite map and then the end x on the sprite map.
-     *
-     * @example
-     * ~~~
-     *\/\/ Define a sprite-map component
-     * Crafty.sprite(16, "images/sprite.png", {
-     *     PlayerSprite: [0,0]
-     * });
-     *
-     * \/\/ Define an animation on the second row of the sprite map (y = 1)
-     * \/\/ from the left most sprite (fromX = 0) to the fourth sprite
-     * \/\/ on that row (toX = 3)
-     * Crafty.e("2D, DOM, SpriteAnimation, PlayerSprite").animate('PlayerRunning', 0, 1, 3);
-     *
-     * \/\/ This is the same animation definition, but using the alternative method
-     * Crafty.e("2D, DOM, SpriteAnimation, PlayerSprite").animate('PlayerRunning', [[0, 1], [1, 1], [2, 1], [3, 1]]);
-     * ~~~
-     */
-    animate: function (reelId, fromX, y, toX) {
-        var reel, i, tile, tileh, pos;
+	/**@
+	* #.reel
+	* @comp SpriteAnimation
+	* Used to define reels, to change the active reel, and to fetch the id of the active reel.
+	*
+	* @sign public this .reel(String reelId, Duration duration, Number fromX, Number fromY, Number frameCount)
+	* Defines a reel by starting and ending position on the sprite sheet.
+	* @param reelId - ID of the animation reel being created
+	* @param duration - The length of the animation in milliseconds.
+	* @param fromX - Starting `x` position on the sprite map (x's unit is the horizontal size of the sprite in the sprite map).
+	* @param fromY - `y` position on the sprite map (y's unit is the horizontal size of the sprite in the sprite map). Remains constant through the animation.
+	* @param frameCount - The number of sequential frames in the animation.  If negative, the animation will play backwards.
+	*
+	* @sign public this .reel(String reelId, Duration duration, Array frames)
+	* Defines a reel by an explicit list of frames
+	* @param reelId - ID of the animation reel being created
+	* @param duration - The length of the animation in milliseconds.
+	* @param frames - An array of arrays containing the `x` and `y` values of successive frames: [[x1,y1],[x2,y2],...] (the values are in the unit of the sprite map's width/height respectively).
+	*
+	* @sign public this .reel(String reelId)
+	* Switches to the specified reel.  The sprite will be updated to that reel's current frame
+	* @param reelID - the ID to switch to
+	*
+	* @sign public Reel .reel()
+	* @return The id of the current reel
+	*
+	*
+	* A method to handle animation reels.  Only works for sprites built with the Crafty.sprite methods.
+	* See the Tween component for animation of 2D properties.
+	*
+	* To setup an animation reel, pass the name of the reel (used to identify the reel later), and either an
+	* array of absolute sprite positions or the start x on the sprite map, the y on the sprite map and then the end x on the sprite map.
+	*
+	*
+	* @example
+	* ~~~
+	* // Define a sprite-map component
+	* Crafty.sprite(16, "images/sprite.png", {
+	*     PlayerSprite: [0,0]
+	* });
+	*
+	* // Define an animation on the second row of the sprite map (y = 1)
+	* // from the left most sprite (fromX = 0) to the fourth sprite
+	* // on that row (toX = 3), with a duration of 1 second
+	* Crafty.e("2D, DOM, SpriteAnimation, PlayerSprite").reel('PlayerRunning', 1000, 0, 1, 3);
+	*
+	* // This is the same animation definition, but using the alternative method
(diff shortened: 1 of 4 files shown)


starwed:
**The interface changes to "SpriteAnimation"**

In the last alpha, `.animate()` was used to define reels, and `.playAnimation()` used to play them.  The duration, number of loops, and starting position were all optional arguments to `playAnimation()`.  So many numeric optional arguments were pretty hard to follow, and for some purposes you needed to pass in `null`.  (This was even worse before a previous patch, since `animate()` used to handle _both_ sets of functionality!)

In the new interface, `reel()` is used to define a reel, including the duration of the animation.  `animate()` always plays an animation from the beginning, with an optional argument indicating how many times to loop.  The position is set with the function `reelPosition`.  There's also a function just to specify the number of loops to play, `.loops()`.

kevinsimper:
I will look at it when i get home, but it looks really awesome!

I can see you are just extending the `Crafty` object instead of using `Crafty.extend()`.

Most of the file use `Crafty.extend()`, but also math.js just extend the Crafty object.

starwed:
> I can see you are just extending the Crafty object instead of using Crafty.extend().

I think I was copying `Crafty.polygon`.  (2D.js L1159).  What are the advantages of using extend here?

kevinsimper:
There is really no advantage of using extend, if you look at the source of the extend() code it is rather trival.

I just think we should use the same method everywhere.

But i like the prototype approach more. It is also cleaner and faster.

src/animation.js
```diff
@@ -1,518 +1,692 @@
...
-     */
-    _reels: null,
+Crafty.easing = function(duration) {
+	this.timePerFrame = 1000 / Crafty.timer.FPS();
+	if (typeof duration == "number"){
+		this.duration = duration * this.timePerFrame;	// duration given in frames
+	} else {
+		this.duration = parseFloat(duration);	//duration given in ms
+	}
+	this.reset();
```
kevinsimper: Just a small thing. I think you should make a function called init(), this would make it more clear, that you are initializing and not  "re setting" the variables from start. I did undetstand it, but not at first.

starwed: Well, the prototype function _is_ the init function; it's run at object creation.  And since we want the same values on reset or init, we don't want to duplicate the code that sets them.

kevinsimper:
I have read and tried your code and I think it is really solid. I like how the code now is more split up and the functions with the new easing object is way better.

I also really like that you now has to define reels and then afterwards animate them.

You say that you like to discussed how durations are handled? What are you thinking about? Like how many frames an animate is played over now that it may be not fixed anymore?

starwed:
> You say that you like to discussed how durations are handled? What are you thinking about? Like how many frames an animate is played over now that it may be not fixed anymore?

Currently you specify a duration in frames.  I kept that as the default in the submitted patch, but if you pass a string it's parsed as a duration in ms.  Measuring durations in ms would probably be a better default.   
- _Pros:_ not dependent on the underlying frame rate, easier to understand intuitively, maps better to the actual mechanism used.
- _Cons:_ would break existing code.

The other decision is when to fire a "FrameChange" event.  
- Option A: The event fires when, due to animation, a frame changes
- Option B: The event fires when the frame changes for _any_ reason, such as resetting the animation or switching reels.

I don't really understand what a typical use of the event would be, so its hard for me to decide which behavior is more useful.  :)  Maybe @grozen has an opinion?  (He did the last rewrite of the component that I continued.)

kevinsimper:
I think we should specify in ms, both because that is how people starting to make games think. Now we are just doing the calculating in our head beforehand. 
Also now because of the variable and semifixed mode. Woundn't that mix something up if you were on either of these modes and used frames as measure?

I don't know about the FrameChange. But i think option a would not be enough.

grozen:
As a matter of fact, I do have an opinion.

As per the way to specify duration, I am all for switching solely to ms durations. I see several reasons for this:
- As you already stated, ms are a much more natural way of thinking about durations.
- Having the input treated differently based on its type is very confusing, so if I had to pick one it would be ms
- My previous change was a breaking change anyhow. Did it make its way into a non-alpha version? If not, there's nothing stopping us from changing the way we break things.

Regarding the events, I would actually go for option A as I see the event as something that happens while an animation is playing, and not while you set it up. When you do set it up, you are actually doing something active so you might as well do whatever it is you would have done in the event handler.
I have to admit, though, I've never actually done anything useful with the FameChange event. I just made sure it kept working.

starwed:
> My previous change was a breaking change anyhow. Did it make its way into a non-alpha version? If not, there's nothing stopping us from changing the way we break things.

Nah, that's actually why I haven't pushed another non-alpha release; totally changing the animation interface twice in short succession seemed like it would be a bad idea!  :)  Now, there are other places we use durations; tweening and some viewport stuff, for instance.

But it seems like we all agree that the ms approach makes the most sense.

> Regarding the events, I would actually go for option A as I see the event as something that happens while an animation is playing, and not while you set it up. When you do set it up, you are actually doing something active so you might as well do whatever it is you would have done in the event handler.

To play devil's advocate, one of the uses of an event-based system is that two pieces of code don't need to know about each other, they just need to subscribe to the correct events.

The main reason I can think of to listen for "FrameChange" events would be to sync up something to the appearance of the sprite, in which case you'd want to know _any_ time it changed.  So perhaps that's best?

Thanks for the feedback.

tests/animation/sprite-animation.js
```diff
@@ -5,92 +5,294 @@ Crafty.sprite(64, 'numbers.png', { 'numbers': [0, 0] });
...
 spriteAnimation.bind("AnimationEnd", function(endData) {
-	finishedAnimations.push(endData.reelId);
+	finishedAnimations.push(endData.id);
 });
 
-module("Sprite Animation", {
+
+module("No reels defined");
+
+test("Test functions when no current reel is defined", function() {
```
grozen: While everything this test tests is indeed important, I am not a fan of mega-tests. Would you consider breaking this down into individual tests for each of the different methods you exercise here?

tests/animation/sprite-animation.js
```diff
@@ -5,92 +5,294 @@ Crafty.sprite(64, 'numbers.png', { 'numbers': [0, 0] });
...
+	spriteAnimation.reel("count");
+
+	
+});
+
+
+test("Test using .reel to set an animation using start and end values", function(){
+
+	spriteAnimation.reel('short-test', 3, 0, 0, 2);
+	spriteAnimation.reel('short-test');
```
grozen: These two lines in succession look a little strange. I would suggest either having calls to define reels always set the reel as well, or turning `reel(String)` to `setReel(String)`.

tests/animation/sprite-animation.js
```diff
@@ -5,92 +5,294 @@ Crafty.sprite(64, 'numbers.png', { 'numbers': [0, 0] });
...
+	var reel = spriteAnimation.getReel('short-test');
```
grozen: I found a definition for `getReel()`. Is there another definition somewhere that takes in an argument?

src/animation.js
```diff
@@ -1,518 +1,695 @@
...
+    * Defines a reel by an explicit list of frames
+	* @param duration - The length of the animation.  
+	* @param reelId - ID of the animation reel being created
+	* @param frames - Array of arrays containing the `x` and `y` values of successive frames: [[x1,y1],[x2,y2],...] (the values are in the unit of the sprite map's width/height respectively).
+    *
+	* @sign public this .reel(String reelId)
+    * Switches to the specified reel.
+	* @param reelID - the ID to switch to
+    *
+	* @sign public Reel .reel()
```
grozen: Why do we have both this and `getReel()`?

tests/animation/sprite-animation.js
```diff
@@ -147,31 +351,36 @@ test("Play an animation while another is already playing", function() {
...
 
 test("Try to play an animation after it ends", function() {
-	spriteAnimation.playAnimation('count', 10);
+	spriteAnimation.animate('count');
 	Crafty.timer.simulateFrames(10);
-	spriteAnimation.playAnimation('count', null);
+	spriteAnimation.resumeAnimation();
 	Crafty.timer.simulateFrames(1);
-
+	// Does this test make sense?
```
grozen: It's either that, throwing an error or doing nothing. I'm not a big fan of doing nothing, as it can create situations where people have a hard time figuring out what they are doing wrong.

tests/animation/sprite-animation.js
```diff
@@ -179,10 +388,10 @@ test("Reset an animation", function() {
 });
 
 test("Reset an animation to a specific frame", function() {
-	spriteAnimation.playAnimation('short', 3);
+	spriteAnimation.animate('short');
 	Crafty.timer.simulateFrames(3);
-	spriteAnimation.resetAnimation(null, 1);
-	spriteAnimation.playAnimation('short', 2);
+	spriteAnimation.reelPosition(1);
+	spriteAnimation.resumeAnimation();
```
grozen: Does setting the reel position pause the animation? I couldn't find any evidence that it is indeed the case. If it is not, then why resume the animation here?

grozen:
Really great stuff! I hope you don't mind me nitpicking/wondering over some things in the tests.

tests/animation/sprite-animation.js
starwed: Yeah, that was actually bugging me too.

tests/animation/sprite-animation.js
starwed: Subjectively it looks fine to me.  I did wonder about automatically switching to a reel when you define it.  But then there's no good way to define a reel when one is already playing.

tests/animation/sprite-animation.js
starwed: Oops, there should be.

src/animation.js
starwed: This returns the ID of the current reel, while `getReel()` returns the actual reel object.

tests/animation/sprite-animation.js
starwed: Hah, I should have left a comment, since I had the same confusion when I looked back at the tests.

The animation is only 3 frames long, so it has ended (and is thus auto-paused) before reelPosition is called.

tests/animation/sprite-animation.js
grozen: Fair enough, though I have to question how common of a scenario it would be to define a reel while another is playing. Seems like the sort of thing you would do during initialization.

starwed: You are probably defining the animations and playing them in separate places anyway, so it won't look as weird in practice.   I'm pretty agnostic on whether overloading `reel()` or having a separate `setReel()` is better -- anyone else have an opinion?

kevinsimper: I like the way that if you call the function with no arguments, you get the object instead. But that is just my personal preference.

grozen: In light of both your personal tastes and what @starwed said about the calls usually not residing next to each other, I will get on the `reel(String)` bandwagon as well.

kevinsimper:
Could this be merged or was there something that should be changed?

tests/animation/sprite-animation.js
potomak: I'm not a fan of overloading in general. It's not intuitive for users and I don't like to see a lot of `if (arguments.length === 0)` in the code.

My favorite API would have three methods:
1. `reel` or `createReel` to create new reels
2. `setReel` to set current reel
3. `getReel` or `currentReel` to retrieve current reel

potomak:
I completely agree with @grozen at https://github.com/craftyjs/Crafty/pull/583#issuecomment-27348843 about durations (only ms, let's break things, let's clarify the API) and also about the `FrameChange` event.

tests/animation/sprite-animation.js
kevinsimper: Yeah, but i think we should hold the API consistent, because we are already doing like starwed suggest with the current .attr().

https://github.com/craftyjs/Crafty/blob/develop/src/core.js#L408

starwed:
Ok!  Pushed a version that I think has complete tests & documentation.  Regarding the changes discussed:
- Durations now default to `ms`, since that seemed unanimous.  You can pass in string arguments that specify either frames or durations as as the unit too.  I kept this in a separate commit, in case we want to revert to the old behavior.
- I made it so that "FrameChange" fires whenever the current frame of a reel changes from one value to another.  This required changing some of the tests, since there won't be an event when you start the animation.  

Also, a couple of behavior changes from the previous version.
- I added "ReelChange" and "StartAnimation" events, since if you care about frame changes you probably care about those too!  
- I also made it so that switching the current reel updates the displayed sprite, since that's what you'd expect.
- I made the sure the test "playground" reflected how the current methods worked.

Before this lands, we need to check that
- [ ] the documentation builds correctly, and
- [ ] the new interface actually works for a few simple games.

src/animation.js
```diff
@@ -1,431 +1,581 @@
...
+	* // Define an animation on the second row of the sprite map (y = 1)
+	* // from the left most sprite (fromX = 0) to the fourth sprite
+	* // on that row (toX = 3), with a duration of 1 second
+	* Crafty.e("2D, DOM, SpriteAnimation, PlayerSprite").reel('PlayerRunning', 1000, 0, 1, 3);
+	*
+	* // This is the same animation definition, but using the alternative method
+	* Crafty.e("2D, DOM, SpriteAnimation, PlayerSprite").reel('PlayerRunning', "1 second", [[0, 1], [1, 1], [2, 1], [3, 1]]);
+	* ~~~
+	*/
+	reel: function (reelId, duration, fromX, y, toX) {
```
potomak: I think that this API is not intuitive.

Now `fromX` is `fromX` times `tile` in the x axis, `y` is `y` times `tileh` and `toX` is used as the limit in the x axis.
Moreover if you set a value for `toX` that is less than `fromX` the sprite is used contrariwise.

In my opinion it would be better to pass only a `reelId`, a `duration`, a `numberOfFrames`.
I would like also an optional parameter `tileSize` of type `Object` that should contain a specific tile size for the reel in the form `{w: Number, h: Number}`.

I think also that the version with the array of positions is _useless_.

starwed: > In my opinion it would be better to pass only a reelId, a duration, a numberOfFrames.

You'd still need to specify a starting location.  The idea is to use spritesheets which might contain multiple animations.

I could definitely see providing a number of frames rather than a finishing frame, though.  I haven't looked to see if the sprite component has the necessary information, but that might make it easier to use animations that wrap around the sprite sheet.

> I would like also an optional parameter tileSize of type Object that should contain a specific tile size for the reel in the form {w: Number, h: Number}.

The animation component depends on a previously defined sprite.  The tilesize is already defined on that sprite.  What would be the advantage of specifying it here, given that it isn't optional on the sprite component?

> I think also that the version with the array of positions is useless.

You want to be able to specify arbitrary sequences of frames, so how else could you do it?  A sprite sheet might contain frames used by multiple, independent animations.

src/animation.js
```diff
@@ -1,431 +1,581 @@
...
-     * A map in which the keys are the names assigned to animations defined using
-     * the component (also known as reelIDs), and the values are objects describing
-     * the animation and its state.
-     */
-    _reels: null,
+Crafty.easing = function(duration) {
+	this.timePerFrame = 1000 / Crafty.timer.FPS();
+	if (typeof duration === "number"){
+        this.duration = duration;   //default duration given in ms
+	} else if (typeof duration === "string") {
```
potomak: Let's simplify, let's accept only numbers!

`duration` is a number of milliseconds, as every duration in JavaScript.

src/animation.js
potomak: > You'd still need to specify a starting location. The idea is to use spritesheets which might contain multiple animations.

You're right! Sorry I forgot that you must specify a starting position!
**Note:** I think that this offset (x, y) should be **expressed in pixels** from the top-left corner of the sprite as usual, not in `tile` times and `tileh` times as it's now.

Also the `duration` could be specified when you start the animation.

To summarize the in my opinion the correct arguments list should comprehend:
- `reelId`
- `offsetX`
- `offsetY`
- `numberOfFrames`
- `tileSize` (optional)

> The tilesize is already defined on that sprite. What would be the advantage of specifying it here, given that it isn't optional on the sprite component?

Yes, you could specify tile size at sprite definition, but this would prevent you from using sprites with different tiles size in the same component. This is also the reason why I think that specify tile size at sprite definition less useful than specify it at reel creation.

**Note:** I'm not sure about this idea of having reels with tiles of different sizes for the same component as it would add a lot of complexity to handle component size updates.

> You want to be able to specify arbitrary sequences of frames, so how else could you do it?

Why would you want to specify arbitrary sequences of frames? This could be achieved by building a spritesheet following specific rules.
I think this is useless and it adds unnecessary complexity, let's **keep it simple**.

src/animation.js
kevinsimper: I think it make sense to do it they way it is, because it is very flexible, so if you have a sprite you only want to play 2 frames of, you do not have to calculate that time it would be in ms.

potomak: @kevinsimper you could just create a new reel with only two frames.

starwed: I mostly put this option in because we use frames right now, and wanted to ease the transition for folk upgrading.  Perhaps a better way is to have a helper function that converts frames to ms.  Something like `Crafty.duration()`?

src/animation.js
starwed: > I think that this offset (x, y) should be expressed in pixels from the top-left corner of the sprite as usual, not in tile times and tileh times as it's now.

I think once you slice a sprite sheet, you should only need to worry about the abstract coordinates.  Remember that it's not just the tile size, but also padding that you need to consider.

> Yes, you could specify tile size at sprite definition, but this would prevent you from using sprites with different tiles size in the same component. This is also the reason why I think that specify tile size at sprite definition less useful than specify it at reel creation.

You can call Crafty.sprite multiple times, to slice the same sprite in multiple sizes.  The advantage of what you're suggesting is to have multiple animations of different sizes on the same entity.  I'm not sure this is something we need to support directly.  (You could still do this by manually swapping out the underlying sprite component used.)

> Why would you want to specify arbitrary sequences of frames? This could be achieved by building a spritesheet following specific rules.

Imagine you had several animations that always had the same beginning and ending state; you don't want to duplicate the image each time.  You're right that you always _could_ prepare the sprite sheet so that each animation is sequential, but that's not necessarily how people want to work, and it might be pretty inefficient.    It also lets people tweak the animation without rebuilding the art assets. 

And this doesn't add much complexity, since it maps pretty easily to the way things work on the backend.

starwed: The optional specified offset should definitely not be part of this patch; since it's an augmentation it could always be added later.  (Though, I think if it makes sense to add it, it should probably be something you do when defining sprites, not just when animating them.)

How do other people feel about declaring a number of frames instead of a final position?

starwed:
Ok, I managed to build the documentation using assemble, and fixed a few doc bugs (and one unrelated logic error when switching reels).  [Here](http://ludonomicon.com/crafty/SpriteAnimation.html) is the SpriteAnimation doc page; some missing resources and slightly off CSS, but good enough to check that the docs look ok.

If anyone has a game using animations, it would be nice to test whether any issues come up when converting from the old api!

Regarding the suggestions:
- I'm ok with removing the string options when passing durations, if people feel like that's a good idea.
- I'd be somewhat pro chagging the `toX` paramter to a `frameLength` instead.  (Specifying how many frames the animation is, instead of where it ends.)  I could see that being a little more flexible.  (It's pretty common to have animations that might wrap around the sprite sheet, and while we don't support that right now the second syntax would make more sense in that context.)

potomak:
> I'm ok with removing the string options when passing durations

:+1: 

> I'd be somewhat pro chagging the `toX` paramter to a `frameLength`

:+1: 

**Note:** I would name it `framesNumber` or `reelLength`, but not `frameLength`

> If anyone has a game using animations, it would be nice to test whether any issues come up when converting from the old api!

I have a deadline for this weekend to present the prototype of a game I'm working on, after that deadline I will be happy to try it on it where I'm using tons of animations.

starwed:
I made the two changes mentioned above; if no one has any objections I'll merge this behemoth tomorrow.  That's probably the best way to get it tested.  :)

mucaho:
can you add https://github.com/talitapagani/craftytweener ?

mucaho:
Github source browser shows me identation issues. Use your sublime to fix them before merging (btw I like it :) )

mucaho:
@grozen @starwed  I'm actually for option B on "FrameChange", as it's more natural to me that you get informed the frame changes from _any_ source. It would also streamline the code.

starwed:
> can you add https://github.com/talitapagani/craftytweener ?

Yeah, adding the ability to use easing effects other than linear is the next step.  I hadn't seen that component, so I'll take a look at their implementation -- I had glanced at the definitions of CSS3 transitions, and they seem fairly messy.  If someone has already done the work, that's great!

You'll also be able to just pass an arbitrary function, of course -- that bits easy.

starwed:
Ok, fixed the default loop bug, added a test for it, made the indentation consistent.

I also changed the behavior of `reel("reelID")` so that it pauses the animation whether we're switching to a new reel or not; that way it ends up in a known state afterwards.  Principle of least surprise and all that.

I think this is now ready to merge; I'll squash the recent commits and do so later tonight, unless someone notices something wrong!

As an aside, this patch almost doubles the number of unit test assertions.  If ever anyone's super bored, the rest of the code could use this sort of coverage.  :)

kevinsimper:
I think it looks really good! really a improvement

starwed:
Ok, merged this.  Hopefully we've caught any bad bugs, but I'm sure there are a couple of small things I missed that'll turn up once people are using it.  :)

I did catch several corner-case bugs while writing unit tests, though, so that was helpful.

mucaho:
:+1:

PR #586, opened 2013-10-27
Fix or suppress warnings produced by jshint

### LINKED ISSUE #539: JSHint on Crafty
If you run the `grunt check`, the JSHint tool is also run on Crafty.js, but returns a lot of errors.

Is it something that should be enforced?

There is around 200 errors right now.

starwed:
This patch makes it so we pass jshint without any issues.
- Moves some function creation outside of loops.  One of those might have actually generated a lot of unnecessary garbage, the other probably doesn't make a difference.
- Jshint complained about our use of eval in `storage.js`; since we need that for IE8, I suppressed the warning for that file.
- It also didn't like the tricky uses of `&&` and `||` to make a bit of code more concise in `loader.js`.   That code is actually so unreadable I just suppressed those warnings for now.  If someone wants to patch that up, go for it!

This would fix #539.

### DIFF
--- a/src/extensions.js
+++ b/src/extensions.js
@@ -220,6 +220,11 @@ Crafty.extend({
         paddingX = parseInt(paddingX || 0, 10); //just incase
         paddingY = parseInt(paddingY || 0, 10);
 
+        var markSpritesReady = function() {
+            this.ready = true;
+            this.trigger("Change");
+        };
+
         img = Crafty.asset(url);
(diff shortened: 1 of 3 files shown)


kevinsimper:
Really cool, now there are a real workflow which we can use for submitting new code! :+1: 

I have tested it, and it works. I will go ahead and merge it.

PR #587, opened 2013-10-29
Get the version from the package.json and inject it into version.js

kevinsimper:
This reads the version from the package.json and inject it into version.js

This solves that the whole package.json file is not included in the browserified crafty.js, when it is build. Saves kilobytes.

I also changes the test so it checks for a non empty string, so the test do not have to be changes every time the version is bumped.

---

I first create a more complex solution where the version.js file contained `module.exports = "dev";`, and only when you was building for release (`grunt release`), it would change the version and afterwards cleanup and put it back to what is was before. But that meant that when we was building for development, there would be no version in the crafty.js file, and i did not think that was a good idea.

---

This is the simple solution. Maybe it is best to Keep It Simple?

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,4 +1,5 @@
 require("coffee-script");
+var fs = require('fs');
 
 module.exports = function (grunt) {
     var pkg = grunt.file.readJSON('package.json');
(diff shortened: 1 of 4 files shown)


potomak:
:+1:

starwed:
Ok, this approach seems pretty sensible

PR #590, opened 2013-10-29
Use `Crafty.math.degToRad`

potomak:


starwed:
We should make sure this doesn't hurt performance.  

I did a quick [jsperf test](http://jsperf.com/inline-vs-function-degtorad) that seemed to indicate it could in principle make a difference.  (The function version was slower in Chrome.)   

I would suspect it has a negligible impact overall, but its best if we explicitly test that.  These are kind of "low level" parts of Crafty.

kevinsimper:
We should definitly use the inline function.

Ran the 2. test in all major browsers and there were a big difference! Especially in IE10

image

http://jsperf.com/inline-vs-function-degtorad/2

See charts on page.

potomak:
@kevinsimper see my comment at http://jsperf.com/inline-vs-function-degtorad/3#comment-1 about version 3 of the test.

kevinsimper:
The closure function now when it is working is only 1-3% faster sometimes, so go with the `rad = rot * Crafty.math.DEG_TO_RAD`

starwed:
So what was the conclusion?  Is there a reason to switch to the function instead of just multiplying by a constant?

starwed:
Closing for now -- if I've misunderstood the consensus just reopen.

PR #592, opened 2013-11-01
Gravity collision detected only on top.

pawurb:
I am working on a runner inspired by http://www.mylostgames.com/play/nyan_cat_lost_in_space . I need a mechanism for main character to jump on platforms from below (like in Icy Tower). I am using Twoway and Gravity components. 

Current implementation of gravity against 'anti' objects make character go immediately up at any contact: http://www.youtube.com/watch?v=kdhKF39-IKc . 

This pull request changes make character being stopped by platforms only when he is on top on them: http://www.youtube.com/watch?v=j6vOzTU3kaY. Treshold of 10 pixels was choosen as it avoids jumpy animation and  ensures good collision detection.

I would be grateful for a tip on how to detect if character is moving up or down, so that I can stop his movement only if he gets on platform from above.

### DIFF
--- a/src/2D.js
+++ b/src/2D.js
@@ -1045,6 +1045,7 @@ Crafty.c("Gravity", {
      */
     gravity: function (comp) {
         if (comp) this._anti = comp;
+        if(isNaN(this._jumpSpeed)) this._jumpSpeed = 0; //set to 0 if Twoway component is not present
 
         this.bind("EnterFrame", this._enterFrame);
 
@@ -1108,8 +1109,10 @@ Crafty.c("Gravity", {
             }
         }
 
-        if (hit) { //stop falling if found
-            if (this._falling) this.stopFalling(hit);
+        if (hit) { //stop falling if found and player is moving down
+            if (this._falling && ((this._gy > this._jumpSpeed) || !this._up)){
+              this.stopFalling(hit);
+            } 
         } else {
             this._falling = true; //keep falling otherwise
         }
--- a/src/controls.js
+++ b/src/controls.js
@@ -908,12 +908,16 @@ Crafty.c("Twoway", {
         });
 
         if (speed) this._speed = speed;
-        if (arguments.length < 2) jump = this._speed * 2;
+        if (arguments.length < 2){
+          this._jumpSpeed = this._speed * 2;
+        } else{
+          this._jumpSpeed = jump;
+        }
 
         this.bind("EnterFrame", function () {
             if (this.disableControls) return;
             if (this._up) {
-                this.y -= jump;
+                this.y -= this._jumpSpeed;
                 this._falling = true;
             }
         }).bind("KeyDown", function () {


starwed:
Maybe change the magic number 10 to a property that defaults to 10?   Something like `_gravityCollisionThreshold`.

(In the long run, I found it easier to write my own movement components than to work with the built-in ones.  But you should be able to figure out the direction of motion by tracking "Move" events.)

pawurb:
We can determine if object is moving vertically up or down by comparing `_gy` with its jump speed value. To do so I had to store `_jumpSpeed` as a player's attribute. In the case when someone uses Gravity without Twoway component it will be set to 0 and also works. Now player will be stopped on top of the platform only when he comes in contact with it when moving down (either in the descending phase of the jump or after falling from higher platform). That is why we no longer need `_gravityCollisionThreshold`. Result: http://www.youtube.com/watch?v=SkKXAzI8J2Y

kevinsimper:
Really cool with the video showing the change.

I think this is a improved, nice done, but what if you want to make a Mario platformer game? Wouldn't that be difficult now?

pawurb:
I think it is still very flexible: http://www.youtube.com/watch?v=CP_y1_93lTY

I achieved this effect by combining Collision module with a custom stopMovement callback function executed onHit. My change just prevents player from 'teleporting' on top of the 'anti' platform when hitting it from below.

mucaho:
@pawurb 
~~In many cases it's better to extend the API than to modify the existing behaviour. Can you make this an optional feature?~~
Edit:
I watched your videos and understand better now.
What happens if the character jumps over a platform, then starts falling down while he is in middle of platform? Will he be teleported up to the platform? If so, i'm all thumbs up for the change :)

If you just want to prevent teleporting on top of the anti platform from below, try to limit movement of player so he can not intersect with the platform above:

``` javascript
var platform = Crafty.e("2D, platform");
var player = Crafty.e("2D, Collision, Gravity, Twoway ")
                          .twoway(...)
                          .bind("Moved", function(from) {
                              var hit = !!this.hit("platform");
                              if (hit) {
                                  // this.x = from.x; // you dont want to limit horizontal movement, but if you do: uncomment the line
                                  this.y = from.y; 
                              }
                           });

```

pawurb:
Just like you said @mucaho . If character is in contact with the platform during the descending phase of the jump then he will be teleported on top of it.

starwed:
This patch makes "Gravity" implicitly depend on "Twoway", due to the references to jump speed.  So you could no longer use it on things like projectiles.

pawurb:
I have to disagree. If Twoway component is not present,  that the `_jumpSpeed` will be set to 0 and old behaviour will be preserved.

starwed:
Ah, I see.  Not a particular fan of hardcoding these things.  But I think in the future we'll rewrite the platforming components completely, so it probably doesn't matter too much what we do right now.

Could you squash these commits together?  Then we can merge it, unless there's some issue I haven't understood!  :)

starwed:
Oh, and the linter says you need a semicolon on L914: `this._jumpSpeed = jump`

You can run `grunt jshint` to see these types of warnings, or `grunt check` to first compile crafty and then run unit  tests and jshint.

pawurb:
I got a bit crazy with git but it should be good to merge now.

starwed:
Hehe, thanks!  

The old behavior of Gravity always struck me as kind of crazy, especially when I was first trying out Crafty.

Issue #540, opened 2013-09-04
Add Travis-ci for continuous integration


kevinsimper:
The repo owner needs to enable Travis-ci integration.

kevinsimper:
![screen shot 2013-09-04 at 13 22 44](https://f.cloud.github.com/assets/1126497/1079874/60aed036-1554-11e3-9a79-c782bd8d9d47.png)

Somebody with admin access has to go to travis-ci.org and activate the service, before it will work.

@louisstow, can you do that? :-)

kevinsimper:
If we added Travis, we could build the crafty.js development source file. Sometimes users report bugs, that we think has been solved, so we ask the user to try it with the latest build. This way we could save a lot of time with bugs.

Normal users who are getting started with Crafty.js, do not know how to build crafty.js dev, without using a lot of time figuring it out.

Lately also many tests has been added or fixed by @sbyrnes321 , so it would be cool to have them ran, whenever a Pull-request is submitted.

potomak:
I think we should also add also the support for running tests using http://ci.testling.com/, some work is required, I'll open a related issue soon.

potomak:
Note: I can't still find any build at https://travis-ci.org/craftyjs/Crafty

starwed:
Someone (@louisstow?)  seems to have flipped the right switch, because I'm now seeing Travis messages integrated into pull requests.

kevinsimper:
No, it is because I have added it to my crafty fork, so on all my pull requests there will be travis-ci messages.

So we still need Louis to add Travis-ci.

starwed:
Aha.  It does seem like this would be quite useful.

kevinsimper:
Yes, very useful, especially on pull request from new people.

starwed:
@louisstow  Could you comment on this?  Even if it's just to say you don't have time to think about the issue right now.  :)

starwed:
Ok, louis added me as an owner, and I enabled travisCI.  It seems to be working -- I got emails about the last two builds passing, any way.

kevinsimper:
@starwed That is really great! Could you enable testling.com also? :)

PR #594, opened 2013-11-02
Change how rotation cascades for 2D entities

### LINKED ISSUE #593: Rotations don't compose correctly
Rotating an entity, attaching it to a parent, and then rotating the parent doesn't work the way you'd expect.

Currently when the parents rotate, we rotate all children about the same origin.  But this doesn't work if the children are already rotated about their own origins.

I think the correct approach is to manually change the `x` and `y` coordinates of children when the parent rotates, while also rotating them about their own origin.

As reported [by Justin](https://groups.google.com/forum/?fromgroups#!topic/craftyjs/FS1UvM9t_ms) on the google group.  [A fiddle that shows the issue.](http://jsfiddle.net/2VnKw/1/)

starwed:
Previously when a parent rotated, it changed the origin and rotation of children directly.  This prevented independent rotation of children.

After this patch, children are rotated about their own origin, and shifted to the correct position relative to the parent.

This fixes #593.

### DIFF
--- a/src/2D.js
+++ b/src/2D.js
@@ -935,12 +935,12 @@ Crafty.c("2D", {
      * Method for rotation rather than through a setter
      */
     rotate: function (e) {
-        //assume event data origin
-        this._origin.x = e.o.x - this._x;
-        this._origin.y = e.o.y - this._y;
-
-        //modify through the setter method
+        var x2, y2;
+        x2 =  (this._x + this._origin.x - e.o.x) * e.cos + (this._y + this._origin.y - e.o.y) * e.sin + (e.o.x - this._origin.x)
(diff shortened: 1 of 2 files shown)


starwed:
I'm going to go ahead and merge this.

src/2D.js
```diff
@@ -935,12 +935,12 @@ Crafty.c("2D", {
      * Method for rotation rather than through a setter
      */
     rotate: function (e) {
-        //assume event data origin
-        this._origin.x = e.o.x - this._x;
-        this._origin.y = e.o.y - this._y;
-
-        //modify through the setter method
+        var x2, y2;
+        x2 =  (this._x + this._origin.x - e.o.x) * e.cos + (this._y + this._origin.y - e.o.y) * e.sin + (e.o.x - this._origin.x)
```
potomak, line 939: I run `grunt check` and I noted you're missing closing `;` at the end of this line.

See #540.

PR #596, opened 2013-11-02
Quality of Life Improvement - Particles: Triggers event when particle animation is finished

mucaho:
e.g. you want to destroy bomb after it has exploded

``` javascript
var bomb = Crafty.e("2D,Canvas,Particles")
                 .particles(options)
                 .bind("ParticleAnimEnded", function() {
                    this.destroy();
                 });
```

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -51,6 +51,7 @@ Crafty.extend({
                 key: key,
                 value: value
             });
+            return value;
         }
     },
     /**@
--- a/src/loader.js
+++ b/src/loader.js
@@ -51,7 +51,6 @@ Crafty.extend({
                 key: key,
                 value: value
             });
-            return value;
         }
     },
     /**@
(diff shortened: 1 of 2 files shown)


starwed:
Just a few small thinngs:
- Don't bother to pass null when triggering the event; just `trigger("ParticleAnimEnd")` is fine
- Consider changing to "ParticleAnimationEnd", since we don't tend to use shortened words.  "ParticleEnd" would probably be ok?
- Is the `parentEntity` change on L16 related?

Also, could you rebase your branch and squash down to one commit?

mucaho:
Made the suggested changes.

`parentEntity` is necessary so that you can trigger the event on the entity that has the component added to it, rather than triggering the event on the child (== particle object) (in the bomb example: the "ParticleEnd" event will be triggered on the bomb entity)
If there was no reference to the bomb object, the event could only be triggered on the particle object (because of code scope), which isn't even a crafty entity but a generic object.

PR #597, opened 2013-11-02
Streamline Twoway input handling

mucaho:
Instead of polling the current keyboard state, use keyboard event data

PR #598, opened 2013-11-02
Use event data in control component Twoway

mucaho:
Instead of polling the current keyboard state, use keyboard event data

remake of #597

### DIFF
--- a/src/controls.js
+++ b/src/controls.js
@@ -916,8 +916,9 @@ Crafty.c("Twoway", {
                 this.y -= jump;
                 this._falling = true;
             }
-        }).bind("KeyDown", function () {
-            if (this.isDown("UP_ARROW") || this.isDown("W") || this.isDown("Z")) this._up = true;
+        }).bind("KeyDown", function (e) {
+            if (e.key === Crafty.keys["UP_ARROW"] || e.key === Crafty.keys["W"] || e.key === Crafty.keys["Z"])
+                this._up = true;
         });
 
         return this;


potomak:
If you don't want to use `isDown` I think you shoud also:
1. delete the `isDown` method (https://github.com/craftyjs/Crafty/blob/develop/src/controls.js#L629-L658), as it's used only at https://github.com/craftyjs/Crafty/blob/develop/src/controls.js#L920
2. remove `Crafty.keydown` (https://github.com/craftyjs/Crafty/blob/develop/src/controls.js#L12-L29 and https://github.com/craftyjs/Crafty/blob/develop/src/controls.js#L291-L299)

I think you could also remove https://github.com/craftyjs/Crafty/blob/develop/src/controls.js#L761-L768

starwed:
Knowing whether a key is down is actually super useful.  For instance, you might want a player to move only while a key is down; better for Crafty to track the key state than have everyone roll their own method.

So just because none of the built in components call it doesn't mean we can remove it.  (It might mean that some of those components _should_ be using it, of course!)

I generally dislike the way TwoWay, MultiWay, and Gravity work.  In the long run, they should be replaced by something a little more flexible.  But right now, this is a strict improvement over what we have.  The current behavior is buggy, because if you push any other key while holding down up the entity will jump.

potomak:
> So just because none of the built in components call it doesn't mean we can remove it. (It might mean that some of those components should be using it, of course!)

So why using event data if we already have the `isDown` method?

src/controls.js
```diff
@@ -916,8 +916,9 @@ Crafty.c("Twoway", {
                 this.y -= jump;
                 this._falling = true;
             }
-        }).bind("KeyDown", function () {
-            if (this.isDown("UP_ARROW") || this.isDown("W") || this.isDown("Z")) this._up = true;
+        }).bind("KeyDown", function (e) {
+            if (e.key === Crafty.keys["UP_ARROW"] || e.key === Crafty.keys["W"] || e.key === Crafty.keys["Z"])
```
potomak, line 920: If you run `grunt check` you should see a warning encouraging the dot notation.

See also #540.

mucaho:
@potomak 
Event data driven input is the better option if you want to have crafty running on a server. You don't have a keyboard state there.

Thanks for checking, but I don't quite follow. What is wrong and what should be changed?

potomak:
Ok, sorry but I didn't get this idea of running Crafty server-side, do you have an example or some sort of documentation about it?

About the `grunt check`, you should always run grunt check to run tests and jshint. More details at http://gruntjs.com/

mucaho:
I'm so glad you asked :) [Here](https://github.com/mucaho/npm_crafty) is a working example of minimal server-side Crafty. That's why I'm lobbying for the change to split the browser context from the source code. It would be immensely easier to automatically build a server-side crafty version without the manual fine-tuning.

Ok I will make new PR fixing those jshint warnings:
- add semicolon to @starwed change to rotation
- object["key"] to object.key from this PR

PR #601, opened 2013-11-03
Extension of Gravity component - moving platforms

mucaho:
In a platformer you sometimes have moving platforms. 
If you do not attach the player to the platform, he just falls off it when the platform moves away.
Therefor I added an optional argument `shouldAttach` to the `gravity` constructor.

Example usage can be seen in the qunit tests.

### DIFF
--- a/src/2D.js
+++ b/src/2D.js
@@ -1019,7 +1019,7 @@ Crafty.c("Gravity", {
     _gy: 0,
     _falling: true,
     _anti: null,
-	_attachedToEntity: null,
+    _attachedToEntity: null,
 
     init: function () {
         this.requires("2D");
@@ -1031,12 +1031,12 @@ Crafty.c("Gravity", {
      * @sign public this .gravity([comp], [shouldAttach])
      * @param comp - The name of a component that will stop this entity from falling
      * @param shouldAttach - The truth value if the entity should attach upon landing on platform.
-	 *
+     *
      * Enable gravity for this entity no matter whether comp parameter is not specified,
      * If comp parameter is specified all entities with that component will stop this entity from falling.
      * For a player entity in a platform game this would be a component that is added to all entities
      * that the player should be able to walk on.
-	 * If you have moving platforms set shouldAttach to true.
+     * If you have moving platforms set shouldAttach to true.
      *
      * @example
      * ~~~
@@ -1048,11 +1048,10 @@ Crafty.c("Gravity", {
      */
     gravity: function (comp, shouldAttach) {
         if (comp) this._anti = comp;
-		this._shouldAttach = !!shouldAttach;
+        this._shouldAttach = !!shouldAttach;
 
         this.bind("EnterFrame", this._enterFrame);
 
-		
         return this;
     },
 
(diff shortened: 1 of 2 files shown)


kevinsimper:
Looks really cool.

I will test it out later today.

Btw, you indentation seems a little out of place. The crafty.js file also uses 4 spaces.

mucaho:
@kevinsimper 
Thanks, yeah I noticed the identation was off, too.
I'm working with notepad++, so I have to find an option to convert tab to 4 spaces, right?
I have looked for an IDE real quickly, but I couldn't find a free one that suits me. What would you recommend me?

starwed:
I'm not sure this should be part of the "Gravity" component.  

Here's an idea: we make a component called "Supportable" that tracks what the entity is standing on.  It emits events every time that changes, and handles things like teleporting the entity to the top of the surface.  Then "Gravity" binds to these events to determine whether the object is falling or not.  

The automatic attachment to platforms could then either be handled by another component, or by "Supportable".  Or, perhaps an event could be triggered on the platform entity.  This seems like a cleaner implementation that's easier to hook into.

It would be a bit of work, so maybe should be punted on for now, if you just want to land something that works.

> I have looked for an IDE real quickly, but I couldn't find a free one that suits me. What would you recommend me?

I use Sublime Text 2, which is more of a text editor than an IDE but is still pretty powerful.  You can use the trial version as long as you like.

mucaho:
Thanks, will definitely check out Sublime.

Yes, I also gave a lot of thought how to integrate attaching onto platform. I thought this would be the most simple way to not change alot of the api. Future motivation was that this should be provided in the core api and not seperately available from craftycomponents.

But we can do a complete overhaul, I'm continuing your idea, let me know what you think:
- Making another component `PlatformDetector`( which does `map.search` (what gravity currently does)). This component triggers `onPlatform` and `offPlatform` events.
- Then make`Gravity` listen to events triggered by `PlatformDetector`. Depending on the current situation `Gravity` adjusts the entity's vertical position. (here we can include #592 also).
- Then we could make another component `PlatformAttacher` which also listens to `onPlatform` and `offPlatform`, attaching / detaching at these appropriate times.
- Grand finale would be to adapt `Twoway`'s jump to work with the above stuff. This depends on #592 also. We could let `Gravity` handle all vertical movements, so we could call `Gravity.jump(amount)` from `Twoway`. Essentially `Twoway` would `requires("Gravity")`.

starwed:
I think that is the general way things should work.

Moving platforms are actually kind of trickier than you'd think, though.  Using `attach()` and dealing with collisions correctly gets complicated in the general case.  (Not so bad if only one thing can sit on a platform.)

Anyway, closing the PR, but feel free to open an issue for discussing our strategy here.

mucaho:
@starwed 

> Moving platforms are actually kind of trickier than you'd think, though. Using attach() and dealing with collisions correctly gets complicated in the general case. (Not so bad if only one thing can sit on a platform.)

Did you try out the code or have some testcases where collisions will not get correctly applied?
I have co-developed a game some while ago. The game was under extensive testing and this functionality is taken directly from it. I just isolated the _attaching_ functionality again and tried it out, I can not find any misbehaviour.

What about overhauling the way Twoway and Gravity work together? Have you given some thought on it?

PR #602, opened 2013-11-03
fix recent jshint warnings

mucaho:
as described in #598


starwed:
Ok, simple fix so I'll merge.

I need to get into the habit of running `grunt check` instead of just the qunit tests!  :)

potomak:
:+1:

Issue #578, opened 2013-10-25
Logic separation of components


potomak:
I think that separation of concerns in components is great and it just works, but I'd like to see **only one component per file**.
Directories could be used to aggregate cohesive components.

See also #577.

dariusk:
I agree completely.

starwed:
> I'd like to see only one component per file.

Hmm, I'm not so certain about this as a hard rule.  Some components are very short, and depend heavily on other components.  For instance, should "FourWay" be in a separate file from "MultiWay"?  All the former component does is call multiway.  Having to constantly hop between files hurts readability just as much as having to navigate a huge monolithic file.

There are scores of components that _should_ be in their own files, though, so I agree in general.  And I think everything in its own file, while not ideal, is certainly an improvement over the current situation.

kevinsimper:
I think the problem is in the files which is over 700 lines, which is the 2D.js, controls.js, core.js, math.js files.

And as you point out, it does not make any sense to split out FourWay and MultiWay, but i do not think that was what @potomak was thinking either :)

potomak:
@starwed @kevinsimper in my opinion there should be one clear rule: **one file, one components**.

A simple solution to the `*way` components problem should be to put them in the same directory (_package_), maybe the `controls` directory.

starwed:
> A simple solution to the *way compoents problem should be to put them in the same directory (package), maybe the controls directory.

That's not a solution to the objection I had; in fact, putting the source code in subdirectories can actually make it harder to jump between files, depending on what sort of development environment you use.

potomak:
Note: one of the main reasons (a practical one) I'm asking for this _feature_ is that when I'm looking for a component I'm always forced to look for the file including it using the _search in project_ function of my editor.

potomak:
Putting more than one component per file is like putting more than one class per file, it's permitted in some languages. I write mostly ruby and java and it just feels wrong to me.

I _challenge_ you @starwed to put all `*way` components api documentation together in one single doc page as they're in a single js file :smile:

starwed:
In the small amount of time I spent with Java I hated exactly that, so I think there's a large subjective component to it!  :P

But clearly the current state is not great.  Why don't we start by identifying files that hold multiple, unrelated components.

One very basic separation we should consider is between code that depends on a browser context, and code that doesn't.  We should make it easy for someone to use Crafty code on the backend, as described in #440.

mucaho:
[...]

Issue #440, opened 2013-04-01
CraftyJS & NodeJS


mucaho:
I have made a node module, which runs a slim version of the current crafty master version (0.5.3). It's still in development: [npm_crafty](https://npmjs.org/package/npm_crafty)

According to a [Google Group Discussion about NodeJS & CraftyJS](https://groups.google.com/d/msg/craftyjs/n6tCNBqiA9U/RDJ_mELoYFQJ) there is already some motivation to allow easy integration of crafty into a node server.

I was able to filter out the bulk of "unneeded" stuff for a server. As you can see [here](https://github.com/mucaho/npm_crafty/tree/master/diff#manual-filtering), some manual fine-tuning is required. If you allow a bit of refactoring, the whole process could be more automatized and clean:
- the bold outlined components and Crafty features could be moved into seperate files

eldad87:
Hi,
I really want to see your project production ready so please let me know if I can help you develop it.

thanks!

mucaho:
Hi

The base goal is accomplished (logical communication between entities on server and clients). 
The pong example works fine on localhost. I hosted the pong example on a node server with xhr polling. Without handling the latency appropriately and introducing seperate game loops, the game is unplayable (see next paragraph).

The next level would be to further adjust input components, to introduce more game loops (network input/output loop, game state update loop) and to provide components for handling latency ( client-side: prediction & correction of client entities, interpolation of other entities; server-side: lag compensation).

I am not sure that providing these next level, high-level features would be such a good idea:
- A lot of work. I do not have the time for it. Partial implementation is bad.
- It would have to be implemented correctly: too much or too less abstraction -> user has to write his own networking code anyway
- I am not sure if crafty is used for fully fledged networked games. Better to write an entire, specific "engine" combining crafty & networking? Networking for turn-based games is entirely different than for shooter games.

If you are still interested, let's continue the discussion on the github repository of npm_crafty.

mucaho:
Just a quick update: [BNG Article: Real Time Multiplayer in HTML 5](http://buildnewgames.com/real-time-multiplayer/)

I have skimmed through the article, he has essentially using the above mentioned techniques for handling latency. It would just be a matter of translating his code into crafty components, right? :) 

I still have to figure out if/how he has made the seperate update loops. Javascript is single-thread and when I hosted the pong example on the web server I got a wierd issue: Input from local client froze all entities except the local ones.

eldad87:
[...]

PR #603, opened 2013-11-03
Server build - split browser & display context from rest of code

mucaho:
A first attempt to split them, mainly using the philosophy from [this document](https://github.com/mucaho/npm_crafty/tree/master/diff#removed-from-crafty);
-  ~~src files have their name appended with "_server" if they represent backend code.~~
- some files have been split in 2 parts, if a chunk of code didn't fit into either worlds:  
  *\*  ~~`extensions` into `extension` and `extensions_server` (has only Crafty.keys and Crafty.mouseButtons)~~  
  *\*  ~~`controls` into `input` and `controls_server` (has only Multiway, Twoway, Fourway)~~  
  *\*  ~~`animation` has had it's `SpriteAnimation` component removed and put into `sprite`~~  
  *\*  ~~`drawing` has had `Crafty.scene` & `zeroFill` removed and they were put into `scene_server`~~  
  **The specific have changed; see later posts!**
- dummy objects have been added (like `window`, `Crafty.viewport`, etc) before & after definition of crafty server build

The old build functionality is intact: all tests pass & api build works.

You can build the server version with `grunt server` (as you guessed only the **_server** files get included in the server build). Of course only some qunit tests pass at the moment.
You can run a simple example with `node test_server.js`.
There is much room for improvement: 
- first of all changing the code so that the backend code does not have to be included. (e.g. instead of calling a method directly on `Crafty.viewport`, we could trigger events that `Crafty.viewport` listens to)
- as much of the functionality has been omitted, extensive testing & special qunit tests are needed

Is this the correct approach? The more PRs get merged into develop branch, the harder it will be to merge something like this if we plan to.


mucaho:
see also #578 #440

starwed:
I think I'd prefer to do this in smaller steps.  For instance, lets do things like fixing calls to viewport _before_ we try to reorganize the code base.

We should think long and hard about the way to proceed; let's punt on this sort of large change until after the next release, which should actually be pretty soon.

Nice work in identifying the problem areas, though!

mucaho:
Ok, but remember this is just moving of code into different files. No "changes" to the code itself.
EDIT:
My point being that if we want to support backend-only code, then the code should be reorganized slowly into appropriate files (the sooner the better imo). It won't change anything to the normal crafty release, while we can take our time and tune the backend build in the background, without worrying to retrofit ongoing changes to the development branch.

mucaho:
On the other hand I'm certain there is a better way to reorganize the files.

mucaho:
I have given some more thought about how to better reorganize the files:
- put all **src** files into three folders (**src/common, src/backend, src/frontend**) using the `git mv` command which makes the changes more transparent
- split & create the **new src files** out of the old ones:
- [x] frontend/drawing.js - scene -> backend/scene.js  
- [x] frontend/controls.js - Multiway, Twoway, Fourway -> backend/controls.js  
- [x] common/animation.js - SpriteAnimation -> frontend/sprite.js  
- [x] frontend/extensions.js - zeroFill -> common/2D.js  
- [x] frontend/extensions.js - keys, mouseButtons -> common/keycodes.js  
- add the **backend setup** specific files
- update the paths in `require(core.js)` statements, in the **grunt build** file and in the **package.json** file

mucaho:
inspired by #763 :

I like the playgrounds very much, but on the other hand the playgrounds and any other _frontend_ functionality (`Color`, ...) won't work if Crafty's codebase is going to be split into _frontend(browser) and backend(node)_ #603 . The plan (for now) is to omit _frontend_ functionality from the _backend_ code & tests. The _backend_ code is almost ready, but many of the tests (that will be in the _backend_) have some kind of _frontend_ functionality (mostly `Color` components). Does anyone know an elegent way to deal with this?

mucaho:
I recently discovered [node-localstorage](https://www.npmjs.com/package/node-localstorage) which acts as a drop-in replacement for the browser `localStorage`.

``` javascript
if (typeof localStorage === "undefined" || localStorage === null) {
  var LocalStorage = require('node-localstorage').LocalStorage;
  localStorage = new LocalStorage('./scratch');
}
```

[Grunt node-qunit](https://www.npmjs.com/package/grunt-node-qunit) can be used to run the tests in a `node` environment, instead of the `phantomjs` environment provided by `grunt-contrib-qunit`.

mucaho:
Superseded by #902, closing.

PR #605, opened 2013-11-05
Use getTime method instead of implicit conversion

potomak:
This doesn't change anything in the behaviour, but it encourages good js
practices.

See also performances http://jsperf.com/date-timestamp

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -986,7 +986,7 @@ Crafty.extend({
             init: function () {
                 // When first called, set the  gametime one frame before now!
                 if (typeof gameTime === "undefined")
-                    gameTime = (+new Date()) - milliSecPerFrame;
+                    gameTime = (new Date().getTime()) - milliSecPerFrame;
                 var onFrame = window.requestAnimationFrame ||
                     window.webkitRequestAnimationFrame ||
                     window.mozRequestAnimationFrame ||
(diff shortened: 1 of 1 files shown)


starwed:
[Here](http://jsperf.com/date-gettime-valueof) is a jsperf that has some historical data etc; seems like `getTime` is faster in all the implementations tested.  And [this](http://jsperf.com/date-timestamp/2) is just to confirm that it's still faster the way we actually use it.

I'm going to leave this open for a bit in case someone knows _why_ we were using the implicit conversion.  Was it an IE thing?  Was it faster in older browsers?

starwed:
I can't find any reason to prefer the conversion over `.getTime()`.    So merging this as a strict improvement.

If we ever drop support for IE8, we could [just use `Data.now()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now).  (Or, we could add a compatibility shim as suggested.)   That would be [even faster](http://jsperf.com/date-timestamp/3), since it doesn't create any new Data objects.

kevinsimper:
@starwed We should use `Date.now()`, then we can use this shim for browsers than do not support it: 

```
if (!Date.now) {
  Date.now = function now() {
    return new Date().getTime();
  };
}
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now

potomak:
@starwed thanks for merging this.

@kevinsimper yes, we could use Date.now() if we add the support for it on old browsers, about this topic I opened a PR with a couple of fixes. I'm working on a game that should partially support IE8 and yesterday trying to debug the game on IE8 I found some minor incompatibilities.

**Note:** I only scratched the surface with #607, what about audio and DOM animations (they are completely broken)?

PR #607, opened 2013-11-07
Feature/ie8 support

potomak:
Latest version of Crafty breaks support for IE8, this is an attempt to fix it.

This PR could also be used to discuss about old browser support.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,8 @@
     "src/loader.js",
     "src/math.js",
     "src/time.js",
-    "src/DebugLayer.js"
+    "src/DebugLayer.js",
+    "src/support.js"
   ],
   "dependencies": {
     "grunt": "~0.4.1",
--- a/src/import.js
+++ b/src/import.js
@@ -24,14 +24,14 @@ Crafty['import'] = function (obj, scene) {
     //if its a string, load the script file
     if (typeof obj === "string") {
         if (levelData) {
-            if (scene) Crafty.import(levelData[scene]);
-            else Crafty.import(levelData);
+            if (scene) Crafty['import'].apply(levelData[scene]);
+            else Crafty['import'].apply(levelData);
         } else {
             var elem;
             elem = document.createElement("script");
             elem.onload = function () {
-                if (scene) Crafty.import(levelData[scene]);
-                else Crafty.import(levelData);
+                if (scene) Crafty['import'].apply(levelData[scene]);
+                else Crafty['import'].apply(levelData);
             };
             elem.src = obj;
         }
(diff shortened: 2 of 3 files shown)


starwed:
[Here](https://developer.mozilla.org/en-US/docs/Project:MDN/About?redirectlocale=en-US&redirectslug=Project%3ACopyrights#Copyrights_and_licenses) is info on using the MDN code snippets.  Looking at the history of the `indexOf` page, I think the current code would be public domain?  (There was an earlier version of the snippet, but it seems to have been completely rewritten by now.)  If its not public domain, it'll be MIT.

kevinsimper:
@potomak I think that the work for getting IE8 working is not worth it, and as you point out the Audio does not work either. A lot have changed, and the new Web Audio API looks cool, which Firefox just now supports.

It is not that much that it is hard to support old browsers, but it makes it more difficult to contribute, because of all the legacy code you have to now and look at.

starwed:
I'm ok with merging this right now, but unless someone volunteers to step up, I think we should probably remove IE8 support in the future.

I get the impression that there's a lot of broken behavior in IE8.  I had trouble even getting anyone to check whether crafty _worked_ in IE8, so I'm not very hopeful about anyone actually fixing stuff at this point.    [Here is the IE thread](https://groups.google.com/forum/#!msg/craftyjs/ePAkLnHs1Bc/mcZcr76R2UQJ) in the google group.

potomak:
I totally agree with you @starwed but I also think that Crafty was cool also because it supported IE8 lack of canvas element by offering a chance to render stuff using the DOM component.
Now if we all want to drop support for IE8 I also think that we could remove a lot of DOM specific code.
It sounds a little bit radical but it's just a proposal.

kevinsimper:
I vote for not merging this, because IE8 is on the way out, and as pointed out, nobody test if it actually works in IE8. Debugging in IE8 is also a pain for that sake.

@potomak I should still be possible to do, because of lot of things like is done easier with DOM like textarea's and input fields, also text sometimes.

sbyrnes321:
@potomak -- there are lots of good reasons for continuing to support DOM rendering. Selectable text ... the performance is better under certain conditions ... helps for combining crafty with stuff from other libraries (like jquery-ui) ... helps for combining crafty with native DOM elements like text boxes, buttons, etc. ...

I have no opinion about whether / when to throw out IE8 support. Whatever y'all prefer, it's fine with me.

starwed:
So does Crafty basically work in IE8 after this patch?  Or would it require more fixes to get simple games working?

Trying to assess whether we should include it in the 0.6 release or not.

potomak:
It fixes only a couple of major errors, but I didn't run tests on IE8.
I'm sorry but I have only limited access to IE8 machines.

starwed:
Closing due to #639.

PR #608, opened 2013-11-07
viewport for IE9/10, let it can init again.

TrigoTao:
I need to use Crafty.stop(), and then use init to restart it, actually the point is to init() again. It seems strange, but I encounter a situation that the Crafty appears in one div, then it must appear in another div. When it is in IE10, I can't init for a second time just because configurable default to be false.

### DIFF
--- a/src/viewport.js
+++ b/src/viewport.js
@@ -629,7 +629,8 @@ Crafty.extend({
                     },
                     get: function () {
                         return this._x;
-                    }
+                    },
+                    configurable : true
                 });
                 Object.defineProperty(this, 'y', {
                     set: function (v) {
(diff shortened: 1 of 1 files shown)


starwed:
Looking at some [documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty) about `configurable`, this seems pretty safe.

TrigoTao:
Sorry, I'm late. Yeah, you are right. But it acts different between browsers. So I want to modify it.

starwed:
> Sorry, I'm late. Yeah, you are right. But it acts different between browsers. So I want to modify it.

Not sure what you mean...  Are you saying you want to modify the PR, or just explaining why you submitted it?

TrigoTao:
Just an explaination , I misunderstood your words before... So, just forget it... :-)

PR #609, opened 2013-11-07
Add `avoidCss3dTransform` flag

potomak:
CSS 3D Transforms are used to set DOM entities position, this should increase performances as these operations are hardware-accelerated but using them for positioning makes CSS effects unavailable.

Setting this flag to true you can force DOM entities positioning using left and top CSS properties.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -92,7 +92,8 @@ module.exports = function (grunt) {
                 'tests/math.html',
                 'tests/isometric.html',
                 'tests/loader.html',
-                'tests/text.html'
+                'tests/text.html',
+                'tests/dom.html'
             ]
         }, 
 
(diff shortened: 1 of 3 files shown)


starwed:
Add something to the docs about it?

potomak:
Ok, @starwed can you please suggest me where should I put it in the docs?

starwed:
If I understand this correctly, it's an optional property of the DOM component, right?  So it should be documented there.  You'd want to add it explicitly with a default value of false, and document it in place.  Something like:

``` javascript

/**@
* #.avoidCss3dTransforms
* @comp DOM
* A flag that (does stuff) when true.  Defaults to false.
*
*/

avoidCss3dTransform: false,
```

potomak:
Thanks for the explanation!

potomak:
@starwed I documented the flag and I also added a test suite.

starwed:
Ok, seems good to me.   Tests pass, etc.

The different casing convention between `css3dtransform` and `avoidCss3dTransform` is a little weird when they're next to each other, but the latter seems more consistent with how we name properties.

potomak:
@starwed yes, I thought the same but as you stated the latter is more consistent with other properties naming style.

PR #610, opened 2013-11-11
Add a bit of functionality to Crafty.audio

starwed:
This makes it so that `Crafty.audio.play` returns the actual audio element used to play the sound.

I also fixed a bug in how `audio.stop()` works; it wasn't properly updated when we switched to the channels approach.

There are probably some other, simple things we could do to make the audio stuff easier to use; since we might switch to using the web audio API in the future, it would be ideal if you never _needed_ to manipulate the audio elements directly.

So anyone have any specific ideas for small improvements?

### DIFF
--- a/src/sound.js
+++ b/src/sound.js
@@ -189,10 +189,11 @@ Crafty.extend({
          * @comp Crafty.audio
          * @sign public this Crafty.audio.play(String id)
          * @sign public this Crafty.audio.play(String id, Number repeatCount)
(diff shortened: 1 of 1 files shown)


potomak:
I was waiting for this and I thought it was already merged, I think I'm going to merge it in a branch of mine...

starwed:
I was just waiting to see if anyone had any related suggestions before merging.  I'll go ahead and merge now.

PR #611, opened 2013-11-13
Fix debug polygon drawing

samccone:
The polygon is always 4 points, where if someone passes in a custom hit area with less than 4 points or more than 4 points the polygon debug drawing will fail, or only draw 4 of N points.

This is not a breaking change, only a bug fix

### DIFF
--- a/src/DebugLayer.js
+++ b/src/DebugLayer.js
@@ -232,7 +232,7 @@ Crafty.c("DebugPolygon", {
     drawDebugPolygon: function () {
         ctx = Crafty.DebugCanvas.context;
         ctx.beginPath();
-        for (var p in this.polygon.points) {
+        for (var p in this.map.points) {
             ctx.lineTo(this.map.points[p][0], this.map.points[p][1]);
         }
         ctx.closePath();


starwed:
Actually, "DebugPolygon" is intended to be useful for more than hit areas.  So this needs to be a change to "WiredHitBox" and possibly "Collision".

The current behavior is that you need to call `.matchHitBox()` anytime the collision polygon is set.  One fix would be to create a new event fired by "Collision" when the polygon is changed.  While there are other ways to approach this, it would be generally useful to know when the hitbox changes.

samccone:
Hmm, so what about this from within the debug method

``` coffeescript
@matchHitBox() if (@map.points.length isnt @polygon.points.length)
```

starwed:
There are other reasons the hitbox might change, so that's not a good general fix.   There isn't anything special about 4 points; it's just not auto-updating the polygon when you pass in a new one, regardless of the shape.

samccone:
fair but... 
https://github.com/samccone/Crafty/commit/de0fa6e37a3613d45e258187190afa2a1feccac5#diff-389e4d267bfc809989c10c5969cbf7d5R235

will actually error out if your points are < 4
we should ideally prevent that... right?

starwed:
Hmm, there shouldn't be any error there.  I was using it on triangular hitboxes just last night!

Could you explain what error you were encountering?

samccone:
yeah ok, So... update your hitarea to something

like in your 

``` coffee-script
init:->
  @collision([34.225,0], [88.311,0], [122.535,36.224])
```

and have the debug rect turned on

You should see the error.

starwed:
What was the error message?  I have code that does that and I don't see any error.

(Or did you just mean the issue with the hitbox not being updated.)

samccone:
hmm mine was actually failing on drawing the polygon, since mine only had 3 points and it looped over all 4 points of the default bounding box.

starwed:
_Oh_, I see the problem.  It should be referencing the polygon object inside the loop, not the map object.  Sorry I didn't see this earlier -- I guess I should never try to debug without coffee.  :)

samccone:
:coffee: ! :dancer:

samccone:
So I think the PR is still valid right?

starwed:
More the opposite of this particular PR -- I'll submit a patch in a sec.

samccone:
Ah I see your point! k cool thanks!

PR #612, opened 2013-11-13
Fix, for applying a flat array to a rotated model

samccone:
The evaluation order was incorrect if you passed in a flat array for the collision area if the model was already rotated.


starwed:
Looks good, and it passes the unit tests/etc.  Thanks!

samccone:
:thumbsup:  hey @starwed  just fyi we ran into these issues making
http://game.zoomtheballoon.com/
during a 48 hour hackathon

PR #615, opened 2013-11-13
Logic error in "DebugPolygon"

starwed:
Currently DebugPolygon iterates through the designated polygon, but uses the equivalent points in the hitmap for drawing.  This can cause errors when they're not the same length, in addition to just being wrong!

Also, test that the polygon exists before trying to draw it.


starwed:
Passes unit tests, linting, and seems to work correctly in practice.  Landing.

PR #617, opened 2013-11-14
Add multijump ability.

pawurb:
I have added a function to Twoway component which gives character an ability to jump multiple times. We specify how many times he can jump, how high and if jumps can be started when he is not on solid ground. 

If function is not executed than the old behaviour is preserved. Changes in 2d component are not going to cause error even if Twoway component will not be activated.

That is how it looks (viewport is following the character):
http://www.youtube.com/watch?v=cX_uHQbVrgE

### DIFF
--- a/src/2D.js
+++ b/src/2D.js
@@ -1124,6 +1124,7 @@ Crafty.c("Gravity", {
         //this._gy = -1 * this._bounce;
         this._falling = false;
         if (this._up) this._up = false;
+        this._multijumpsLeft = this._multijumpsLimit;
         this.trigger("hit");
     },
 
--- a/src/controls.js
+++ b/src/controls.js
@@ -921,10 +921,37 @@ Crafty.c("Twoway", {
                 this._falling = true;
             }
         }).bind("KeyDown", function (e) {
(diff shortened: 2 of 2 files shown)


kevzettler:
Hey this is cool! However, I don't feel like this belongs in the Crafty core. This would be great as a standalone module and loaded through the crafty components. However on that thought does twoway or multiway even belong in Crafty core? 

This is probably grounds for a larger discussion on the crafty architecture. Maybe crafty could utilize something like Grunt does with plugins and using NPM.

potomak:
Nice feature but I totally agree with @kevzettler also about how to load components in your Crafty projects

starwed:
Yeah, I think this is maybe a bit too specific for right now.  The current platforming components are so inflexible that adding functionality in this kind of _ad hoc_ way isn't good in the long term; better to rewrite them in a way that makes them easier to extend, compose and modify.

(Also, the better way to do this would be to have the "Gravity" component trigger an event when the entity lands.  That way the components don't have to know anything about each other directly.  That's good because it makes them more flexible; a component should only ever manipulate another components methods/properties directly if it always requires that component.)

I'm closing this for now, but thanks for the PR!

PR #619, opened 2013-11-15
Scene parameters

starwed:
This is based on #508, and allows a scene to be called with an optional parameter, which is then passed to the init function of the scene.

It also adds unit tests for both the new and original functionality of Crafty.scene.

I also moved `Crafty.scene` into a new file, `scene.js`.  I have no idea why it was in `drawing.js`; it has nothing to do with drawing...

thanks for the original PR, mikepaulson!


potomak:
:+1: for tests
:+1: for scene parameters
:+1: for moving scenes out of `drawing.js`

kevinsimper:
I think now that the function scene both take two parameters, when creating scenes and changing scene, they should be called something else. The only way now you can see if it creates a scene or calling it, if it is a function, instead of a object or likewise.

I think it should be easier to distinguish. Like:

```
Crafty.scene("mountain", function(data){
  Crafty.e("2D, DOM, Text")
    .attr(data)
    .text("Loading")
    .css({ "text-align": "center"})
    .textColor("#FFFFFF");
});

Crafty.playScene("montain", { w: 100, h: 20, x: 150, y: 120 });
```

This way it would be easy to see what is going on.

potomak:
@kevinsimper I'm not a fan of function overloading in JavaScript because it _doesn't exist_, that's why I totally agree with you.

starwed:
Ok, I'd suggest that for now we mark `Crafty.scene` as deprecated, but still let people call it.  What should the new functions be called?  Random suggestions:
- runScene, gotoScene, playScene, setupScene
- defineScene, createScene, makeScene

I'd currently lean towards define and run, but my preferences here are pretty weak.

> I'm not a fan of function overloading in JavaScript because it doesn't exist, that's why I totally agree with you.

Well, it doesn't exist in _any_ language when you look at the machine code, right?  At some point the type of the arguments is examined and a particular logic path is chosen; we just have to do it manually and dynamically in js.  :)    From an API standpoint the actual implementation shouldn't be a concern; we just care whether the interface might be confusing or not.  (It's true that in type-checked languages you remove some class of errors here.)

I suspect that Crafty's use of overloading is a case of copying how jQuery does things -- it's pretty clear that a lot of the original design decisions were inspired that way.  (And since jQuery is probably the single most commonly used javascript library, it probably isn't such a terrible idea.)

kevinsimper:
Okay, that would be cool.

I like that you `define` a scene and then `run` it later.

I think that function overload can be fine with getter and setter function, but here with scene function i think it is now something else.

kevinsimper:
I have made the scenes we discussed about in #625. You can check it there, where it also contains @starwed code.

starwed:
Ah, i also have made those changed in a local branch.  I was just trying to decide what the docs should look like. :)

Also, your branch gets rid of Crafty.scene -- I'd prefer that it stick around for now, for backwards compatibility.

starwed:
I went ahead and pushed my local branch for comparison.  I went with `enterScene` rather than `runScene`, just because I liked the sound of it better.

kevinsimper:
That's cool :+1: , i would just help since it was a couple of days ago we talked about it, and yes I think also the old scene function should be there, I forgot it.
I do think though that there should be a console.warn that it is deprecated, and they should use the other functions.

And what do you think about the test for the uninit scene when the scene change?

```
    test("Scene uninit called", function(){
        var x = 0;
        var y = 0;
        var sceneInit = function(){ x = 13;}
        var sceneUninit = function(){ x = 20;}
        var sceneGame = function(){ y = 5;}
        Crafty.defineScene("test-uninit", sceneInit, sceneUninit);
        Crafty.defineScene("game", sceneGame);
        Crafty.runScene("test-uninit");
        Crafty.runScene("game");
        equal(x, 20, "Uninit scene called successfully when chanced to another scene");
    });

```

kevinsimper:
I will make this console.warn later and then merge it.

starwed:
Ok, updated the pull request with a test and a couple typo fixes.  Three separate commits because the three sets of changes are pretty independent of each other.

Passes `grunt check`.

PR #620, opened 2013-11-15
Make "Text" work better with canvas

starwed:
This is based on @kcaze's patch #499.  Thanks for the original PR!

The major change is that a "Canvas, Text" entity will automatically resize when either its font or text are changed.
This is accomplished by using measureText() to get the width, and estimating the font size in pixels for the height.
Currently the width is adjusted _after_ drawing, and the height not at all.  This often caused the text to be rendered with portions cut-off.

Text is now also drawn at the (x,y) coordinate of the entity, with the baseline set to "top".  Previously text was translated to the bottom left corner of the entity, resulting in descenders getting cut-off.

Tests were added for the new resizing feature.


sbyrnes321:
Works for me! Thanks! Of course, I had to shift the canvas text in my game upwards to compensate for the new positioning.

PR #621, opened 2013-11-15
Remove .setVersion on IndexedDB as it has been deprecated.

starwed:
Based on #518 by @chadams.  Removes a deprecated method of opening an IndexedDB datastore.

I have no idea if this actually works, just wanted to update the PR so it was in a state that could be tested.


kevinsimper:
This can be closed now since #692. If anybody is up for building a version of IndexedDB into Crafty.js, it will make a great plugin.

PR #622, opened 2013-11-17
Image Component Fixed

luizbills:
Now **Image** component can resize/repeat an image with Canvas or DOM.

Tests: https://dl.dropboxusercontent.com/u/23008941/image-fix-test.zip

### DIFF
--- a/src/drawing.js
+++ b/src/drawing.js
@@ -109,7 +109,7 @@ Crafty.c("Image", {
     init: function () {
         this.requires('2D');
         if (!this.has('DOM')) this.requires('Canvas');
-        this.bind("Draw", _imgDraw).bind("RemoveComponent", _imgRemove);
+        this.bind("Draw", this._imgDraw).bind("RemoveComponent", this._imgRemove);
     },
 
     /**@
@@ -149,10 +149,10 @@ Crafty.c("Image", {
         this.img = Crafty.asset(url);
         if (!this.img) {
             this.img = new Image();
-            Crafty.asset(url, this.img);
 
             var self = this;
             this.img.onload = function() {
+                Crafty.asset(url, self.img);
                 self._imgConfig();
             }
             this.img.src = url;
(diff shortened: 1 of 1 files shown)


starwed:
This doesn't seem to bind to the "Draw" event at all to render itself, and actually overrides the Canvas/DOM `draw` method.

Also, it's unclear from the PR what bug this fixes?

luizbills:
With the actual Image component you can't do much things, as sketch a image.
I'm just trying fix a bug when you try resize a image.

kevinsimper:
@luizbills Could you make a test that your code passes? We are trying to make more test coverage, so fewer errors is made in the future.

starwed:
To be clear, we can't accept this right now -- it's not compatible with the technique Crafty uses for rendering.  In brief, the steps for drawing a particular entity are:
- The `draw()` method of that entity is called.  This is defined by either the  "Canvas" or "DOM" component
- The `draw` method sets up the generic 2D properties; it sets things like the alpha level and rotation.
- It then triggers an event called "Draw"
- Individual components like "Image" are listening to this event.  The render function specific to the component is bound to it.  

This patch directly overrides the `draw` method for Canvas component, rather than binding to "Draw".  This makes it so that its no longer compatible with the various generic 2D properties.

luizbills:
Oh sorry, I go fix it soon.

starwed:
I haven't had time to look at what this changes in detail, but the unbind function won't work as written.  It tries to unbind a function called `draw`, which doesn't exist.

Was there a specific reason why you're making the draw routine a global rather than defining it locally?  If you look at other components, that's generally how they work.

Thanks for the PR, I just want to make sure it plays well with the existing code we have.  :)

luizbills:
@starwed 
sorry, the unbind is now correct. And now works perfectly with or without loaded images.

I use global (now is a member of component) functions because I want avoid to create the same function multiple times, saving memory. See the example:

``` javascript
var a = {msg:'hello'}, b = {msg:'hello'};

// test #1 - using a locally created function
function extend(obj) {
    obj.test1 = function() {
        console.log(this.msg);
    }
}

extend(a);
extend(b);

a.test1(); //=> hello
b.test1(); //=> hello
console.log(a.test1 === b.test1); //=> false

// test #2 - using an existing function
function extend2(obj) {
    obj.test2 = test2fn;
}
var test2fn = function() {
    console.log(this.msg);
}

extend2(a);
extend2(b);

a.test2(); //=> hello
b.test2(); //=> hello
console.log(a.test2 === b.test2); //=> true
```

luizbills:
up

src/drawing.js
```diff
@@ -103,32 +103,13 @@ Crafty.c("Tint", {
...
-                    e.style.background = "url(" + this.__image + ") " + this._repeat;
-            }
-        };
-
-        this.bind("Draw", draw).bind("RemoveComponent", function (id) {
-            if (id === "Image") this.unbind("Draw", draw);
-        });
+        this.requires('2D');
+        if (!this.has('DOM')) this.requires('Canvas');
+        this.bind("Draw", this._imgDraw).bind("RemoveComponent", this._imgRemove);
```
kevinsimper, line 112: Is there a reason why you do not use the normal RemoveComponent function?

```
this.bind("Draw", draw).bind("RemoveComponent", function (id) {
    if (id === "Image") this.unbind("Draw", draw);
});
```

src/drawing.js
```diff
@@ -163,41 +144,89 @@ Crafty.c("Image", {
...
+            var style = this._element.style;
+            style["background-image"] = "url(" + this.__image + ")";
+            style["background-repeat"] = this._repeat || 'no-repeat';
         }
+        this.ready = true;
+        this.trigger("Change");
+    },
 
+    // Draw event 'Image' callback 
+    _imgDraw: function (e) {
```
kevinsimper, line 188: This function should have the name Draw like all other Crafty components.

src/drawing.js
```diff
@@ -163,41 +144,89 @@ Crafty.c("Image", {
      */
     image: function (url, repeat) {
         this.__image = url;
-        this._repeat = repeat || "no-repeat";
+        this._repeat = repeat;
```
kevinsimper, line 147: You should add a default option, if you do not specify anything.

src/drawing.js
```diff
@@ -163,41 +144,89 @@ Crafty.c("Image", {
...
+    },
 
-        return this;
+    // Remove event 'Image' callback 
+    _imgRemove: function(id) {
+        if (id === "Image") {
+            this.unbind("Draw", this._imgDraw);
+            this.unbind("Remove", this._imgRemove);
+            if (this.has('DOM')) {
+                this.undraw();
```
kevinsimper, line 225: The DOM component already undraw when it receives the RemoveComponent event.

https://github.com/craftyjs/Crafty/blob/dfc66d328e659ec46a640dc00c1e3abf8e1199a5/src/DOM.js#L81

kevinsimper:
I made some notes :)

You should also, if you can, squash the commits together to one when you are done.

src/drawing.js
starwed, line 188: Most Crafty components use a _local_ function named draw.  If it's going to be a method on the component, calling it draw will clobber the main Canvas/Dom methods.   I don't see any advantage to the local method, and it does end up creating a new function for every single entity -- all else being equal that seems inefficient.

 Though maybe use `_imageDraw`, to promote a standard naming convention?

src/drawing.js
starwed, line 112: Basically same as the other point; doing this creates anonymous functions for every entity, rather than a single shared one.  Not the end of the world, but I don't see any advantage to doing it the inefficient way.

We actually now have a `remove` function (basically the opposite of init)  to cleanup, so instead of binding `_imgRemove` to "RemoveComponent" you could just define `remove` as

``` javascript
remove: function(){ 
    this.unbind("Draw", this._imgDraw); 
    this.trigger("Change") 
}
```

It'll only be called when this specific component is removed, so you don't need to check for that.

kevinsimper, line 112: Ahh, yes, that is right, i can see that now.

src/drawing.js
kevinsimper, line 188: That is right, so should the sprite component also be changed then? Because that does create a local function, but as what I can see, that function could easily also be global :)

> Though maybe use _imageDraw, to promote a standard naming convention?

:+1:

starwed, line 188: Yeah, all the components like that (Sprite, Color, etc) could be switched, but it's not something I'd consider a pressing matter.  There might be some places where the local function binds some variables to its closure, in which case it won't be as easy to switch over.

starwed:
Closing, since this PR is against a pretty old version of Image.

PR #624, opened 2013-11-19
Bug Fix: Don't clobber background CSS property for the Image component

glaksmono:
Expected: when adding a CSS property to an image entity, it should apply the CSS to the image element

Example: Crafty.e("2D, DOM, Image").image("bg.png").css({"background-size": "contain"});

Currently: background-size would not be added to the image element because the general background property wouldn't allow for specific background property

### DIFF
--- a/src/drawing.js
+++ b/src/drawing.js
@@ -121,8 +121,10 @@ Crafty.c("Image", {
                 context.fillRect(0, 0, this._w, this._h);
                 context.restore();
             } else if (e.type === "DOM") {
-                if (this.__image)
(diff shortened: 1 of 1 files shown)


ryansmith3136:
:+1:

ryansmith3136:
Also related: #307

starwed:
D'oh, Firefox doesn't like the style object to be set with `style["dashed-property"]`.  It wants the syntax `style["dashedProperty"]` instead.

This causes "DOM, Image" elements to fail in Firefox, but work in Safari/Chrome.  Submitting a follow-up fix.  ( #637 )

PR #625, opened 2013-11-20
Split up the scene function to defineScene and runScene

kevinsimper:
As discussed in #619, I have split up the function scene, because now when the scene takes two arguments for playing a scene, and takes the same when defining a scene, it could easily be confused exactly what was going on.

Now there is a `defineScene`, which as the name suggest defines a scene, and a `runScene`.

I also created another test, to ensure the uninit function is getting called probably.


starwed:
Since #619 got merged, closing this.

PR #626, opened 2013-11-25
Fix tweening of same properties by different entities

Dandandan:
When tweening entities at the same time, only the last entity actually tweens because tweenGroup is global instead of local to the entity. This commit is a fix for this issue. It also makes tweenStart local and removes the unused _step and _numProps variables.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -93,7 +93,8 @@ module.exports = function (grunt) {
                 'tests/isometric.html',
                 'tests/loader.html',
                 'tests/text.html',
-                'tests/dom.html'
+                'tests/dom.html',
+                'tests/tween.html'
             ]
         }, 
 
--- a/src/animation.js
+++ b/src/animation.js
@@ -563,12 +563,10 @@ Crafty.c("SpriteAnimation", {
  * Component to animate the change in 2D properties over time.
  */
 Crafty.c("Tween", {
-	_step: null,
-	_numProps: 0,
-	tweenStart:{},
-	tweenGroup:{},
 
 	init: function(){
+		this.tweenGroup = {};
+		this.tweenStart = {};
 		this.tweens = [];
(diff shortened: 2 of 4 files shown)


starwed:
Ah, not literally a global, but shared by components.  Nice catch.

You've got mixed spaces and tabs on the lines added to `init`, which jshint complains about.  Could you fix those?  (You can run the linter and existing qunit tests with `grunt check.)  If you feel like adding a unit test for multiple entities being tweened, all the better!  :)

Dandandan:
Thank you for your quick response. Indeed, shared by components is what I meant.
I fixed the identation and added tests to test for tweening one and two entities.
Is it possible to give credit to "Vorm - Serious games" somewhere?
Thank you in advance.

starwed:
Thanks!  Just another small request: there are already some tween tests in core.html.  Could you move them in to tween.html so that everything is in one place?  (Or move your added tests to core.html, but core is pretty bloated.)  And then preferably squash down to a single submit so that the merge history is cleaner.  (Not the end of the world if you don't.)

> Is it possible to give credit to "Vorm - Serious games" somewhere?
> Thank you in advance.

I guess we don't really have an official policy on credits, do we?  We should compile a list and keep it in the repo; I imagine git can auto-generate something useful?  

I'll make sure the credit to Vorm gets in there, but for the 0.6 release it might be via a link to a list of contributors.

starwed:
Could you do a rebase to squash the commits down?  The basic form is `rebase -i HEAD~N`, where N is however many commits back you need to go; then you choose which commits to squash.

(And for this sort of thing it's ok to rebase against develop, rather than merging it in.  It helps keep the history of the main repo a bit cleaner.  Unless you're collaborating with someone else from your repo!)

starwed:
Thanks!

PR #627, opened 2013-11-27
Fixed bug in sound

jamespic:
Expanding channels without actually adding channels causes errors, as it tries to call methods on undefined.


jamespic:
Closing this. Will create a new pull request for a slightly more ambitious change

PR #628, opened 2013-11-27
Chrome workaround and bug in setChannels

jamespic:
A couple of patches.

One for a bug in setChannels (we unnecessarily increase the size of an array, without populating it - as a consequence, getOpenChannel finds undefined where it was expecting a channel).

Another is a workaround for a bug in Chrome (https://code.google.com/p/chromium/issues/detail?id=280417). Audio objects sometimes don't fire "ended" events for short sounds, so we leak channels. The patch allows us to re-use channels that haven't been marked as inactive, but are ended and over their repeat count.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -93,7 +93,8 @@ module.exports = function (grunt) {
                 'tests/isometric.html',
                 'tests/loader.html',
                 'tests/text.html',
-                'tests/dom.html'
+                'tests/dom.html',
+                'tests/sound.html'
             ]
         }, 
 
--- a/src/sound.js
+++ b/src/sound.js
@@ -262,9 +262,8 @@ Crafty.extend({
         maxChannels: 7,
         setChannels: function (n) {
             this.maxChannels = n;
-            if (n > channels.length)
+            if (n < channels.length)
                 this.channels.length = n;
-
         },
 
         channels: [],
@@ -277,7 +276,7 @@ Crafty.extend({
                 }
             }
             // If necessary, create a new element, unless we've already reached the max limit
-            if (i <= this.maxChannels) {
+            if (i < this.maxChannels) {
                 var c = {
                     obj: this.audioElement(),
                     active: true,
--- a/src/sound.js
+++ b/src/sound.js
@@ -217,6 +217,7 @@ Crafty.extend({
             if (!c)
                 return null;
             c.id = id;
+            c.repeat = repeat
             var a = c.obj;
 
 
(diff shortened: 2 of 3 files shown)


kevinsimper:
Looks nice :+1: 

We generally need more test, so would it be possible to write a test for this problem?

jamespic:
Writing a test for the maxChannels issue shouldn't be a problem.

The Chrome workaround is for a bug that only affects certain versions, platforms and hardware, so I'm tempted to leave that test-less, although I guess I could try using a mock Audio object that doesn't fire "ended" events, to simulate the issue

--- Original Message ---

jamespic:
I've added some unit tests. They're a little mucky (I'm using a fake Audio object, so it works on PhantomJS, and so we can emulate the Chrome bug it works around), but they fail before the changes, and pass after them

Gruntfile.js
```diff
@@ -93,8 +93,17 @@ module.exports = function (grunt) {
                 'tests/isometric.html',
                 'tests/loader.html',
                 'tests/text.html',
-                'tests/dom.html'
-            ]
+                'tests/dom.html',
+                'tests/sound.html'
+            ],
+//            debug: {
```
kevinsimper: Could these comments be removed?

starwed:
I think its ok in a case like this to just test the functionality.

Does PhantomJS really not support audio?  It supports every other DOM or Canvas test I've checked.

starwed:
> One for a bug in setChannels (we unnecessarily increase the size of an array, without populating it - as a consequence, getOpenChannel finds undefined where it was expecting a channel).

The code in question is this:

``` javascript
this.maxChannels = n;
if (n > channels.length)
    this.channels.length = n;
```

It's supposed to shrink the size of the array when you make it smaller, but it looks like I flubbed the comparison.  Rather than remove it, maybe we should change the `>` to a `<`.  

Also, could you squash the commits down when you think this is ready to be merged?  Unless you have someone else working off your repo, it's ok to just force push -- that seems to be github's intended workflow.

jamespic:
PhantomJS doesn't implement an `Audio` object, and `<audio>` elements don't implement `play` (I double-checked). Crafty marks audio as unsupported on Crafty. Even if it did, I'd be inclined to test with mock objects anyway, as it avoids browser bugs.

I'll change the setChannels code as-per your suggestion, and squash the commits.

starwed:
Ok, looks like the tests currently fail because `channels.length` should actually be `this.channels.length`.  Clearly this (the previously existing code) wasn't tested very well before... :)

starwed:
Also, jshint complains about
-  missing semicolons on lines 220, 274
- the line break on 280 where you inserted the comment mid conditional.  Probably easier to read if you describe the check before the conditional.

Sorry for the nitpicks, but in the long run it helps future contributors if we have a consistent style.  :)  You can run the linter with `grunt jshint` from the commandline.

starwed:
Ok, I pulled this into my local repo,  fixed that one `this.channels` bug, rebased, and then pushed it to the develop branch without a PR associated.    `grunt check` passes all tests/etc.

Thanks for the patch!

Issue #630, opened 2013-11-28
Twoway and Gravity should trigger a "Moved" event


heniotierra:
Like Multiway does, Twoway and Gravity should also trigger a "Moved" event

heniotierra:
Correcting the second commit's comment: Gravity triggers "Moved" event when falling

kevinsimper:
Helping on #631

Issue #629, opened 2013-11-28
Twoway should require Gravity


heniotierra:
Twoway does not make sense without gravity, so I think the first should require the later.

kevinsimper:
#631 discussing there

PR #631, opened 2013-11-29
Twoway requires Gravity; Twoway and Gravity trigger a "Moved" event

heniotierra:
addressed issues https://github.com/craftyjs/Crafty/issues/629 and https://github.com/craftyjs/Crafty/issues/630, opened by myself, making Twoway require Gravity, and also making Twoway and Gravity trigger a "Moved" event

### DIFF
--- a/src/2D.js
+++ b/src/2D.js
@@ -1012,6 +1012,8 @@ Crafty.c("2D", {
 /**@
  * #Gravity
  * @category 2D
+ * @trigger Moved - When entity has moved on y-axis a Moved event is triggered with an object specifying the old position {x: old_x, y: old_y}
+ * 
  * Adds gravitational pull to the entity.
  */
 Crafty.c("Gravity", {
@@ -1079,6 +1081,7 @@ Crafty.c("Gravity", {
             //if falling, move the players Y
             this._gy += this._gravityConst;
             this.y += this._gy;
+	    this.trigger('Moved', { x: this._x, y: this._y - this._gy });
         } else {
             this._gy = 0; //reset change in y
         }
--- a/src/controls.js
+++ b/src/controls.js
@@ -865,17 +865,17 @@ Crafty.c("Fourway", {
 /**@
  * #Twoway
  * @category Input
+ * @trigger NewDirection - When direction changes a NewDirection event is triggered with an object detailing the new direction: {x: x_movement, y: y_movement}. This is consistent with Fourway and Multiway components.
+ * @trigger Moved - When entity has moved on x-axis a Moved event is triggered with an object specifying the old position {x: old_x, y: old_y}
+ * 
  * Move an entity left or right using the arrow keys or `D` and `A` and jump using up arrow or `W`.
- *
- * When direction changes a NewDirection event is triggered with an object detailing the new direction: {x: x_movement, y: y_movement}. This is consistent with Fourway and Multiway components.
- * When entity has moved on x-axis a Moved event is triggered with an object specifying the old position {x: old_x, y: old_y}
  */
 Crafty.c("Twoway", {
     _speed: 3,
     _up: false,
 
     init: function () {
-        this.requires("Fourway, Keyboard");
+        this.requires("Fourway, Keyboard, Gravity");
     },
 
     /**@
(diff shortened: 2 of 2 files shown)


kevinsimper:
This looks fine :+1: Right now you do not get any events when your entity is falling because of gravity. But Crafty gets notified on the updates because else it would not be drawed, so I think it maybe should be another place it should be triggered.

Could you squash the commits to one commit?

@starwed Is this the right place to trigger a moved event?

starwed:
it's a little tricky -- one of the reasons you'd want a "Moved" event is for collision detection.  But gravity does its own collision detection.    

In the long run, the whole thing should be restructured.  This is probably an ok change for now; it won't mess with anything if you don't hook in to the "Moved" event and then cancel the motion.  If you _do_ hook in, you just need to be aware that you'll have to handle the whole business of landing yourself.

Note that the "2D" component uses a different event "Move" to handle internal motions, and "Change" to notify the drawing code.

heniotierra:
I did that because I'm using the "Moved" event to manage the various moving(running, falling, jumping) animations. Of course that could be done in another way, but in this case of mine I think that was the easiest one.

Right now I'm overriding Crafty's Twoway and Gravity with these modifications, but I thought that might be useful for other folks, that's why I decided to do a PR.

@kevinsimper About squashing these to one commit, well, I tried, but I'm not so good with git yet, and just reverting those other commits (about sound and the prior ones) have already given me some trouble. But I'll try. I guess I could also add those triggers to the comments, yes?

starwed:
That use case makes sense, so when you fix up the commits we can merge.  (Assuming it passes tests/linting.)  :)

The simplest way to squash the commits is with an interactive rebase: `git rebase -i develop`.  (Assuming you create this branch from the develop branch.)  

That'll pull up (in your text editor) a list of all the commits since you branched off develop, something like this:

```
pick f7f3f6d Added "Foo" component
pick 310154e Fixed bug with "Foo"
pick a5f4a0d Added documentation
```

You want to change all the "pick"s except the first to "squash" (or just "s"):

```
pick f7f3f6d Added "Foo" component
s 310154e Fixed bug with "Foo"
s a5f4a0d Added documentation
```

save the file, and let git do its thing.

I started a [wiki page](https://github.com/craftyjs/Crafty/wiki/Rewriting-History) about this process, since it comes up a lot.

heniotierra:
There you are, dudes! I hope this is fine now. Please, tell me if something still needs to be changed. I added a comment for trigger in Gravity's description. 
Do I have to do a new PR? I can see the 2 new commits right above. Suppose that means they were added automatically.

@starwed, that command didn't work, maybe because I'm dumb and had made these changes directly in the develop branch, and thus rebase didn't do anything.

starwed:
> that command didn't work, maybe because I'm dumb and had made these changes directly in the develop branch, and thus rebase didn't do anything.

Ah, in that case you can specify upstream/develop, instead of just develop.  (Well, if you've set the upstream repo!)  Or specify by commit hash, or by HEAD~N, where N is the  number of commits back.

There are two commits because you merged into a new branch, I'm guessing.

starwed:
There was still a jshint issue, so I fixed and merged it locally and pushed to the main repo.  Thanks for the PR!

heniotierra:
Nice, thank you! Your previous comment made me a little bit more confused, and I had difficulty figuring out what exactly "upstream" meant in this context, as well as how to set it. Google didn't help so much. :/

starwed:
Ah, github actually explains how to set this, in their instructions on [how to fork a repo](https://help.github.com/articles/fork-a-repo).  Since you already have a repo setup, skip to their section on configuring remotes.  (step 3)

The idea is that there are two remote repos you care about:
-  `origin` is your github fork, which you _push_ to so that you can then share your code 
- `upstream` is this repo, where you'd _pull_ or _fetch_ the most recent versions of the develop branch.

The nice thing is that you can refer to specific commits in the upstream version: `upstream/develop` would refer to the most recent commit in the `develop` branch of your `upstream` repo.

I can push directly to upstream if necessary (that's what I did with your patch!) but it's generally preferable for me to push to my own repo and open a pull request.

heniotierra:
:+1: Got it.

PR #632, opened 2013-11-29
Bump version to 0.6.0

starwed:
Is this everything necessary to bump the version number?

_e:_ Originally I had just `0.6`, but since `npm` wants proper semantic versioning, it has to be `0.6.0`.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,7 +1,7 @@
 {
   "name": "crafty",
   "main": "crafty.js",
-  "version": "0.5.4",
+  "version": "0.6",
   "repository": {
     "type": "git",
     "url": "git://github.com/craftyjs/Crafty.git"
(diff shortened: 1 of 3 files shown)


kevinsimper:
You would also have to change the version in the bower.json file.

I also think it would be a good idea to put the version in the readme file.

starwed:
Ok, fixed the bower file.  (Still have no idea what that's really for!)

The README can be fixed separately, since most user's will see the develop branch version of that in any case!

kevinsimper:
Saw this version reference
https://github.com/craftyjs/Crafty/blob/b0fdcefaf349173f2907f503c5e3d0b152cf6c83/src/core.js#L864

starwed:
That one probably doesn't need to be updated, since it's just an example.

PR #634, opened 2013-11-30
Delete trailing spaces.

pawurb:
How about cleaning all the trailing spaces in the project ?

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -95,7 +95,7 @@ module.exports = function (grunt) {
                 'tests/text.html',
                 'tests/dom.html'
             ]
-        }, 
+        },
 
         jsvalidate: {
             files: "crafty.js"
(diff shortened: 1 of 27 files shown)


starwed:
Hmm, this sort of PR tends to not play well with others; changing so many files at once causes merge conflicts all over the place.

Is there a standard git technique for handling formatting changes like this?  Something that minimizes the pain?

starwed:
[This SO post](http://stackoverflow.com/questions/9776527/git-merging-without-whitespace-conflicts) describes the exact use case we'd want: how to deal with a patch that strips trailing spaces.

pawurb:
The way I understand it is that after merging my PR you will have to use '-Xignore-space-change' flag to avoid conflicts when merging stuff in the future. It is up to you if this if an acceptable price for getting rid of the trailing spaces.

starwed:
Reminder: If we merge this, we should also tell jshint to warn about [trailing spaces](http://www.jshint.com/docs/options/#trailing) in the future.

potomak:
Related: we could also introduce JSBeautifier (maybe in the grunt `check` task).
It could be used to get a consistent style between all the code base.

Example: http://flippinawesome.org/2013/12/02/eliminating-code-smell-with-grunt/

starwed:
I merged this against the current develop branch, fixed a couple of new white space issues that had come up, and opened a new PR #673

PR #635, opened 2013-12-03
Removing a component should actually remove the class from DOM

dkapadia:
When using the DOM component, I found that removing a previously added component did not correctly remove the component from the DOM element's className.

The reason updateClass doesn't work as expected is that the RemoveComponent event is triggered before the component is actually removed from the entity. A different solution to the same problem would be to move that trigger further down in core.js, but that seemed like the sort of thing that could break other folks.

### DIFF
--- a/src/DOM.js
+++ b/src/DOM.js
@@ -58,7 +58,20 @@ Crafty.c("DOM", {
             this._element.className = str;
         }
 
-        this.bind("NewComponent", updateClass).bind("RemoveComponent", updateClass);
+        function removeClass(removedComponent) {
+            var i = 0,
+                c = this.__c,
+                str = "";
+            for (i in c) {
+              if(i != removedComponent) {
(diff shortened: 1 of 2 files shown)


starwed:
Doesn't this override any class names that aren't in `.__c`?  (Someone might manually add others.)  Instead of a loop to construct the string, we might want to use a regex on what already exists.   A [post on stackexchange](http://stackoverflow.com/questions/9959781/remove-classname-from-element-with-javascript/9959811#9959811) suggests an appropriate pattern.

Heh, there'd be a [much easier way](https://developer.mozilla.org/en-US/docs/Web/API/Element.classList) to do this if we didn't care about IE9.  :(

_e:_ Oh, the existing function _also_ clobbers manually added classes.  Nevermind!

starwed:
Merging this as a strict improvement over what we have now.

Thanks for the patch!

PR #638, opened 2013-12-06
Force entities to be rendered at integer coordinates

boe007:
Due to the limitation in canvas that it can only have whole pixels to draw on I inserted a bitwise Math.round (faster on older browsers than the normal Math.round). This removes white lines due to half pixels.

### DIFF
--- a/src/canvas.js
+++ b/src/canvas.js
@@ -88,8 +88,8 @@ Crafty.c("Canvas", {
         }
 
         var pos = this.drawVars.pos;
-        pos._x = (this._x + (x || 0));
-        pos._y = (this._y + (y || 0));
+        pos._x = ~~((this._x + (x || 0))+0.5);
+        pos._y = ~~((this._y + (y || 0))+0.5);
         pos._w = (w || this._w);
         pos._h = (h || this._h);
 


src/canvas.js
```diff
@@ -88,8 +88,8 @@ Crafty.c("Canvas", {
         }
 
         var pos = this.drawVars.pos;
-        pos._x = (this._x + (x || 0));
-        pos._y = (this._y + (y || 0));
+        pos._x = ~~((this._x + (x || 0))+0.5);
+        pos._y = ~~((this._y + (y || 0))+0.5);
```
boe007, line 92: Bitwise Math.round

(remove + 0.5 for a bitwise Math.floor)

starwed:
I could see this being an option, but don't want to make it mandatory.  AFAICT the canvas doesn't just ignore half-pixels; it'll use them to alias differently, etc.  Since that's the standard behavior of the canvas, it should also be our default behavior.

Also, rounding only the rendered position like this would produce incorrect behavior with the current code.  Crafty tracks the bounding rectangles of an entity, and uses that information to know which areas of the canvas to draw on.  Rounding the position up might push the rendered image just a little bit out of bounds, which could lead to a single row of pixels being clipped.

_Example:_ an entity with a `w=3.3` will actually take 4 pixels to draw.  If, say `x = 3.6` then the entity would be assumed to lie within `[3, 7]`.  After rounding up the x, it would actually need pixels `[4, 8]` to render correctly.

If that seems contrived, it's exactly the sort of error people have reported before!  Tweening width while moving an object could naturally lead to this setup.  :)

Always rounding down would be safe.  Or altering some of the rendering code along with this?

kevinsimper:
Now that we added a option called pixelart, would it be an idea now to do the rounding when pixelart is true, or should we just say that in the documentation, that you should remember to round the numbers to whole integers?

starwed:
> Now that we added a option called pixelart, would it be an idea now to do the rounding when pixelart is true, or should we just say that in the documentation, that you should remember to round the numbers to whole integers?

In some pretty limited testing I did when we landed pixelart,  I'm not sure this is necessary.  Moving pixel art around by fractional pixels didn't seem to cause any visual problems; perhaps the same filters that make sure it scales correctly also handles that case?

I'll want to double check in a couple of browsers, though.

kevinsimper:
I just test in Chrome, Safari and Firefox and with the ìmage-rendering` settings, they do the round up/down themselves. Just as you would imagine, 0.5 >= 0 || 0.5 < x is how the do they rounding.

http://jsfiddle.net/VAXrL/1045/

I am going to close this issue then.

Issue #575, opened 2013-10-25
What is missing before publishing Crafty.js 0.6.0?


kevinsimper:
It has been 5 months since Crafty.js has been released, and a lot of the errors we get from the users has already been solved in the development branch. I think the users will really be glad for the changes, especially the ones @starwed has made :+1: 

Also from the outside it seems to people that Crafty.js is no longer under development.
http://html5gameengine.com/details/3/crafty

What do we miss before we can publish version 0.6.0?
- [x] Land desired features 
- [x] Compile [release notes](https://github.com/craftyjs/Crafty/wiki/Release-note-0.6.0)
- [x] Ensure the new documentation is ready to go live (@kevinsimper has access to the site, I think?)
- [x] Run the unit tests in IE9 and IE10.  Checked: Firefox, Safari, Chrome, PhantomJS. 
- [x] Get [public feedback](https://groups.google.com/forum/?fromgroups#!topic/craftyjs/Eq4TClc6WCQ)
- [x] Build Crafty.js and put the compiled files in /dist folder for Bower support

potomak:
I suggest creating a milestone associated with the release of version 0.6.0 collecting all issues that must be closed before tagging the new version.

kevinsimper:
I have created a milestone now, which marks the completeness of version 0.6.0.

@potomak Any issues you think that should be in that milestone?

potomak:
Sorry but I can't help @kevinsimper, I'm involved in this community since a short time.

starwed:
The main thing I want to land before 0.6 is a large change to `animation.js`.  

Currently all animations and tweens run in a kind of tick based way.  I have an almost finished patch that switches this to a time based method, that will allow for more flexible game loops.

It also changes the "SpriteAnimation" interface pretty substantially, so I really want to land it before the next official release.  (We already changed the interface once in the alphas.)  All the functionality is there now, I was just writing tests and docs.

kevinsimper:
That sounds great, and a good idea to have in the 0.6 release.

kevzettler:
@starwed do you feel like pushing your branch or no? Asking incase you think it would be valuable to get some more eyes on it and maybe help contribute.

starwed:
I opened #583 for the branch I've been working on.

kevinsimper:
#583 has been merged.

We should tag it now for release 0.6 :)

starwed:
Yeah, I found a silly bug that was missed by the unit tests and submitted a fix for.

[...]

PR #640, opened 2013-12-06
Bower Support

kevinsimper:
#575 One of the task for 0.6.0 release.


starwed:
Out of curiosity, what does adding the slashes to `crafty.js` in gitignore do?

kevinsimper:
The slash will make git ignore the crafty.js version which is in the root of the git repo, but not the ones that is nested, like in the /dist folder.

Issue #641, opened 2013-12-06
css3dtransform support not detected correctly in IE11


kevinsimper:
DOM: Internet Explorer 9 and 11 error, 10's fine:
![image](https://f.cloud.github.com/assets/1126497/1693860/df71f714-5e96-11e3-9f58-482d0cf24788.png)

That is the only error i encounted, when doing all the test in internet explorer 9 - 10 - 11.

The test in Crafty.support.css3dtransform is properly wrong in IE11, because it shoud support it.

potomak:
I think it's my fault. The test should actually distinguish between browser which support CSS 3d transforms and those who don't.

starwed:
I think this has been fixed, but I don't have access to a windows machine with IE11, so feel free to reopen if I'm wrong.  :)

PR #642, opened 2013-12-06
Fix failing DOM test on IE

potomak:
Note: the `avoidCss3dTransforms` property doesn't have any effect on
browsers that don't css3dtransform.

See also #641

### DIFF
--- a/tests/dom.html
+++ b/tests/dom.html
@@ -28,16 +28,22 @@ $(document).ready(function() {
   Crafty.init(500,500);
   
   module("DOM");
-  
+
   test("avoidCss3dTransforms", function() {
     var useCss3dTransforms = Crafty.e("2D, DOM")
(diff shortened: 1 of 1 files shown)


kevinsimper:
Nice :+1: 

But do you not think the test for `Crafty.support.css3dtransform` is not working correctly. Should we use something like this:
http://stackoverflow.com/questions/5661671/detecting-transform-translate3d-support

potomak:
I don't know if test for 3D transform is right, this PR fixes a different bug about tests.

If you think that `css3dtransform` detection is wrong I think you could open a new issue.

starwed:
Yeah, better to treat them as separate issues, even though they had the same symptoms.  :)

(Although, we only noticed the IE11 issue because of the "broken" test.)

PR #644, opened 2013-12-06
Bump version to 0.6.1

starwed:
Bump the version to 0.6.1 for bower.

Unfortunate that I don't see a way to test whether this will work without actually doing a release!


starwed:
Ok, I successfully used `bower install crafty`.  Hope it's useful to someone. :)

kevinsimper:
This is very useful! :+1:

PR #646, opened 2013-12-06
Default actions is not ignored on input and textareas.

### LINKED ISSUE #645: Virtual keyboard not shown when input in tapped on (mobile devices)
There is the same bug on touch as there was fix on desktop.

Mention on the forum here:
https://groups.google.com/forum/#!topic/craftyjs/w5B9Vas5RuM

Basicly i think the same fix that was made here:
https://github.com/craftyjs/Crafty/blob/develop/src/controls.js#L310

Should be applied here:

https://github.com/craftyjs/Crafty/blob/develop/src/controls.js#L231

kevinsimper:
Fixes #645

### DIFF
--- a/src/controls.js
+++ b/src/controls.js
@@ -228,8 +228,14 @@ Crafty.extend({
             first.target.dispatchEvent(simulatedEvent);
         }
 
-        if (e.preventDefault) e.preventDefault();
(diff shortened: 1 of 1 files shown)


starwed:
Looks good to me.

Issue #616, opened 2013-11-14
Says time instead of frames - viewport documentation


kevinsimper:
It says frames in the description, but says time in the arguments list.

http://ludonomicon.com/crafty/api/Crafty-viewport.html

starwed:
Well, it is a time -- just measured in frames.   But yeah, probably misleading.

fwiw,I plan on rewriting these to be time based after the 0.6 release.

starwed:
As of #648, it _is_ a time.  :)

Issue #339, opened 2012-08-30
Zooming slowly pulls Southeast.


thehack:
If I do:  

``` javascript
Crafty.viewport.zoom(scale, center.x, center.y, numberUnder100);
```

It works as expected, but if I do:

``` javascript
Crafty.viewport.zoom(scale, center.x, center.y,numberOver100);
```

It moves the camera to the Southeast of the map. See http://jsfiddle.net/czeeA/14/

The end result is you can't do a slow zoom over a couple seconds to create a cutscene from still images.

mpetrovic:
I'll take a look at this tonight.

starwed:
To be clear, if you use DOM entities it doesn't zoom at all!  Just the panning.

Amusingly, if you use Canvas it does zoom, but also pulls to the NW of the target point.  ([fiddle](http://jsfiddle.net/czeeA/22/))

starwed:
Update: I'm working on a rewrite of the viewport animations that fixes this.

starwed:
Should now be fixed by #648!

Issue #285, opened 2012-05-30
viewport.mouselook is not taking zoom-level into account.


nocksock:
I already posted it into the [google-groups](https://groups.google.com/forum/?hl=en&fromgroups#!topic/craftyjs/_-pVfOMnkZU) but to avoid to let this slip into the back of heads I created an Issue for this. 

Here I recreated the problem: http://jsfiddle.net/noxoc/dQyy2/ It should basically not be possible to move the image horizontally.

sorenbs:
Do you know what the status is on this?

starwed:
Should now be fixed by #648!

PR #648, opened 2013-12-07
Rework the viewport animation functions

### LINKED ISSUE #163: Animation Over Time Fns Should Default to 1 Frame
For example, the following two method calls should have the same behavior:

Crafty.viewport.centerOn(entity)
Crafty.viewport.centerOn(entity, 1)

In viewport.pan(), the line change would be

<pre>
- if (time == 0) time = 1;
+ if (time &#060; 1) time = 1;
</pre>

starwed:
This isn't quite ready to land yet, but thoguht I'd open the PR so folk could see the changes.  This alters several aspects of the viewport:
- Changes pan and zoom to work with Crafty.easing. (Time based instead of tick based.)
- Makes sure the zoom animation is nice and smooth.  (Surprisingly tricky)
- Fixes a bug in how scale is treated by the DOM/Canvas renderers
- Adds a new "StopCamera" event for stopping any extant camera animation (and uses it when a new animation is requested.)
- Fires an "CameraAnimationDone" event when a camera animation completes

Along the way this fixes #163, #285, #339, and #616

### DIFF
--- a/src/drawing.js
+++ b/src/drawing.js
@@ -520,7 +520,7 @@ Crafty.DrawManager = (function () {
 
             if (dirtyViewport) {
                 var view = Crafty.viewport;
-                ctx.setTransform(view._scale, 0, 0, view._scale, view.x, view.y);
+                ctx.setTransform(view._scale, 0, 0, view._scale, view._x*view._scale, view._y*view._scale);
 
             }
             //if the amount of changed objects is over 60% of the total objects
@@ -629,8 +629,8 @@ Crafty.DrawManager = (function () {
                     view = Crafty.viewport;
 
                 style.transform = style[Crafty.support.prefix + "Transform"] = "scale(" + view._scale + ", " + view._scale + ")";
-                style.left = view.x + "px";
-                style.top = view.y + "px";
+                style.left = view.x * view._scale + "px";
+                style.top = view.y * view._scale + "px";
                 style.zIndex = 10;
             }
 
--- a/src/viewport.js
+++ b/src/viewport.js
@@ -8,6 +8,8 @@ Crafty.extend({
      * @trigger ViewportScroll - when the viewport's x or y coordinates change
      * @trigger ViewportScale - when the viewport's scale changes
      * @trigger InvalidateViewport - when the viewport changes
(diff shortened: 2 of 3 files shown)


starwed:
Ok, this is a candidate to land now.  I had to fix some issues with the tests.

Forgot to mention that this changes the way you use `pan()`, by specifying a direction along both axes at once, rather than specifying an axis individually.

There was a feature that forced the viewport to stick to integer coordinates, to prevent weird aliasing issues.  But this was actually really ugly when zooming slowly; it caused the position to stutter a lot.  So I disable this, but we should probably allow it as optional behavior.

src/drawing.js
```diff
@@ -520,7 +520,7 @@ Crafty.DrawManager = (function () {
 
             if (dirtyViewport) {
                 var view = Crafty.viewport;
-                ctx.setTransform(view._scale, 0, 0, view._scale, view.x, view.y);
+                ctx.setTransform(view._scale, 0, 0, view._scale, view._x*view._scale, view._y*view._scale);
```
kevinsimper, line 523: This solves the bug where the zoom is going southeast?
http://www.w3schools.com/tags/canvas_settransform.asp

kevinsimper:
:+1:  Nice

But this pull request could indeed need different commit because it can be a little hard which changes are solving what problems :-)

Like the code together with the test that would be awesome, some of this do not make sense unless your are really dig into canvas and use time reading the code. 

There is a reason not everybody is doing this (because it is friggin' hard, and canvas is still "new" so every changes for learning something new is appreciated! :+1:

src/drawing.js
starwed, line 523: Probably, but since I rewrote the zoom code at the same time I'm not 100% sure it was the cause of the original bug.

starwed:
Heh, I just fixed things until the zoom code worked, which is why it's all one lump.

The change to the drawing code could definitely be a separate commit.

starwed:
Ok, split the drawing commit out, and added some tests for the CameraAnimationDone event.

starwed:
Going to go ahead and merge this.

PR #650, opened 2013-12-09
Fix naming convention on the Crafty Core file

kevinsimper:
The file, Crafty core, gets the wrong name when the API is built.

The wrong file name it gets is: "Crafty Core.html", but it should be "Crafty-Core.html".

The craftyjs website is fixed.
https://github.com/craftyjs/craftyjs.github.com/pull/29

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -53,7 +53,7 @@ var Crafty = function (selector) {
 initState();
 
 /**@
- * #Crafty Core
+ * #Crafty.Core
  * @category Core
  * @trigger NewEntityName - After setting new name for entity - String - entity name
  * @trigger NewComponent - when a new component is added to the entity - String - Component


starwed:
Hmm, semantically it's confusing to call it "Crafty.Core", since that implies it's a property of Core.

We might want to fix this in the doc generator instead?

starwed:
Ok, I'm kind of confused -- when I run the api-gen script, it produces the correct link, with a space and not a dash.  How did the version on github end up with a dash?

kevinsimper:
I do not know, the old 0.5.4 documentation had dash in filename (avoiding ugly %20), but space in name.
http://craftyjs.com/release/0.5.4/api/Crafty-Core.html

I do also think that it has to be that way, I just tested and though somebody maybe had removed a dot before I did the browserify change.

Because when i do `grunt api` i get a file called `Crafty Core.html`, but the navigation which is created has a dash.

starwed:
Looks like it was a problem with an out-dated index.html to me.  (I submitted a PR in the site repo.)

We could adjust the doc generator to avoid spaces in the URLs, but that's a separate issue, so lets close this PR.

kevinsimper:
Yes, that confused me about the space in the filename which makes a less nice url.

The old php documentation creator did not do that :) But nice that the docs is fixed now.

PR #651, opened 2013-12-11
Add get() method to crafty core

### LINKED ISSUE #349: Crafty selector returns IDs not Entities
If I run this command:

```
var entities = Crafty("ComponentName");
```

I would expect an array of entities, however, I receive an array of entity IDs. To retrieve the actual entity  I must call `Crafty(entities[i])` for each ID. 

The documentation states that this function returns entities, not IDs. The documentation should be changed to reflect this, or (as I would suggest) the function should be changed to return the actual entities.

starwed:
This adds a get method that will convert selections to arrays of entities.  It can also be used to fetch a particular specified entity by index.  It will work as expected an a single entity.  This is modeled directly after jquery's get method: http://api.jquery.com/jQuery.get/

Documentation about the new feature is added (it is now referenced by the Crafty Core docs, as well as  .toArray).  Includes a few tests added to core.html.

This fixes #349

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -27,6 +27,14 @@ var version = require('./version');
  *   Crafty(1)
  * ~~~
  * Passing an integer will select the entity with that `ID`.
+ *
+ * To work directly with an array of entities, use the `get()` method on a selection.
+ * To call a function in the context of each entity, use the `.each()` method.
+ *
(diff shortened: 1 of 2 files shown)


starwed:
Since this doesn't mess with any existing code, I'm going to go ahead and land this.

PR #652, opened 2013-12-11
Add support for line-height and variant to the Text component

### LINKED ISSUE #494: CSS line-height property cannot be set for entities with Text component
As of v0.5.4, setting the CSS `line-height` (or `lineHeight`) property has no effect for DOM entities with the Text component. The value of the line-height property is clobbered when the style's font is set in the Draw callback of the Text component (src/text.js:47).

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/font), line-height is a component of the font property. When the font property is set, but no line-height is specified, the line-height will default to `normal`. As such, it seems as though the only correct fix for the current system is to allow line-height to be specified using the `.textFont` method of the Text component, and have it incorporated in the font when it is set.

I'll prepare a pull request as soon as I can. If Im missing something here, please let me know.

starwed:
These work using DOM, but not canvas.  The textFont method now covers all the possible parts of the [font declaration.](https://developer.mozilla.org/en-US/docs/Web/CSS/font).

Fixes #494.

### DIFF
--- a/src/text.js
+++ b/src/text.js
@@ -28,6 +28,8 @@ Crafty.c("Text", {
     _text: "",
     defaultSize: "10px",
     defaultFamily: "sans-serif",
+    defaultVariant: "normal",
+    defaultLineHeight: "normal",
     ready: true,
 
     init: function () {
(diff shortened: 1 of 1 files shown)


starwed:
No comments, seems to work in tests, so merging.

PR #657, opened 2013-12-18
Kill ie8 specific code in Crafty

starwed:
This removes several chunks of code that exist only to support IE8.

grep was helpful in finding the most explicit stuff, but I'm sure there's more.

I think there was also some stuff added to the tests to work around IE8?

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -34,7 +34,7 @@ var Crafty = function (selector) {
 },
     // Internal variables
     GUID, frame, components, entities, handlers, onloads,
-    noSetter, slice, rlist, rspace, milliSecPerFrame;
+    slice, rlist, rspace, milliSecPerFrame;
 
 
     initState = function () {
@@ -747,8 +747,6 @@ Crafty.fn = Crafty.prototype = {
      * Will watch a property waiting for modification and will then invoke the
      * given callback when attempting to modify.
      *
-     * *Note: Support in IE<9 is slightly different. The method will be executed
-     * after the property has been set*
      */
     setter: function (prop, callback) {
         if (Crafty.support.setter) {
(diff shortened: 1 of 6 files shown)


starwed:
Oh, I think I might have accidentally included IE9 specific stuff -- I'm now a bit confused about the `nosetter` code.  I'll have a look at that later, but if someone actually knows what's going on there, would be nice to hear from you.  :)

sbyrnes321:
Looking good, thanks!!

Yeah, in tests/core.html, you would delete the lines with "no_getter_setters". Very obvious, you can't miss it!

src/core.js
```diff
@@ -1573,24 +1565,6 @@ function clone(obj) {
     return temp;
 }
 
-Crafty.bind("Load", function () {
-    if (!Crafty.support.setter && Crafty.support.defineProperty) {
```
starwed, line 1577: This is the line that's confusing me -- it implies that the code below should be applied when the `defineProperty` approach is used.  (That's what IE9 supports.)

But actually, the code that folows is just the fallback IE8 approach.  Maybe the conditional is just a typo?

starwed:
Ok, pulling the trigger on this!

potomak:
There's also an interesting fix for IE8 at https://github.com/craftyjs/Crafty/blob/develop/src/import.js#L23 where `Crafty['import']` is used instead of `Crafty.import` as `import` seems to be a reserved work in IE8.

potomak:
Also the `storage.js` script contains IE8 specific _evil_ code, see also the comment at https://github.com/craftyjs/Crafty/blob/develop/src/storage.js#L4-L5.

In my opinion the whole Storage component should be excluded from Crafty core and added to any project only if needed (it needlessly implements HTTP communication through at low level using `XmlHttpRequest`).

PR #658, opened 2013-12-18
Allow visibility MBR to be manually offset

starwed:
This patch adds internal properties to "2D" that indicate how far beyond an object's default dimensions the visible MBR should be.

These can be set with the public method `offsetBoundary`.

The point of this is that custom rendering might want to create effects that extend beyond the standard dimensions of the entity -- a glow around a sprite, for instance.  It shouldn't really be used anywhere _except_ custom rendering stuff.

### DIFF
--- a/src/2D.js
+++ b/src/2D.js
@@ -324,6 +324,13 @@ Crafty.c("2D", {
             x: 0,
             y: 0
         };
+
+        // offsets for the basic bounding box
+        this._bx1 = 0;
+        this._bx2 = 0;
+        this._by1 = 0;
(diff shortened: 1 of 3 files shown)


starwed:
[Here](http://jsfiddle.net/c3597/15/) is a jsfiddle which shows why this might be necessary.  A custom component is defined, and moved both with and without an offset to the MBR.

starwed:
No comments, so merging.

kevinsimper:
Good job and great jsfiddle example!

PR #660, opened 2013-12-19
extracted code from extensions.js

mucaho:
moved keys & mouseButtons to keycodes.js
moved zeroFill to 2D.js (only 2D.js uses zeroFill)

as suggested, small steps towards #603


kevzettler:
:+1:

starwed:
Make sense, thanks!

PR #661, opened 2013-12-19
added simple executable example

mucaho:
added the example from the front page to be quickly accessible after building crafty.js

just an idea :)

### DIFF
--- a/example.html
+++ b/example.html
@@ -1,0 +1,56 @@
+<html>
+  <head></head>
+  <body>
+    <div id="game"></div>
+    <script type="text/javascript" src="crafty.js"></script>
+    <script>
+        Crafty.init(600, 300, document.getElementById('game'));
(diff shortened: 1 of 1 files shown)


starwed:
We actually did have some demos included in Crafty previously; louis removed them without commenting as to why.

I always found them useful to check if any obvious functionality was broken, so we might want to include something similar again?

mucaho:
> I always found them useful to check if any obvious functionality was broken, so we might want to include something similar again?

Agreed. 
Perhaps it was cumbersome to update the examples according to the api changes every time?

mucaho:
There is a [demo repository](https://github.com/craftyjs/demos). Could we put them in there (e.g. labeled as developer demos)?

starwed:
@mucaho  Yeah, if you like go ahead and put anything you like there.  :)

mucaho:
Ok thanks, did it

Issue #419, opened 2013-02-15
Twoway jump while falling (fix included)


melanke:
in current implementation of Twoway the player can jump in middle air while falling, just like in Transformice flash game.

I think most developers dont want this so I've made a little hack and developed this quickfix.

```
Crafty.c("Twowayfix", {
            _speed: 3,
            _up: false,

            init: function (){
                this.requires("Fourway, Keyboard")
            },

            twoway: function (speed, jump){

                this.multiway(speed, {
                    RIGHT_ARROW: 0,
                    LEFT_ARROW: 180,
                    D: 0,
                    A: 180,
                    Q: 180
                });

                if(speed){
                    this._speed = speed
                }

                jump = jump || this._speed * 2;
                this.bind("EnterFrame", function(){
                    if(this.disableControls){
                        return
                    }

                    if(this._up){
                        this.y -= jump;
                        this._falling = true
                    }
                }).bind("KeyDown", function(){

                    if(!this._falling && (this.isDown("UP_ARROW") || this.isDown("W") || this.isDown("Z"))){
                        this._up = true
                    }
                });

                return this

            }

        });
```

the only change was on the last if of the code above:
**!this._falling**

I hope I am helping

thanks for this great Game Engine, it is the best! :D

PR #662, opened 2013-12-20
fix jump while falling

mucaho:
PR for #419


starwed:
Merging, since it's an improvement over what we have right now.

Issue #656, opened 2013-12-16
SpriteAnimation not working in Safari/Chrome on DOM based entities


matthewsimo:
Hi, I just picked up Crafty this weekend and was playing around with it and I came across a bug. I couldn't get SpriteAnimation to work but after lot's of facepalms I found out why... It stayed on the first frame and didn't animate, but also returned back isPlaying() as true.

It seems to only appear when attempting to animate on entities that are DOM based in webkit/blink. I'm on Mac, so I don't know if it's also a windows issue, any one follow up on that?

My simple test used:

```
  c = Crafty.e("2D, DOM, SpriteAnimation, CursorSprite")
                .reel('CursorBlink', 1000, [[0,1], [2,1]])
                .animate('CursorBlink', -1);
```

And I spent way to long debugging it, haha... changing to the following resolved:

```
  c = Crafty.e("2D, Canvas, SpriteAnimation, CursorSprite")
                .reel('CursorBlink', 1000, [[0,1], [2,1]])
                .animate('CursorBlink', -1);
```

demo gif...
![crafty-bug](https://f.cloud.github.com/assets/207644/1751541/aa7daeca-65f0-11e3-856b-269674a05f5d.gif)

Hope that helps someone else from pulling out their hair... 

Thanks!

starwed:
Ah, it seems to be some issue with how we set the CSS.  Currently the code sets a bunch of properties in one line:

``` javascript
 style.background = style.backgroundColor + " url('" + this.__image + "') no-repeat -" + co.x * hscale + "px -" + co.y * vscale + "px";
```

Using the inspector in Safari, I figured out that the background position wasn't being set when animated.  On a hunch I tried setting the background position on a separate line:

``` javascript
style.background = style.backgroundColor + " url('" + this.__image + "') no-repeat";
style.backgroundPosition = "-" + co.x * hscale + "px -" + co.y * vscale + "px";
```

It then works fine.

Since it worked in Firefox, I'm not sure if this is a bug in what we're doing, or a bug in webkit.

Also, I'm not sure why this bug is only triggered when using SpriteAnimation.  If you manually add a sprite component, the position is set correctly, even though it uses exactly the same code to do so.

I'll investigate a little further, but if I don't turn up anything I'll land the fix above.  Just wish I knew why it worked.  : (

matthewsimo:
Yea, that's very odd indeed, thanks for the update!

starwed:
Should be fixed by #664 

[...]

PR #664, opened 2013-12-24
Fix problem with DOM sprite animation in webkit browsers

starwed:
The backgroundPosition CSS attribute isn't always getting set correctly in Safari.  This seems to be some sort of hard to trigger webkit-only bug that occurs when we do sprite animation, but not when setting regular sprite positions.

I was unable to produce a simple test case without using Crafty, so I'm not sure what the exact problem is.  This "fixes" things by setting the background position separately from the rest of the background properties.

This fixes issue #656.

### DIFF
--- a/src/sprite.js
+++ b/src/sprite.js
@@ -56,7 +56,8 @@ Crafty.c("Sprite", {
                     hscale = this._w / co.w,
                     style = this._element.style;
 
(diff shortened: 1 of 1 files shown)


starwed:
This should have basically no effect in other cases, so merging.

Issue #416, opened 2013-02-09
Tilesets with a margin not supported?


cawoodm:
It seems `Crafty.sprite()` doesn't support tilesets with a margin around the outside
![tmw_desert_spacing](https://f.cloud.github.com/assets/498834/141935/4135cfb0-7299-11e2-981a-9176b10515b9.png)

I had to add the paddingX to the x value on line 4475:

``` javascript
            x = temp[0] * (tile + paddingX)+paddingX;
            y = temp[1] * (tileh + paddingY)+paddingY;
```

darrentorpey:
@cawoodm, FYI I'm taking this on for addition to the engine.

Out of curiosity, where did you get that nice tileset? Is it free to be used elsewhere outside of your game?

cawoodm:
Super. That map came with the free "Tiled Map Editor" SW you can download at http://mapeditor.org It's cool because you can export your creation as JSON and then use the TiledMapBuilder (https://github.com/Kibo/TiledMapBuilder) component to employ the sprites.

mucaho:
@starwed can you close this now, please? as it was actually the primary objective of #665 :)

starwed:
Hah, sure.  :)

There is some magic phrase that will close an issue automatically when a PR lands -- I think if you include the words `Fixes #416` in the commit message it will happen?  

Although maybe you need the correct permissions.  No idea!  :)

kevinsimper:
https://help.github.com/articles/closing-issues-via-commit-messages :+1:

Issue #482, opened 2013-06-02
"Change" event overused and overloaded


starwed:
Every time you use `attr()`, a change event is triggered.  It is handed a map of property names and their new values.

When you assign any of the 2D properties (x, y, w, h, rotation) a change event is also triggered.  In this case, it is handed an object representing the object's _previous_ state.

If you use `attr` to set a 2D property, both will be triggered.

There are two places in Crafty that are bound to this event
- 2D entities need "Change" to know when they need to be redrawn.  The current code doesn't need to be told what the old state was (that's tracked by the drawing routine now).  This was the consumer of the second type of event, which could maybe be replaced by something like "VisibleChange"?
- The viewport.follow function uses "Change" to know when to follow an object.  It likewise doesn't care what object the handler gets.

If anyone tries to consume "Change" events currently they'll be confused by the two types, so we need to clear this up!

starwed:
I propose using an event called `"Invalidate"` to mark when an object needs to be redrawn.  That (somewhat) matches the same concept in Flash and Android environments.  It wouldn't be passed any data -- the object itself will contain all information necessary to draw itself.

`"Change"` could then be used to indicate that some property of the entity had changed.  It could be triggered by anything that functioned like a setter.  I don't think this would need to be used internally in crafty at all -- the one use by `viewport.follow` could just be tied to the `"Moved"` event.  

Thoughts?

grozen:
Sounds like a most blessed fix.

PR #665, opened 2013-12-24
adjust padding offset

mucaho:
PR for #416

This is in current build:

``` javascript
x = temp[0] * (tile + paddingX);
y = temp[1] * (tileh + paddingY);
```

However the author of the issue suggests:

``` javascript
x = temp[0] * (tile + paddingX) + paddingX;
y = temp[1] * (tileh + paddingY) + paddingY;
```

Look at the example tileset in the referenced issue. The tileset has padding before the first tile, so his solution is correct:
let `temp[0] = 0`, then `x = 0*(...) = 0`. Instead `x = 0*(...) + paddingX = paddingX` is required.
left `temp[0] = 1`, then `x = 1*(tileWidth + paddingX)`. Instead  `x = 1*(tileWidth + paddingX) + paddingX` is required (2 paddings required for 2nd x tile).

I have tested it with the example tileset and in the current build the black border is indeed visible. With this fix the black border is not visible. The question is whether **every** spriteset has padding like this?

### DIFF
--- a/src/extensions.js
+++ b/src/extensions.js
@@ -139,13 +139,14 @@ Crafty.extend({
     /**@
      * #Crafty.sprite
      * @category Graphics
-     * @sign public this Crafty.sprite([Number tile, [Number tileh]], String url, Object map[, Number paddingX[, Number paddingY]])
+     * @sign public this Crafty.sprite([Number tile, [Number tileh]], String url, Object map[, Number paddingX[, Number paddingY[, Boolean paddingAroundBorder]]])
      * @param tile - Tile size of the sprite map, defaults to 1
      * @param tileh - Height of the tile; if provided, tile is interpreted as the width
      * @param url - URL of the sprite image
      * @param map - Object where the key is what becomes a new component and the value points to a position on the sprite map
      * @param paddingX - Horizontal space in between tiles. Defaults to 0.
      * @param paddingY - Vertical space in between tiles. Defaults to paddingX.
+     * @param paddingAroundBorder - If padding should be applied around the border of the sprite sheet. If enabled the first tile starts at (paddingX,paddingY) instead of (0,0). Defaults to false.
      * Generates components based on positions in a sprite image to be applied to entities.
      *
      * Accepts a tile size, URL and map for the name of the sprite and its position.
@@ -186,7 +187,7 @@ Crafty.extend({
      *
      * @see Sprite
      */
-    sprite: function (tile, tileh, url, map, paddingX, paddingY) {
+    sprite: function (tile, tileh, url, map, paddingX, paddingY, paddingAroundBorder) {
         var spriteName, temp, x, y, w, h, img;
 
         //if no tile value, default to 1.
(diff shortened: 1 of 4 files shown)


starwed:
I don't think we can assume anything in particular about spritesheets -- I've definitely seen some that have internal padding, but none around the edge of the sheet.

starwed:
I guess you'd also need to make it play nice with sprite animation.

It seems silly that we have to duplicate code like that, so maybe that should be something that's abstracted out by the sprite component.  No need to worry about that in this PR, but it does need to work with animations.

mucaho:
`__coord` is only updated through `.sprite(x, y, [w, h])` now

mucaho:
ah we need to think about performance:
SpriteAnimation calculated the positions upfront; with this PR the positions are calculated each time the frame changes.
I will update the PR to revert to the old way of determining the positions upfront, while still abstracting it

starwed:
This will presumably fix #178, since it normalises the code used by regular sprites and spriteAnimation.

mucaho:
exactly; it should also allow other dynamic modifications to the sprite component (as the positions are relative instead of absolute).
@starwed what do you think? is it worth the performance overhead?

starwed:
> what do you think? is it worth the performance overhead?

I think that would be fine -- it's pretty low to begin with, and switching the sprite isn't something that would happen every frame.

If you're _really_ worried, some of those calculations could be cached, but I just can't see it making a difference in practice.

mucaho:
ok then, I removed a duplicate change event, if your ok with the performance, this PR is ready.

In the future we could trigger a "SpriteCoordChange" event which would recache the SpriteAnimation frames every time the sprite is changed by the user. That should be added together with #482 .

starwed:
> this PR is ready.

Yup, thanks!  Though it does need rebasing against HEAD right now.

mucaho:
I have updated my forked repo - develop branch with the commits from the remote (original) repo - develop branch.
How do I do the rebase against HEAD now?

starwed:
From your feature branch, you use the command `git rebase develop`.  That last argument specifies what you're rebasing against, in this case the `develop` branch.

You'll then have to manually resolve the merge conflicts.  I think in this case your commits touched the same code as my change to "SpriteAnimation", removing a stray reference to `toX`-- if so you can just overwrite that with your new code.

If you have difficulties with the merge I can try to walk you through it.  (That was the most confusing part of learning git when I was getting started.)

mucaho:
Ok thanks, I managed to get it merged somehow. I hope it's ok now.

starwed:
Ok, looks good to me!

Thanks for cleaning up this stuff.

src/extensions.js
```diff
@@ -235,12 +236,13 @@ Crafty.extend({
             this.requires("2D, Sprite");
             this.__trim = [0, 0, 0, 0];
             this.__image = url;
-            this.__coord = [this.__coord[0], this.__coord[1], this.__coord[2], this.__coord[3]];
             this.__tile = tile;
```
starwed, line 239: Oops, this rather odd looking line was vital!  It prevents different entities from sharing the same __coord array.

I never tested this change with more than one sprite of the same type, but as soon as I did all sorts of weird things started happening.  :)

I'll open a PR that addresses this.  (As soon as I have time to write some tests.)

PR #666, opened 2013-12-24
added Crafty.pixelart

### LINKED ISSUE #330: Canvas scale and context.imageSmoothingEnabled
Pixel art is quite popular, especially in web games, since low-res graphics are faster to load. But if we work with graphics so small they need to be scaled. What do you think about including Crafty.canvas.scale(x [, y]) that would call context.scale(x,y)? Also, by default, browsers use some kind of image smoothing in canvas when scaled. What do you think of adding Crafty.canvas.imageSmoothingEnabled(enable), and set enable to context.mozImageSmoothingEnabled (Firefox 3.6+) and context.webkitImageSmoothingEnabled (Chrome 21+). I'm a beginner, but if someone would be kind enough to correct me later I'd be happy to do that.

mucaho:
fixes #330 . However this is not working on FF (see original issue)

**Updated Example**
Example test that should work for all browsers:

``` javascript
    Crafty.init(600, 300, document.getElementById('game'));

    Crafty.background('rgb(0,0,0)');
    Crafty.viewport.scale(6);

    Crafty.canvas.init();
    Crafty.pixelart(true);

    /*
    // Not working on FF
    Crafty.e("2D, Canvas, Image")
        .image("squares.gif");
    */

    /*
    // Not working on Webkit
    Crafty.e("2D, DOM, Image")
        .image("squares.gif");
    */

    //Should work cross-browser
    Crafty.sprite(50, 50, "squares.gif", {squares:[0,0]});
    Crafty.e("2D, Canvas, squares");
```

### DIFF
--- a/src/drawing.js
+++ b/src/drawing.js
@@ -652,3 +652,63 @@ Crafty.DrawManager = (function () {
 
     };
 })();
+
+Crafty.extend({
+    /**@
+     * #Crafty.pixelart
+     * @category Graphics
+     * @sign public void Crafty.pixelart(Boolean enabled)
+     *
+     * Sets the image smoothing for drawing images (for both DOM and Canvas).
+     * Setting this to true disables smoothing for images, which is the preferred
+     * way for drawing pixel art. Defaults to false.
+     *
+     * This feature is experimental and you should be careful with cross-browser compatibility. 
+     * The best way to disable image smoothing is to use the Canvas render method and the Sprite component for drawing your entities.
+     *
+     * This method will have no effect for Canvas image smoothing if the canvas is not initialized yet.
+     *
+     * Note that Firefox_26 currently has a [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=696630) 
+     * which prevents disabling image smoothing for Canvas entities that use the Image component. Use the Sprite
+     * component instead.
+     * Note that Webkit (Chrome & Safari) currently has a bug [link1](http://code.google.com/p/chromium/issues/detail?id=134040) 
+     * [link2](http://code.google.com/p/chromium/issues/detail?id=106662) that prevents disabling image smoothing
+     * for DOM entities.
+     *
+     * @example
+     * This is the preferred way to draw pixel art with the best cross-browser compatibility.
+     * ~~~
+     * Crafty.canvas.init();
+     * Crafty.pixelart(true);
+     * 
+     * Crafty.sprite(imgWidth, imgHeight, "spriteMap.png", {sprite1:[0,0]});
+     * Crafty.e("2D, Canvas, sprite1");
+     * ~~~
+     */
+    pixelart: function(enabled) {
+        var context = Crafty.canvas.context;
+        if (context) {
+            context.imageSmoothingEnabled = !enabled;
+            context.mozImageSmoothingEnabled = !enabled;
+            context.webkitImageSmoothingEnabled = !enabled;
+            context.oImageSmoothingEnabled = !enabled;
+            context.msImageSmoothingEnabled = !enabled;
+        }
+
+        var style = Crafty.stage.inner.style;
+        if (enabled) {
+            style[Crafty.DOM.camelize("image-rendering")] = "optimizeSpeed";   /* legacy */
+            style[Crafty.DOM.camelize("image-rendering")] = "-moz-crisp-edges";    /* Firefox */
+            style[Crafty.DOM.camelize("image-rendering")] = "-o-crisp-edges";  /* Opera */
+            style[Crafty.DOM.camelize("image-rendering")] = "-webkit-optimize-contrast";   /* Webkit (Chrome & Safari) */
+            style[Crafty.DOM.camelize("-ms-interpolation-mode")] = "nearest-neighbor";  /* IE */
+            style[Crafty.DOM.camelize("image-rendering")] = "optimize-contrast";   /* CSS3 proposed */
+            style[Crafty.DOM.camelize("image-rendering")] = "pixelated";   /* CSS4 proposed */
+            style[Crafty.DOM.camelize("image-rendering")] = "crisp-edges"; /* CSS4 proposed */
+        } else {
+            style[Crafty.DOM.camelize("image-rendering")] = "optimizeQuality";   /* legacy */
+            style[Crafty.DOM.camelize("-ms-interpolation-mode")] = "bicubic";   /* IE */
+            style[Crafty.DOM.camelize("image-rendering")] = "auto";   /* CSS3 */
+        }
+    }
+});


starwed:
Image smoothing in Firefox works fine when you're using sprites -- it's just the method Image uses that's broken.  The note should make that distinction clear.

starwed:
This should probably also handle the DOM side of things too.  (Or is that already handled somewhere?)

[This SO answer](http://stackoverflow.com/questions/7615009/disable-interpolation-when-scaling-a-canvas/11751817#11751817) suggests the following CSS for image rendering:

```
image, canvas {
    image-rendering: optimizeSpeed;             // Older versions of FF
    image-rendering: -moz-crisp-edges;          // FF 6.0+
    image-rendering: -webkit-optimize-contrast; // Webkit
                                                //  (Safari now, Chrome soon)
    image-rendering: -o-crisp-edges;            // OS X & Windows Opera (12.02+)
    image-rendering: optimize-contrast;         // Possible future browsers.
    -ms-interpolation-mode: nearest-neighbor;   // IE
}
```

[This MDN page](https://developer.mozilla.org/en-US/docs/Web/CSS/image-rendering?redirectlocale=en-US&redirectslug=CSS%2Fimage-rendering) documents the CSS stuff a bit.  (Though note that the `pixelated` value isn't actually supported by anyone yet.)

mucaho:
I was able to use CSS as you suggested to create a DIV manually.
However I am stuck in adding the code to `DOM.draw()` call.
Why can't I do

``` javascript
style["image-rendering"] = "-moz-crisp-edges";
```

in there? I don't see the property being added to the HTML.

**Change style from javascript:**
You can only add CSS styles to single elements with Javascript, right?
Can't I set a cascading style for  the `Crafty.stage.inner` element  that would apply to every existing and future child?

**Scope of image smoothing:**
If the image smoothing has to be specified at each `DOM.draw()` call, should the image smoothing be applied per entity like `Crafty.e("2D, DOM/Canvas").setImageSmoothing(false)` or should the image smoothing be specified globally as it is now  `Crafty.setImageSmoothing(false)`?
I should be able to change the image smoothing individually between the `context.save()` and `context.restore()` calls in each `Canvas.draw()` call.

starwed:
> Why can't I do `style["image-rendering"]`

From javascript, `css-property-name` becomes `style.cssPropertyName`.  (Although IIRC some browser will accept either, at the very least Firefox wants the camel case version.)  That's probably the issue here.

> You can only add CSS styles to single elements with Javascript, right?

I'd guess you could set cascading styles by dynamically creating a style sheet?  I'm not really sure which approach would be best here.  Setting the DOM style when it hasn't changed should basically be a no-op, so it shouldn't cause performance issues to do it per entity.

> should the image smoothing be applied per entity?

Lets just worry about the global case right now; that should be the easiest to implement, and what most people would want.  Overriding the style on individual elements wouldn't be so hard to add later, if someone requests it.

mucaho:
`Crafty.DrawManager.setImageSmoothing(boolean): void` is the new method which enables/disables image smoothing for DOM and Canvas

Found another bug which prevents webkit from image smoothing when using DOM render method: [link1](http://code.google.com/p/chromium/issues/detail?id=134040) and [link2](http://code.google.com/p/chromium/issues/detail?id=106662).

Thus the most reliable way is to use Canvas and the Sprite component. I did not test this for IE and Opera though.

src/drawing.js
```diff
@@ -647,8 +647,64 @@ Crafty.DrawManager = (function () {
...
+         *
+         * @example
+         * This is the preferred way to draw pixel art with the best cross-browser compatibility.
+         * ~~~
+         * Crafty.canvas.init();
+         * Crafty.DrawManager.setImageSmoothing(false);
+         * 
+         * Crafty.sprite(imgWidth, imgHeight, "image.png", {image:[0,0]});
+         * Crafty.e("2D, Canvas, image"); // lower case "image", not the Component "Image"
+         * ~~~
```
starwed: Maybe switch "image" to something less confusing, like "spriteName".

src/drawing.js
```diff
@@ -647,8 +647,64 @@ Crafty.DrawManager = (function () {
             //reset DOM array
             dom.length = 0;
 
-        }
+        },
 
+        /**@
+         * #Crafty.DrawManager.setImageSmoothing
+         * @comp Crafty.DrawManager
```
starwed: Huh, this should probably be labeled as just `#.setImageSmoothing`, but since all the other methods of DrawManager are done this way, I'll just submit a separate patch to switch those.  (Once this lands.)

starwed:
Looks good, if you could just change that nit about the example code.  :)

kevinsimper:
:+1: Great contribution

This is a good step against all the weird things the different browsers do, but there is still IE which does not support this attribute, and iPhone and Macbook Retina :-)

But i think there so be a property which does not lie under `Drawmanager`, which sounds scary for beginners, it could be like `Crafty.pixelart(true)`;

I nearly think that imagesmoothing should be false as default, because when would you like this ugly scaling?
http://jsfiddle.net/VAXrL/1020/

starwed:
Definitely agree with a `pixelart` shorthand, that would be very friendly.  It could possibly do some other stuff, like coercing images to be drawn at integer coordinates.  (That can be a separate PR, though.)

> I nearly think that imagesmoothing should be false as default, because when would you like this ugly scaling?

The pixel art stuff works best when scaling by integers.  Check out [this fiddle](http://jsfiddle.net/VAXrL/1022/) in Chrome for a comparison between scaling from 64x64 to 77x77 with and without image smoothing.  I lean towards smoothing being the default, because turning it off only works if you've specifically designed your assets and game to work that way.

starwed:
Thanks, mucaho!

I'm going to go ahead and land this, but i realised there's a small flaw that'll need a followup.

If you call `pixelart` before the canvas context is created, the canvas flags won't be set correctly.  So it'll need to toggle a flag that canvas.init can look at, and if necessary apply those settings.

mucaho:
That's why I have included that in the documentation: "It will have no effect if canvas context is not yet created". And the example code explicitly creates the canvas context. Can't we just let the user initialize it manually?
**UPDATE**
or should we automatically create the context for the user by calling `canvas.init()` if the context does not exist yet?

kevinsimper:
> or should we automatically create the context for the user by calling canvas.init() if the context does not exist yet?

Better solutions :+1:

starwed:
> or should we automatically create the context for the user by calling canvas.init() if the context does not exist yet?

Nah, because some crazy people don't use the canvas at all, which is why it's created lazily in the first place.  :)

What we do is have a global flag that tells us what the current state is, and call `pixelart` with the correct flag when the canvas is created.

(We should also keep in mind that at some point in the future, we'll probably want to support having multiple canvas contexts.)

PR #667, opened 2013-12-25
Work with RequireJS

Hilzu:
require function in the compiled crafty.js expects the exported symbols the be defined in module.exports object. If using RequireJS Crafty was never added to the object. That leads to `var Crafty = require('./core.js'),` returning an empty object which leads to Crafty failing to initialize.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1626,9 +1626,9 @@ if (typeof define === 'function') { // AMD
     define('crafty', [], function () {
         return Crafty;
     });
-} else if (typeof exports === 'object') { // CommonJS
-    module.exports = Crafty;
 }
+
+module.exports = Crafty;
 
 window.Crafty = Crafty;
 


potomak:
The current code should work fine defining the `Crafty` AMD module only if `define` function is present.

Hilzu:
The problem isn't in defining the AMD module. When requiring the module Crafty fails initialization with `[Error] TypeError: 'undefined' is not a function (evaluating 'Crafty.c') (crafty.js, line 79)`. This is because the `var Crafty = require('./core.js')` call on line 11 in the release version returns an empty object because Crafty was not added to the exports object when created. The else if block is always executed if RequireJS isn't used. HashMap is added similarly to the exports object on line 2553 in the release version.

potomak:
Well, it should be already _exported_ at https://github.com/craftyjs/Crafty/blob/develop/src/core.js#L1629-L1631

potomak:
Note also that the code at https://github.com/craftyjs/Crafty/blob/develop/src/core.js#L1624-L1633 could be removed after using the `standalone` flag when compiling Crafty with browserify, see https://github.com/substack/node-browserify#usage and https://github.com/jmreidy/grunt-browserify#other-options

Hilzu:
The lines https://github.com/craftyjs/Crafty/blob/develop/src/core.js#L1629-L1631 aren't executed if RequireJS is present. `standalone` can only be used with a single entry point. Not sure if a file that would recursively need all other files in Crafty exists.

potomak:
Indeed, there's one only _entry point_ in Crafty: the `Crafty` module.

Am I still missing the point?

Hilzu:
If I understood the documentation of Browserify correctly the `standalone` switch needs a single file and bundles every file that it can find by recursively following `require` calls from the root file. I don't know if such file exists in Crafty and am no expert in Browserify so I can't really say if this problem can be fixed by configuring it differently.

potomak:
@Hilzu you're right about the entry point thing, I'm trying to figure out the best way to make Crafty usable with different loading techniques.

starwed:
@potomak  Does this PR cause problems for some other loading technique, then?

potomak:
I think that it could cause problems trying to _require_ Crafty in environments that are not using the CommonJS standard, where `module` and `exports` are not defined.

I tried a couple of solutions at https://github.com/potomak/Crafty/tree/feature/umd without success.

starwed:
Would changing the code to

``` javascript
 else if (typeof exports === 'object' || typeof module === 'object') { // CommonJS and RequireJS
     module.exports = Crafty;
 }
```

 work?  That at least would never cause a direct error.  But I've never really looked at how these systems work.

kevinsimper:
The reason why that AMD line does not works it is because it would only work if you did require the core.js file, which does not require any other files and therefore files like 2d and canvas will not load properply even through they are all in the same compiled file.

They way to get it to work would be to shim crafty, or by wrapping the whole compile crafty.js file in a AMD file defenition.

So this:

```
define('crafty', [], function () {
  // Compiled Crafty.js code here
  return window.Crafty;
});
```

or doing this in your require.js config to shim it, the `crafty-min` word should match the file name: 
http://requirejs.org/docs/api.html#config-shim

```
requirejs.config({
    shim: {
        'crafty-min': {
            exports: 'Crafty'
        }
});
```

kevinsimper:
@Hilzu Has you tried this? :)

Hilzu:
@kevinsimper Shimming Crafty doesn't work because the initialization of Crafty fails the same way as in my original description.

starwed:
@Hilzu Silly request: could you lay out the specific steps to reproduce this problem?  What you did, what happens, what you expect to happen?

Even if it seems obvious to you, because I've never used RequireJS I can't really follow your original report.  :)

Hilzu:
@starwed
The basic setup with RequireJS is to have this in your index.html head block

``` html
<script data-main="scripts/main" src="scripts/require.js"></script>
```

and then in your main.js file that was specified by the data-main attribute have something like this

``` JavaScript
require(["crafty"], function(Crafty) {
  Crafty.init(800, 600);
  Crafty.background('rgb(0, 43, 54)');
});
```

That causes RequireJS to look for a module named crafty from a crafty.js file. Let's assume it's the [current compiled unminified release version](https://github.com/craftyjs/Crafty/blob/release/dist/crafty.js). Evaluation reaches [line 6358](https://github.com/craftyjs/Crafty/blob/release/dist/crafty.js#L6358) and Crafty is added to the known modules of RequireJS with the name `crafty`.

Crafty uses CommonJS style require and exports and compiles the release file using Browserify that also assumes it. In the presence of RequireJS the addition of Crafty to the `module.exports` object is skipped which is incorrect, because other parts of the compiled crafty.js can't find it since they search it from the exports object.

The problem first arises on [line 79](https://github.com/craftyjs/Crafty/blob/release/dist/crafty.js#L79). The Crafty object retrieved by the require function on [line 11](https://github.com/craftyjs/Crafty/blob/release/dist/crafty.js#L11) from the exports object is empty because it wasn't added there. The require function is a CommonJS one and not a RequireJS one (this is correct) and is given as an parameter to the anonymous function on [line 10](https://github.com/craftyjs/Crafty/blob/release/dist/crafty.js#L10) after the minified setup code.

I hope this helps.

kevinsimper:
Have you tried taking out the define statement at line 6358 and wrap the whole crafty.js in a define statement?

```
define('crafty', [], function () {
  // Compiled Crafty.js code here
  return window.Crafty;
});
```

Hilzu:
@kevinsimper that does work when using RequireJS but would break when not using it for obvious reasons.

kevinsimper:
> that does work when using RequireJS

Cool

> but would break when not using it for obvious reasons.

I am sorry, but I do not understand that part of the sentence.

Hilzu:
@kevinsimper I don't see how wrapping the compiled crafty.js would fix this problem in Crafty without breaking it when not using RequireJS. The simple changes in this pull request makes Crafty work with or without RequireJS. The define statement works fine. Components like 2D and Canvas are added to the Crafty object known by RequireJS correctly with these changes.

starwed:
> I think that it could cause problems trying to require Crafty in environments that are not using the CommonJS standard, where module and exports are not defined.

@potomak  For now, module is defined by the code browserify injects.  So, I reckon it is ok to merge this PR?  (Or is there something I'm not understanding.)

I think this should be part of a 0.6.2 release, since it's a small change that enables functionality.

kevinsimper:
Made an simple example of how to use Crafty.js and Require.js together, also to test that is works.
https://github.com/kevinsimper/craftyjs-requirejs/

We can also delete the define function and just have `module.exports` and `window`.
https://github.com/kevinsimper/craftyjs-requirejs/blob/master/app/bower_components/crafty/dist/crafty.js#L6358-L6359

```
module.exports = Crafty;
window.Crafty = Crafty;
```

potomak:
@starwed yes, I think @Hilzu is right. It should work for everyone even with `module.exports = Crafty`.

starwed:
I'm going to go ahead and merge, but if we ever try to make Crafty usable _without_ browserify, we might want to revisit this.

PR #670, opened 2013-12-26
extracted spriteanimation

mucaho:
small steps towards #603


potomak:
I'm pro to extract components into single files, see #578, but I'd like to use the _delimiter-separated_ convention for file names, where _delimiter_ is `_`.

mucaho:
changed name to sprite_animation.js

potomak:
:+1:

I think that you also need to update `package.json` accordingly.

starwed:
> I'd like to use the delimiter-separated convention for file names, where delimiter is `_`.

What's the advantage?

potomak:
> What's the advantage?

Well, there are no advantages except for case-sensitiveness, it's only a convention, so let's choose a convention.

Note that almost all files in `src/` follow that convention except:
- `2D.js`, should be `2d.js`
- `DOM.js`, should be `dom.js`
- `DebugLayer.js`, should be `debug_layer.js` or `debug-layer.js`
- `HashMap.js`, should be `hash_map.js` or `hash-map.js`

I like that convention, it is also documented at "Google JavaScript Style Guide" even if that guide suggests to use `-` instead of `_`:

> Filenames should be all lowercase in order to avoid confusion on case-sensitive platforms. Filenames should end in `.js`, and should contain no punctuation except for `-` or `_` (prefer `-` to `_`).

More details at http://google-styleguide.googlecode.com/svn/trunk/javascriptguide.xml#Naming

mucaho:
PR will be updated after #665 hits (including the package.json filename string).

starwed:
@potomak  Makes sense!  I would probably prefer `-` to `_` as well, just because it's easier to reach on a keyboard.  :)

potomak:
So let's use `-`!

@starwed I proposed `_` only because it's the standard for Ruby file names and I mostly write Ruby code.

kevinsimper:
:+1: 

@mucaho  You updated the filename, but forgot that package.json filename string.

mucaho:
can we close this?

starwed:
Passes tests just fine, merging!

(Something slightly annoying about github is that it doesn't notify you when someone adds commits to a PR, so I didn't see that you'd updated this until you commented.)

PR #671, opened 2013-12-27
Adjust bounding region to include hitbox when it exists outside of the entity

### LINKED ISSUE #412: Collision detection fails when collision polygon lies outside of entity
I'm not sure if this is intended behavior or not, but if you define a collision polygon larger than the object that it is attached to, it will not get picked up in a call to .hit. In my reading of .hit:
It first does a broad search for objects intersecting with the rectangle that is defined by the objects x/y coordinates and height/width:

``` javascript
results = Crafty.map.search(area, false),
```

Then it runs the SAT collision calculations only if the broad phase search returned results:
So, if you do something like (where SomeSprite is a 5x5 sprite you've already set up):

``` javascript
Crafty.e("2D, DOM, SomeSprite, Collision, Test").attr({ x: 5, y : 5})
         .collision([0, 0], [10, 0], [5, 5], [0, 5]);
var other = Crafty.e("2D, DOM, SomeSprite, Collision").attr({ x : 8, y : 8 });
var hits = other.hit("Test");
```

hit won't have any results.

starwed:
Introduce a new bounding rectangle `._cbr` (collision bounding rectangle) that is guaranteed to include both the entity and it's hitbox.
When the hitbox sits inside the entity, the CBR will not be used.

Also, simplify the way `_calculateMBR` is called.  A follow up patch could optimize that function for the case where no rotation exists.

This patch also adds tests and a small note to the docs that external hitboxes might have a small performance penalty.

(This fixes #412)

### DIFF
--- a/src/2D.js
+++ b/src/2D.js
@@ -353,7 +353,8 @@ Crafty.c("2D", {
 
         //when object changes, update HashMap
         this.bind("Move", function (e) {
-            var area = this._mbr || this;
+            // Choose the largest bounding region that exists
+            var area = this._cbr || this._mbr || this;
             this._entry.update(area);
             // Move children (if any) by the same amount
             if (this._children.length > 0) {
@@ -362,7 +363,8 @@ Crafty.c("2D", {
         });
 
         this.bind("Rotate", function (e) {
-            var old = this._mbr || this;
+            // Choose the largest bounding region that exists
+            var old = this._cbr || this._mbr || this;
             this._entry.update(old);
             // Rotate children (if any) by the same amount
             if (this._children.length > 0) {
@@ -423,7 +425,7 @@ Crafty.c("2D", {
         this._by1 = y1;
         this._by2 = y2;
         this.trigger("BoundaryOffset");
-        this._calculateMBR(this._origin.x + this._x, this._origin.y + this._y, -this._rotation * DEG_TO_RAD);
+        this._calculateMBR();
         return this;
     },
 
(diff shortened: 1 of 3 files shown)


starwed:
Added some comments that explain what's going on with this code.

starwed:
I guess I should mention that I tried implementing the [Ritter algorithm](https://github.com/erich666/GraphicsGems/blob/master/gems/BoundSphere.c) for a better bounding circle, but it didn't seem to really improve things for the very simple point sets I tried.

potomak:
@starwed nice to know, maybe you could also cite the algorithm in the code.

> but it didn't seem to really improve things for the very simple point sets I tried.

Are you making performance benchmarks?

starwed:
> Are you making performance benchmarks?

In this case, I was just drawing the resultant bounding boxes.  In at least one case the Ritter method actually produced larger bounding regions, and it required the code to be [significantly more complex.](https://github.com/starwed/Crafty/blob/cbr2/src/collision.js#L116-L164)

The Ritter algorithm apparently gets within 20% of the optimal circle in the general case.  The naive way I implemented will be 41.4% too large in the absolute worst case scenario.  That would be something like a square rotated 45 degrees -- any time when the minimal bounding circle exactly inscribes the minimal bounding square.  ~~But actually, that particular scenario is also just as bad for the Ritter algorithm.~~  (Nope!)

If we were to implement a non-naive algorithm, it might be better to actually find the minimal circle, given that you pay the cost only when assigning the hitbox.  But since this is just used for the broad phase pass, I don't _think_ it would really make much of a difference.

All that said, it would be really nice to have concrete benchmarks of some sort!

starwed:
Added a link in the comments to the [wikipedia article](http://en.wikipedia.org/wiki/Bounding_sphere) on bounding spheres, which describes a few different algorithms.

starwed:
Ok, I'm going to go ahead and land this.

Issue #669, opened 2013-12-26
SpriteAnimation Examples not working


mucaho:
The examples listed in the [SpriteAnimation API](http://craftyjs.com/api/SpriteAnimation.html) do not work for me.

In particular when defining the reel the automatic way

``` javascript
// Define an animation on the second row of the sprite map (y = 1)
// from the left most sprite (fromX = 0) to the fourth sprite
// on that row (toX = 3), with a duration of 1 second
Crafty.e("2D, DOM, SpriteAnimation, PlayerSprite").reel('PlayerRunning', 1000, 0, 1, 3);
```

it does not work. When defining the reel manually

``` javascript
// This is the same animation definition, but using the alternative method
Crafty.e("2D, DOM, SpriteAnimation, PlayerSprite").reel('PlayerRunning', 1000, [[0, 1], [1, 1], [2, 1], [3, 1]]);
```

it works.

potomak:
I used the first signature without problems in Loop Desert only one week ago, see source at http://giovanni.curlybrackets.it/ld28/loop-desert.js and example at http://giovanni.curlybrackets.it/ld28/

mucaho:
Thx potomak, I have investigated further using the official Crafty v0.6.1

``` html
<html>
  <head></head>
  <body>
    <div id="game"></div>
    <script type="text/javascript" src="crafty.js"></script>
    <script>
        Crafty.init(600, 300, document.getElementById('game'));
        Crafty.background('rgb(127,127,127)');

        // Define a sprite-map component
        Crafty.sprite(16, "squares.gif", {
            PlayerSprite: [0,0]
        });

        var ent = Crafty.e("2D, Canvas, PlayerSprite, SpriteAnimation"); //"DOM" does not work on Chrome for ANY reel animation

        //ent.reel('PlayerRunning', 1000, 0, 0, 1); //this does not work
        ent.reel('PlayerRunning', 1000, [[0, 0], [1, 0]]);

        ent.animate('PlayerRunning', -1);
    </script>
  </body>
</html>
```

The above code does not work with DOM rendering on Chrome.
The commented reel definition does not work (tested in FF&Chrome).

starwed:
> The above code does not work with DOM rendering on Chrome.

This sounds like #664, which should be fixed in the develop branch now.  I suspect that was a buggy behavior in webkit, but I couldn't figure out how to reproduce it outside of Crafty.

> The commented reel definition does not work (tested in FF&Chrome).

[...]

PR #672, opened 2013-12-28
Remove stray occurrences of toX in code and docs

starwed:
Both the docs and the code still referenced toX -- the former causes confusion, the latter an accidental global variable!

Fixes an issue raised by #669.


kevinsimper:
:+1:  Great

PR #673, opened 2013-12-28
Delete trailing spaces.

starwed:
An updated version of pawurb's patch #634.

> How about cleaning all the trailing spaces in the project ?


starwed:
I went through and ensured that this only changes the end spaces -- there are no other accidental changes.

The only merge conflicts were in the tests.

This includes the new jshint check for spaces, so they shouldn't creep into the codebase in the future.

PR #674, opened 2013-12-30
Edit docs for formatting

starwed:
There are some formatting issues on several of the automatically generated [doc pages](http://craftyjs.com/api/).  This patch is for cleaning up formatting and (where I notice them) grammar issues.

The basic idea here is to just skim the docs, and fix anything that jumps out at me.  :)  It'll be ready to land when we've inspected each of the following modules for obvious problems:
- [x] Core
- [x] Game Loop
- [x] Events
- [x] 2D (But not the math stuff)
- [x] Graphics
- [x] Utilities
- [x] Misc
- [x] Stage
- [x] Input
- [x] Animation
- [x] Scenes
- [x] Audio
- [x] Assets
- [x] Debug

### DIFF
--- a/src/2D.js
+++ b/src/2D.js
@@ -72,7 +72,7 @@ Crafty.extend({
  * Component for any entity that has a position on the stage.
  * @trigger Move - when the entity has moved - { _x:Number, _y:Number, _w:Number, _h:Number } - Old position
  * @trigger Change - when the entity has moved - { _x:Number, _y:Number, _w:Number, _h:Number } - Old position
- * @trigger Rotate - when the entity is rotated - { cos:Number, sin:Number, deg:Number, rad:Number, o: {x:Number, y:Number}, matrix: {M11, M12, M21, M22} }
+ * @trigger Rotate - when the entity is rotated - { cos:Number, sin:Number, deg:Number, rad:Number, o: {x:Number, y:Number}}
  */
 Crafty.c("2D", {
     /**@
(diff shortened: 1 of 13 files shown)


starwed:
I'll note more complicated issues here, since this PR is just for cleaning the formatting:
- Crafty.modules should mention that there are issues with the components site.
- The SaveData and LoadData events aren't documented in the standard way.  (Storage component.)
- Crafty.extend has basically no documentation (core.js)
- Several of the docblocks in HashMap.js aren't being parsed?
- Crafty.isometric.centerAt missing docs for parameters
- The stuff in math.js apparently was assuming a different docblock style than the rest of the codebase -- there are tons of issues there.

Random thoughts
- We use triple tilde for codeblocks; it might be nicer to use triple backticks like github does?  Also, syntax highlighting for codeblocks would make everything look a lot better.

starwed:
Ok, I've now gone through all the doc pages and tried to fix any gross errors that jumped out at me.

The math.js docs seemed to have enough systematic problems that this should probably be a separate patch.  They seem to have a different style of docs that breaks the new parser.

There are some places where I need to make sure that the docs still look correct -- a bit tricky because of the way docs are generated right now.

potomak:
@starwed about `Crafty.modules` I think we should just remove that part. I'm working on a new version of Crafty Components, see https://github.com/potomak/CraftyComponents, it is a node.js app and it will be only a simple directory of components without any CDN feature.

We could just use browserify to load npm packages _a la_ voxel, more details at http://voxeljs.com/#modules

starwed:
Ok, built the docs, fixed a few issues, and rebased to one big commit.

Since this only touches the documentation, going ahead and landing.

PR #420, opened 2013-02-15
rewrite Crafty.unbind

sbyrnes321:
see discussion at issue 417 --
https://github.com/craftyjs/Crafty/issues/417

This implements (and documents) the behavior suggested by @mpetrovic at https://github.com/craftyjs/Crafty/issues/417#issuecomment-13556902

I haven't had time to test this patch, I'm just posting it now to contribute to the conversation.

PR #676, opened 2014-01-01
Add unbinding by bind reference back

kbrock:
I'm new to craftyjs, and following through a great tutorial on http://buildnewgames.com/introduction-to-crafty/

While upgrading the tutorial to 0.6.1, the first 2 things that I've noticed that broke are:
1. The `animate`/`reel` portion of the code.
2. The `unbind` behavior on the victory page.

The `animate` code is well documented and trivial to track down and fix. Both due to the error messages as well as the clear documentation of the incompatibility.

The `unbind` code was more difficult to track down. It ate the error message (not saying it should be any different) and while #420 ( commit 35d9585caaf17315b ) said it was a rewrite, it highlighted removing the ability to `bind` to regular elements rather than global elements. It said nothing of removing the ability to `unbind` with the reference returned by `bind`. This seems like a simple oversight and possibly wants to make its way back into the code.

Example of code that worked in the 0.5.x but not the 0.6.x branch:

``` javascript
var ref = bind('KeyDown', function() { console.debug('x') });
unbind('KeyDown', ref)
```

The solution is quite easy, but I wonder if there is reason to break this.
working code:

``` javascript
function printDebug() { console.debug('x'); }
bind('KeyDown', printDebug);
unbind('KeyDown', printDebug);
```

The alternatives are to document this change, or have `bind` return the command function passed in, so unbind will work unchanged. **UPDATE**: it is now returning the command passed in, so `one()` and `bind()` both work.

If nothing else, I just wanted to document this changed behavior in an issue, so someone else coming along may be able to stumble upon the fix.

Thanks for the great library,
K

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1366,7 +1366,7 @@ Crafty.extend({
      * @sign public Number bind(String eventName, Function callback)
      * @param eventName - Name of the event to bind to
      * @param callback - Method to execute upon event triggered
-     * @returns ID of the current callback used to unbind
+     * @returns callback function which can be used for unbind
      *
      * Binds to a global event. Method will be executed when `Crafty.trigger` is used
      * with the event name.
@@ -1398,7 +1398,8 @@ Crafty.extend({
         var hdl = handlers[event];
 
         if (!hdl.global) hdl.global = [];
-        return hdl.global.push(callback) - 1;
+        hdl.global.push(callback);
+        return callback;
     },
 
 
@@ -1408,7 +1409,7 @@ Crafty.extend({
      * @sign public Number uniqueBind(String eventName, Function callback)
      * @param eventName - Name of the event to bind to
      * @param callback - Method to execute upon event triggered
-     * @returns ID of the current callback used to unbind
+     * @returns callback function which can be used for unbind
      *
      * Works like Crafty.bind, but prevents a callback from being bound multiple times.
      *
@@ -1416,8 +1417,7 @@ Crafty.extend({
      */
     uniqueBind: function (event, callback) {
         this.unbind(event, callback);
-        this.bind(event, callback);
-
+        return this.bind(event, callback);
     },
 
     /**@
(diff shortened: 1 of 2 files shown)


starwed:
I think if we wanted to add this back in, it wouldn't be by the wonky id system.  Instead, `bind` would just return a direct reference to the function itself, and you'd unbind it in the standard way.

This would also be useful for `one()`, which otherwise (due to how its implemented) could not be unbound.

kbrock:
@starwed thank you for your feedback. I re-implemented using the callback as the return.

`one()` should now work as well.

I am unable to get grunt working on my computer, so I kinda winged the tests.

src/core.js
```diff
@@ -1398,7 +1398,8 @@ Crafty.extend({
         var hdl = handlers[event];
 
         if (!hdl.global) hdl.global = [];
-        return hdl.global.push(callback) - 1;
+        hdl.global.push(callback) - 1;
+        return callback;
```
starwed: The `-1` is doing nothing now!  :)  Just remove it.

starwed:
This would need to change the documentation of `bind` as well, I think.

> I am unable to get grunt working on my computer, so I kinda winged the tests.

Was it that grunt wouldn't install, or that it didn't run correctly once installed?  (Also, what platform are you on, etc.)

kbrock:
@starwed - done
1. fixed documentation
2. ran tests via `grunt check` and they all pass (needed to installed `grunt-cli` in addition to `grunt` - not a js developer)

kevinsimper:
@kbrock You can see here how to build Crafty.js https://github.com/craftyjs/Crafty/wiki/Building (Linked from the frontpage)

kbrock:
@kevinsimper Thank you

Sorry I left a cryptic comment. What I was trying to say was:

I figured out that I needed to install grunt-cli.

All tests pass.

starwed:
Looks good, landing.

Thanks for the PR!

PR #679, opened 2014-01-02
Clean up test and put them all in javascript files

kevinsimper:
I made a lot of changes in this pull request, many of them are just small increment changes, so it is not that overwhelming as it may seem.

I have moved all the test to javascript files and then included them in a index.html file.
I have done this so it is much easier to test in real browsers, so you do not have to type in all the different files. 
QUnit groups them in modules which can be access at the upper right corner.
It is also possible to run only specefic modules, but using url parameters like: 
http://localhost:9000/tests/?module=Events

image

I did also break out the 2D test module, which had become large enough to be in its own file.

Now the test can also easily be run without using any compatilitets settings by using harpjs (harpjs.com);
First install harp.js, `npm install harp -g` and then in the crafty repo run the following command, `harp server`.
Then navigate to http://localhost:9000/tests/ when you have run `grunt check` first to generate crafty.js file first.

I have also made so that every module run `Crafty('*').destroy();` after every test, so the person who is writing test do not need to remember to clean up.

I also made `grunt check` run on the test to ensure quality, and also corrected all the errors that did came up.


Gruntfile.js
```diff
@@ -86,22 +86,12 @@ module.exports = function (grunt) {
...
-                'tests/dom.html',
-                'tests/tween.html',
-                'tests/sound.html'
+                'tests/index.html'
             ]
         },
 
         jsvalidate: {
-            files: "crafty.js"
+            files: ["crafty.js", 'tests/*.js']
```
potomak: For the sake of consistency I'd use `'` instead of `"`.

potomak:
Nice clean up.

I'd suggest connect through `grunt-contrib-connect` as static server instead of harpjs (no global install and included as a development dependency).

starwed:
Just a couple of questions:
- What does this do with the "playground" functionality of the animation test page?
- Any noticeable change in how long it takes to run `grunt check` this way?

Thumbs up to having the module code cleanup properly.

kevinsimper:
@potomak Thanks.

I do not want to be defiant, but i am little against putting in a disposable webserver into the crafty project, because there is a lot of ways you can start a disposable webserver without any configuration.
I think Harpjs is a good project for its nice features, but you can infact you any disposable webserver, whether it's php, python, rubys.

@starwed Nothing, it should work as it has done all the time :) No, it has gone under 1 sec the QUnit tests, but since JSHint and JSValidate now also checks the test files it takes a second more now. I was thinking about making another grunt task for checking tests only, but then I think that task should be a part of the workflow so it makes sense also checking them along the sourcecode.

starwed:
Heh, for me `grunt qunit` takes a full minute to run, whereas the other steps are almost instantaneous.

(`grunt qunit` will run only the tests, although it doesn't build crafty beforehand.)

kevinsimper:
I was wrong, this helps a lot on the time it takes to complete the tests. Although yours is a lot longer that mine, so you should see a greater improvement.

Current takes 16 secs for `qunit` and 18 for `check`.

With this it takes 4 secs for `qunit` and 9 secs for `check`.

kevinsimper:
You did also broad to my attention that I forgot to include the animation tests, they are now included also.

mucaho:
@kevinsimper wrote:

> I have done this so it is much easier to test in real browsers, so you do not have to type in all the different files. 

Is there an already existing way to run `grunt qunit` for a single test file (instead of all the tests)?

kevinsimper:
@mucaho 

> Is there an already existing way to run grunt qunit for a single test file (instead of all the tests)?

Yes, you can open the html file directly in a browser. But not programmerable.

potomak:
@mucaho you could build a task that get a module name from stdin using http://gruntjs.com/api/grunt.option and pass it to qunit using the `module` query param. Note that you should also change qunit task to get URLs instead of files.

starwed:
This needs to be updated against the most recent HEAD.

Also, maybe `tests.html` or `qunit.html` would be a bit more descriptive than index?

mucaho:
Thanks kevin and potomak!

potomak:
@starwed I like `index.html` as it could be accessed from `localhost:port/tests/`.

kevinsimper:
> @starwed I like index.html as it could be accessed from localhost:port/tests/.

That was also what i thought of.

starwed:
> I like index.html as it could be accessed from localhost:port/tests/.

Makes sense!

starwed:
Also, `grunt check` is like an order of magnitude faster for me now.  I guess loading and setting up qunit only once makes things much faster!  :D

kevinsimper:
> Also, grunt check is like an order of magnitude faster for me now. 

And it is even doing more things now ;)

PR #680, opened 2014-01-02
Add grunt-contrib-connect package and connect task

potomak:
`grunt connect` task can be used to run a local HTTP server.

See also #679

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -104,6 +104,14 @@ module.exports = function (grunt) {
             files: "crafty.js"
         },
 
+        connect: {
+            server: {
+                options: {
+                    keepalive: true
+                }
+            }
+        }
+
     });
 
     // Load the plugin that provides the "uglify" task.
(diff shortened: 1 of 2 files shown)


starwed:
Do you have an estimate of about how large this+dependencies ends up being?

Something like this would be nice, but we want to make sure our install size/time doesn't get out of control.  Last time I checked the total was about 30MB.  If this is just a small fraction of that, no problem!

potomak:
I tested with a dummy package including only `grunt-contrib-connect`, it's 14 MB, but I think that there are shared dependencies between modules.

Moreover I think that the total size of the package doesn't matter as soon as you include dependencies through npm:
- Crafty repo: _net_
- npm dependencies: _tare_
- development installation: _gross_

starwed:
I see that grunt used to have a built-in server task, and this was moved to `grunt-contrib-connect`.   So it seems like this is in some sense the "official" way to do this, in which case it's probably a good idea to prefer it over other options.  It also means it probably makes some effort to share dependencies with other mainstream grunt plugins.  

(npm will often install multiple versions of the same dependency if different plugins need it; that's how it avoids conflicts.)

PR #681, opened 2014-01-04
Fix the matrix inversion test to avoid floating point issues.

### LINKED ISSUE #677: Math Matrix2D invert fails in Chrome
There is one test which fails in Chrome, but not in PhantomJS, Safari or Firefox.

It is because of the long decimals:

image
image

But then i saw in Firefox that the e value switch between -0, and 0;
image
image

In safari is there no problems.

starwed:
The current matrix inversion test fails in Chrome due to reliance on very particular floating point errors.

Best just to do the test with a matrix that has an inverse consisting of only integers.

This fixes #677

### DIFF
--- a/tests/math.js
+++ b/tests/math.js
@@ -247,8 +247,9 @@ test("determinant()", function() {
 });
 
 test("invert()", function() {
-  equal((new Matrix2D()).scale(2, 3).rotate(Math.PI / 2).invert().equals(new Matrix2D(3.061616997868383e-17, -0.3333333333333333, 0.5, 2.041077998578922e-17, 0, 0)),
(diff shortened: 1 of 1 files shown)


starwed:
Huh, not sure why git thinks the change applies to every line.

kevinsimper:
Cool, but yeah, it is weird that is thinks it is only one line.

starwed:
Yeah, I double-checked using diffmerge, there don't seem to be any accidental changes introduced.  Not sure what's going on -- maybe some whitespace issue that diffmerge ignores?

kevinsimper:
Hmm, the original tests/math.html has normal lines.

But could this be merged after my cleanup? I have already moved this code to is own javascript file.

starwed:
Ok, landing now that Kevin's cleanup has gone through.

PR #688, opened 2014-01-09
Attempt to fix error with Crafty.frame()

highwaycoder:
Was returning undefined, should now return the frame number as expected.  Untested though.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -33,7 +33,7 @@ var Crafty = function (selector) {
     return new Crafty.fn.init(selector);
 },
     // Internal variables
-    GUID, frame, components, entities, handlers, onloads,
+    GUID, frame=0, components, entities, handlers, onloads,
     noSetter, slice, rlist, rspace, milliSecPerFrame;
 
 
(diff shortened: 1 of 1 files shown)


highwaycoder:
Needs to be tested before it's merged, I'm moving on to other things in the mean time.

kevinsimper:
Thanks for the PR. 

Where was it undefined? Code? By moving a variable could indicate you are asking for the frame number before Crafty has started. 

Also you should check the code by running grunt check. You can see how in the Contribute file.

highwaycoder:
http://jsfiddle.net/pnZN7/

Demonstrates the problem quite nicely, I think.  I'm assuming by the time an EnterFrame event occurs, Crafty is definitely running, yes?

I haven't got a local WC of this repo at the moment and I have a fair amount on my plate right now so it'll be a while before I can test this fix myself (doesn't occur to me to be worth much effort given how you can get the frame count from the EnterFrame event anyway).  If my schedule calms down and nobody's fixed this or tested my fix, I'll put some more effort into it at that point.

kevinsimper:
Submitted a patch for this at #689

highwaycoder:
I'll close this PR then.

PR #689, opened 2014-01-10
Crafty.frame function not working - Fixes #688

kevinsimper:
The frame function is initiated in the Crafty object but also in the Crafty.timer function as a local variable.

Thanks to @cjbrowne for pointing it out.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -46,7 +46,8 @@ var Crafty = function (selector) {
 
 
     initState = function () {
-        GUID = 1; //GUID for entity IDs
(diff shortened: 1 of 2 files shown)


kevinsimper:
Also, @starwed, what is the best way to reset Crafty, because the frame number is 3 in this test i just created?
It would be nice if Crafty could start from zero in each test, but if you do `Crafty.stop(true)` and then `Crafty.init()`, a lot of other tests fails.

kevinsimper:
Going to merge it.

Issue #663, opened 2013-12-20
Crafty.init() prevents propagation of `touchmove` events


ryansmith3136:
The problem is that other JS libraries (e.g. scrolling libraries) might rely on this event. Crafty silently prevents the touchmove event from being executed.

Perhaps `Crafty.stop` can remove this listener?

ryansmith3136:
Relevant code: [viewport.js line 16](https://github.com/craftyjs/Crafty/blob/ca1d96d825d4b37013c2b7248eba823bcb0b6d4b/src/viewport.js#L618-620)

potomak:
I had the same problem but I didn't investigate, thanks @ryandotsmith.

ryansmith3136:
Here is a patch that we applied to our local copy of crafty.

https://gist.github.com/ryandotsmith/8057623

kevinsimper:
That is a relevant problem and should be optional.

But i do not see how your gist change anything, you just moved the anonymous function to a named function?

starwed:
> But i do not see how your gist change anything, you just moved the anonymous function to a named function?

They want to unbind the function when Crafty.stop() is called.  To unbind the function, it's necessary to have a reference to it -- impossible with anonymous functions.

ryansmith3136:
That is correct. A named function makes it possible for us to remove the listener.

Here are the docs regarding removing listeners: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget.removeEventListener

kevinsimper:
Ah, the first thing you get blind on is the eyes.

potomak:
> The problem is that other JS libraries (e.g. scrolling libraries) might rely on this event. Crafty silently prevents the touchmove event from being executed.

I think that the problem is worse than that because Crafty is preventing default on `window` so scrolling is completely disabled on touch screens.

starwed:
That's normally what you'd want with games, of course.  But there should definitely be a way to disable it.

potomak:
> That's normally what you'd want with games

when they are full screen...

> But there should definitely be a way to disable it.

I agree.

ryansmith3136:
> On Dec 29, 2013, at 16:11, starwed notifications@github.com wrote:
> 
> That's normally what you'd want with games, of course.

This is true, but we were dealing with a home screen as well. So in this case, crafty modified the global environment which consequently affected screens which were outside of the game.

kevinsimper:
> when they are full screen...

[...]

PR #690, opened 2014-01-10
Remove mobile conditions

kevinsimper:
Remove a lot of mobile behavior, which as smartphones gets bigger screens it not wanted behavior.

Removing scroll on viewport is not something you want as it also happends on a iPad for an example.
The window.scrollTo does not work on iOS7 anymore, so that is also useless now, also because it will not scroll if the viewport height is equal to the document height.
Setting that the touchmove on is ignored on the Crafty element and not on the whole page, so that the webpage is still functional.

There should properly be something like this, but is should be optional and thought through, which this is not.

### DIFF
--- a/src/viewport.js
+++ b/src/viewport.js
@@ -467,7 +467,7 @@ Crafty.extend({
          * @param Number height - Height of the viewport
          * @param String or HTMLElement stage_elem - the element to use as the stage (either its id or the actual element).
          *
-         * Initialize the viewport. If the arguments 'width' or 'height' are missing, or Crafty.mobile is true, use Crafty.DOM.window.width and Crafty.DOM.window.height (full screen model).
+         * Initialize the viewport. If the arguments 'width' or 'height' are missing, use Crafty.DOM.window.width and Crafty.DOM.window.height (full screen model).
          *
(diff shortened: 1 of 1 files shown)


starwed:
I'm not sure if we want to completely strip this away.

Having an option to disable it is one thing, or even making it off by default.  But in a lot of games you _would_ want it to be fullscreen on mobile.

kevinsimper:
Yes, and as mentioned there should be something like this, but this has unintended effects like you can't scroll in the tests when you load /test/index.html.
Also on craftyjs.com if you are on an iPad and there is an example you can not scroll.

There should be something like "do fullscreen on mobile devices" option, when you init the game.

There is also some in the forum that complain about it that Crafty does not respect the values in the init function.

For an example (and they are just the ones who took the time to complain)
#663 
https://groups.google.com/forum/#!searchin/craftyjs/mobile/craftyjs/HtFTCHFfAko/-rLp-Tdbc5kJ
https://groups.google.com/forum/#!searchin/craftyjs/mobile/craftyjs/4tGr_2ou964/hlorng3W6zkJ
https://groups.google.com/forum/#!searchin/craftyjs/mobile/craftyjs/NXu4F8b_-Fg/cguuyMvjq7QJ 
https://groups.google.com/forum/#!searchin/craftyjs/mobile/craftyjs/pwvCiJetMN4/dbUaoBNbqVUJ

kevinsimper:
This is an important fix and relative small change, because if you want to go fullscreen on mobile you can just do `Crafty.init()`, just like you would on the computer.

These changes are easy to add yourself, but hard to remove from Crafty when used.

PR #691, opened 2014-01-11
Abstract viewport resizing

starwed:
Add setters/getters for the viewport, and trigger a "ViewportResize" event.  Use that event to alter the dimensions of the stage and canvas appropriately.

There are several places in the viewport code where it's effect on the canvas and DOM is hardcoded.  Using events instead of such special cases will make it easier to land something like a webgl backend, or support multiple canvas/DOM layers.

Adding setters for the viewport width/height should make it easier to change the game size.  That should partially help with the problem described [here](https://groups.google.com/forum/?fromgroups#!topic/craftyjs/gE1oEqh-pQ0).

### DIFF
--- a/src/DOM.js
+++ b/src/DOM.js
@@ -298,8 +298,15 @@ Crafty.extend({
                 this.height = window.innerHeight || (window.document.documentElement.clientHeight || window.document.body.clientHeight);
 
                 // Bind scene rendering (see drawing.js)
-                Crafty.unbind("RenderScene", Crafty.DrawManager.renderDOM);
-                Crafty.bind("RenderScene", Crafty.DrawManager.renderDOM);
+                Crafty.uniqueBind("RenderScene", Crafty.DrawManager.renderDOM);
+                // Resize the viewport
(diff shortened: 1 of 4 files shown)


kevinsimper:
Nice edits :+1: 

A question about the tests, now you are modifing the viewport size, which properly also affects the other tests (not anything serious), but what is the best way to reset it?

starwed:
> A question about the tests, now you are modifing the viewport size, which properly also affects the other tests (not anything serious), but what is the best way to reset it?

This has actually been a problem with tests in the past -- some of them were accidentally depending on the viewport size, and thus passing the phantomjs tests but failing in a real browser.   So at any rate, the fact that my test changes the viewport shouldn't matter, because tests that care about that need to set the viewport directly.

Is it possible in qunit to run the same module multiple times?  It might be interesting to do that with different viewport sizes in some cases.

kevinsimper:
Not that i know of, but you could make the test lookup a url parameter which sets the viewport, or through a large for loop around the tests, that should also work.

PR #692, opened 2014-01-11
New storage based on localStorage

kevinsimper:
I made a new Crafty.storage function which is based on the localStorage function which is very presence in the browsers we support. http://caniuse.com/namevalue-storage

I choose to make it a single function just as we do with `.attr`.

The object if that is passed is `JSON.stringify` and tried parsed when you try to get it again.

I have not made any function like, `.hasKey`, because you should try to get an value and then deal with the result instead of doing two calls, first `.hasKey` and then after `.storage`. That would make you hit localStorage 2 times instead of just one. Made an example to show how you should do it.

There could be an fallback to cookiesupport, but as seen on the caniuse.com site, localStorage is very supported so  i do not think it has to be more complicated than it should.

### DIFF
--- a/src/storage.js
+++ b/src/storage.js
@@ -8,608 +8,69 @@ var Crafty = require('./core.js'),
 /**@
  * #Storage
  * @category Utilities
- * Utility to allow data to be saved to a permanent storage solution: IndexedDB, WebSql, localstorage or cookies
+ * Very simple way to get and set values, which will persist when the browser is closed also.
  */
 /**@
- * #.open
+ * #.storage
  * @comp Storage
- * @sign .open(String gameName)
- * @param gameName - a machine readable string to uniquely identify your game
+ * @sign .storage(String key)
+ * @param key - a key you would like to get from the storage. It will return null if the key does not exists.
+ * @sign .storage(String key, String value)
+ * @param key - the key you would like to save the data under.
+ * @param value - the value you would like to save.
+ * @sign .storage(String key, [Object value, Array value, Boolean value])
+ * @param key - the key you would like to save the data under.
+ * @param value - the value you would like to save, can be an Object or an Array.
  *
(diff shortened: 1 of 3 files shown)


starwed:
I think we would not want to use local storage except as a fallback; see [here](https://blog.mozilla.org/tglek/2012/02/22/psa-dom-local-storage-considered-harmful/) for instance.  The tldr is that local storage is synchronous, which might have especially bad performance implications in a game.

We had a working implementation of indexedDB at some point; I think the main problem is that it was  created before the spec had been finalized.  I haven't taken the time to look at what's going on, but I think it's mostly just some slight API changes.  Assuming that it really was working in the past, it shouldn't be that much work to update it.

kevinsimper:
Yes, indexedDB is cooler and localStorage is sync, but i do not think that is a problem because it should only be used when you save things and not in the loop.

LocalStorage is fast. It clocks in at about 1.4 million ops with getItem on a macbook pro with 2.5 harddrive.
and 40 thousand in write ops. This could be improved by providing a simple object cache. 

IndexedDB is not supported in IE9 and Safari and iOS, but they support WebSQL, but that is not supported by IE and Firefox. That makes a nightmare.

This localStorage could also be a component and thereby optional. I think most just want a key value store and something to persist on a server anyway, where they could use many of these no-backend services for storing values nice and simple. This would just be something if you wanted quick and dirty storage for your game without searching for a new library or learn something new you properly do not need. In the previous you need to `.open` a database and such, ugh!

This does also takes size into consideration. This well pointed to indexedDB polyfill fills 26 KB minified. That would increase minified Crafty but 30% as it is now 96 KB.
http://nparashuram.com/IndexedDBShim/
Also the increased complexity with callbacks and error handling makes it a weaker choice for a beginner javascript game engine.

kevinsimper:
And if they want more advanced storage on the client we should point them to PouchDB which is awesome and have a great website: http://pouchdb.com/

PR #693, opened 2014-01-13
Create the Crafty model component and core helper method.

### LINKED ISSUE #683: Crafty.m('Score') - Crafty Models!
What are thoughts on adding model support? Components I feel have logic but should only have view-related logic like position, movement, keyboard shortcuts, etc. Where a model might have more business logic.

For instance you might win or lose the game when one score object gets to 10? I might also need to sync that with the server to keep track of things. One model might also be shared across multiple components as one component displays the score, and another increases it when I score.

I can see it being used like this:

``` javascript
// Scene
var score = Crafty.m('Score').attr({points: 0});
Crafty.e('Score').setupScore(score);
Crafty.e('Ball').setupScore(score);

// Model
Crafty.m('Score', {
    init: function() {
      this.points = 0;
    },
    increment: function() {
      return this.set('points', this.get('points') + 1);
    },
});

// Components
Crafty.c('Score', {
  init: function() {
    this.requires('DOM, 2D, Text, Grid')
      .text('0 Points')
      .attr({w: 300});
  },
  setupScore: function(score) {
    this.score = score;
    this.score.on('change:points', this.updateScore.bind(this));
    return this;
  },
  updateScore: function() {
    this.text(this.score.get('points') + ' Points');
    return this;
  },
});
Crafty.c('Ball', {
  init: function() {
    // stuff
  },
  setupScore: function(score) {
    this.score = score;
    return this;
  },
  gainPoint: function() {
    this.score.increment();
    return this;
  },
});
```

The model events are import to noticing differences:

```
this.score.on('change:points', function() {});
```

I'd be happy to submit a pull request if I knew it was wanted in this repo. If not maybe this could be extracted to another library, or maybe another library could be recommended for doing this for me? Thank you for your time.

bayleedev:
All tests pass, any thoughts on getting travis setup to automate this?

I got the docs to generate correctly, but if there are some standards I'm not following let me know and I'll fix asap!

Fixes #683

### DIFF
--- a/src/model.js
+++ b/src/model.js
@@ -250,15 +250,3 @@ Crafty.c('Model', {
     }
   }
 });
-
-/**@
- * #Crafty.m
- * @category Core
- * Model entity helper.
- * @see Model
- */
-Crafty.extend({
-  m: function(name) {
-    return this.e(name + ', Model');
-  }
-});
--- a/src/model.js
+++ b/src/model.js
@@ -1,0 +1,264 @@
+var Crafty = require('./core.js'),
+    document = window.document;
+
+/**@
+ * #Model
+ * @category Model
+ * Component for data/model behavior.
+ * @trigger DataChange - When any data on the model has changed.
+ * @trigger DataChange[key] - When the specific key on the model has changed.
+ * @trigger DataChange[key.key] - The nested key value has changed.
+ * @example
+ * ~~~
+ * Crafty.c('Person', {
+ *   defaults: {name: 'Fox'}
+ * });
+ * person = Crafty.m('Person').setup({name: 'blainesch'});
+ * ~~~
+ */
+Crafty.c('Model', {
+  init: function() {
+    this.defaults = this.defaults || {};
+    this.attributes = this.extend.call(this.attributes || {}, this.defaults);
+    this.original = this.attributes;
+    this.changed = this.changed || [];
+
+    this.bind('DataChange', this._changed_attributes);
+    this.bind('DataChange', this._changed_triggers);
+  },
+
+  /**@
+   * #.setup
+   * @comp Model
+   * Sets up the current model without triggering events.
+   *
+   * @example
+   * ~~~
+   * person.setup({foo: 1, bar: {baz: 2}});
+   * ~~~
+   */
+  setup: function(data) {
+    this.set.apply(this, data, {
+      silent: true
+    });
+    return this;
+  },
+
+  /**
+   * Fires more specific `DataChange` events.
+   *
+   * For instance a `DataChange[name]` may get fired when you
+   * update the name data attribute on the model.
+   */
+  _changed_triggers: function(data, options) {
+    var key, trigger_data;
+    options = Crafty.extend.call({pre: ''}, options);
+    for (key in data) {
+      this.trigger('DataChange[' + options.pre + key + ']', data[key]);
+      if (data[key].constructor.name === 'Object') {
+        this._changed_triggers(data[key], {
+          pre: options.pre + key + '.'
+        });
+      }
+    }
+  },
+
+  /**
+   * Pushes all top-levle changed attribute names to the
+   * changed array.
+   */
+  _changed_attributes: function(data) {
+    var key;
+    for (key in data) {
+      this.changed.push(key);
+    }
+    return this;
+  },
+
+  /**@
+   * #.data
+   * @comp Model
+   * Delegates to the appropriate get/set method.
+   *
+   * @example
+   * ~~~
+   * person.data('name', 'Foxxy');
+   * person.data('name'); // 'Foxxy'
+   * ~~~
+   *
+   * @see .get
+   * @see .set
+   */
+  data: function() {
+    if (arguments.length === 1 && typeof arguments[0] === 'string') {
+      return this.get.apply(this, arguments);
+    } else {
+      return this.set.apply(this, arguments);
+    }
+  },
+
+  /**@
+   * #.get
+   * @comp Model
+   * Getter method for data on the model.
+   *
+   * @example
+   * ~~~
+   * person.get('name'); // Foxxy
+   * person.get('contact'); // {email: 'fox@example.com'}
+   * person.get('contact.email'); // fox@example.com
+   * ~~~
+   */
+  get: function(key, context) {
+    var first, keys, subkey;
+    if (typeof context === "undefined" || context === null) {
+      context = this.attributes;
+    }
+    if (key.indexOf('.') > -1) {
+      keys = key.split('.');
+      first = keys.shift();
+      subkey = keys.join('.');
+      return this.get(keys.join('.'), context[first]);
+    } else {
+      return context[key] || null;
+    }
+  },
+
+  /**@
+   * #.set
+   * @comp Model
+   * Main setter function for data attributes.
+   *
+   * Options:
+   *
+   * `silent`: If you want to prevent it from firing events.
+   *
+   * `recursive`: If you pass in an object you could overwrite
+   * sibling keys, this recursively merges instead of just
+   * merging it. This is `false` by default, unless you are
+   * using dot notation `name.first`.
+   *
+   * @example
+   * ~~~
+   * person.set('name', 'Foxxy', {silent: true});
+   * person.set('name', 'Foxxy');
+   * person.set({name: 'Foxxy'}, {silent: true});
+   * person.set({name: 'Foxxy'});
+   * person.set('name.first', 'Foxxy');
+   * ~~~
+   */
+  set: function() {
+    var args = this._set_normalize_arguments.apply(this, arguments),
+      options = args.options,
+      data = args.data;
+
+    options = Crafty.extend.call({
+      recursive: typeof arguments[0] === 'string' && arguments[0].indexOf('.') > -1,
+      silent: false
+    }, options);
+
+    if (!options.silent) {
+      this.trigger('DataChange', data);
+    }
+
+    if (options.recursive) {
+      this.attributes = this._recursive_extend(data, this.attributes);
+    } else {
+      this.extend.call(this.attributes, data);
+    }
+    return this;
+  },
+
+  /**
+   * Normalizes the arguments for the `set` method.
+   *
+   * The set method accepts a lot of various inputs it's necessary
+   * to move the argument logic out to keep it single-responsibility.
+   */
+  _set_normalize_arguments: function() {
+    var data, length, options;
+    length = arguments.length;
+    if (length === 3 || (length === 2 && typeof arguments[0] === 'string')) {
+      data = this._set_create_object.apply(this, arguments);
+      options = arguments[2] || {};
+    } else {
+      data = arguments[0] || {};
+      options = arguments[1] || {};
+    }
+    return {
+      data: data,
+      options: options,
+    };
+  },
+
+  /**
+   * If you are setting a key of 'foo.bar' or 'bar', this creates
+   * the appropriate object for you to merge with the current
+   * attributes.
+   */
+  _set_create_object: function(key, value) {
+    var data = {}, keys, first, subkey;
+    if (key.indexOf('.') > -1) {
+      keys = key.split('.');
+      first = keys.shift();
+      subkey = keys.join('.');
+      data[first] = this._set_create_object(subkey, value);
+    } else {
+      data[key] = value;
+    }
+    return data;
+  },
+
+  /**
+   * Recursively puts `new_data` into `original_data`.
+   */
+  _recursive_extend: function(new_data, original_data) {
+    var key;
+    for (key in new_data) {
+      if (new_data[key].constructor.name === 'Object') {
+        original_data[key] = this._recursive_extend(new_data[key], original_data[key]);
+      } else {
+        original_data[key] = new_data[key];
+      }
+    }
+    return original_data;
+  },
+
+  /**@
+   * #.is_dirty
+   * @comp Model
+   * Helps determine when data or the entire component is "dirty" or has changed attributes.
+   *
+   * @example
+   * ~~~
+   * person = Crafty.e('DirtyData').setup({name: 'Fox', age: 24})
+   * person.is_dirty() // false
+   * person.is_dirty('name') // false
+   *
+   * person.set('name', 'Lucky');
+   * person.is_dirty(); // true
+   * person.is_dirty('name'); // true
+   * person.is_dirty('age'); // false
+   * person.changed; // ['name']
+   * ~~~
+   */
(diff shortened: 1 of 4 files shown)


starwed:
Hmm, I think you misunderstood my intent in suggesting the `Change[key]` syntax.  The idea was that this would be integrated into the event model -- you wouldn't trigger `"Change[key='val']"`, but rather trigger `"Change"` with a data object that looked like `{key:'val', otherstuff:'etc'}`.  When you bound to an event with that special syntax, Crafty would parse the conditions, and then only trigger the callback when they were met.

starwed:
> `this.trigger('DataChange', {new: data, old: this.attributes});`

One thing to be careful of is creating temporary objects.  This can have unfortunately dire effects in a game framework -- GC pauses kill a game's performance.

I'm not really clear how this would be used, but if you'd anticipate these DataChange events occurring frequently (Either every frame, or on a lot of entities) then you'd probably want to reuse objects instead of creating new ones.  One assumption (which should be documented!) that Crafty makes is that the consumer of an event won't try to hang on to the event data.  That makes it ok to use an object from event to event.

starwed:
The system as a whole looks pretty well written, though!  I'll try to look at in more depth tomorrow, just wanted to post the first couple of things that jumped out at me.

Thanks for the PR!

starwed:
> any thoughts on getting travis setup to automate this?

Unfortunately the repo owner @louisstow hasn't responded to requests to enable it.  (See #540.)

bayleedev:
I fixed an issue with using set and dot notation not setting the recursive flag to true. More details [here](https://github.com/blainesch/Crafty/blob/c16466db5e9302c46a416d9d17a6395ecab9ab68/tests/model.js#L62).

src/model.js
```diff
@@ -0,0 +1,268 @@
...
+   * For instance a `DataChange[name]` may get fired when you
+   * update the name data attribute on the model.
+   */
+  _changed_triggers: function(data, options) {
+    var key, trigger_data;
+    options = Crafty.extend.call({pre: ''}, options);
+    for (key in data.new) {
+      trigger_data = {
+        new: data.new[key],
+        old: data.old[key] || {}
```
kevinsimper: There is no need for passing the old data, because the ones who bind to get updates from the model, already have the old data.

Like a Score model and a text compoenent who shows the score. The text component bind to the score model to get updates, so it can update the text accordingly, but it has already stored the current score from last time it updated the score and just need the new number to update.

src/model.js
```diff
@@ -0,0 +1,268 @@
...
+   * ~~~
+   *
+   * @see .get
+   * @see .set
+   */
+  data: function() {
+    if (arguments.length === 1 && typeof arguments[0] === 'string') {
+      return this.get.apply(this, arguments);
+    } else {
+      return this.get.apply(this, arguments);
```
kevinsimper: `return this.set.apply(this, arguments);` I think this should be set.

src/model.js
bayleedev: I originally did this because of a misunderstanding of @starwed's idea. I assumed I need old and  new data to throw the correct events `ChangeData[foo=bar]` etc. I'll remove it. Thanks [:

src/model.js
bayleedev: Good catch, I'll add a test.

bayleedev: Test lives [here](https://github.com/blainesch/Crafty/blob/4622f5b0f109559db30961e4536f0bbbacde33d4/tests/model.js#L184-L200) thanks!

bayleedev:
Updated.

src/model.js
```diff
@@ -0,0 +1,264 @@
...
+   * Fires more specific `DataChange` events.
+   *
+   * For instance a `DataChange[name]` may get fired when you
+   * update the name data attribute on the model.
+   */
+  _changed_triggers: function(data, options) {
+    var key, trigger_data;
+    options = Crafty.extend.call({pre: ''}, options);
+    for (key in data) {
+      this.trigger('DataChange[' + options.pre + key + ']', data[key]);
```
kevinsimper, line 76: Is there a reason why it is called DataChange and not Change? That would make it much simpler and easier to remember.

kevinsimper:
Nice! :+1: 

When you update your code in a pull request, make a new commit instead of squash them together. Makes it easier to see what you changed since last we commeted :-)

I do think we should stick to the default function called ".attr" for getter/setter to keep nice and clean and avoid confusing, when to use what. Also you could use attr function on a model now and nothing would trigger.

src/model.js
bayleedev, line 76: The `Change` event fires for top level items changing. For instance when I do:

``` coffeescript
paddle.attr(color: 'blue');
paddle.color # blue
```

Compared to the `DataChange` which happens when the attributes of the model change. You can access it like this, although it's preferred to use the `data` or `get` methods.

``` coffeescript
paddle.data(name: 'PC')
paddle.attributes.name # PC
paddle.get('name') # PC
paddle.name # undefined
```

So in theory binding to `Change` and looking for a change to `name` (e.name) would be different than binding to `ChangeData` and looking for `name` (e.attributes.name).

starwed, line 76: Something to keep in mind is that any "Change" triggered by a 2D entity will cause it to be flagged for redraw.  See #482.

kevinsimper, line 76: @blainesch Would that not be confusing for beginners that model attributes is not component attributes?

@starwed Hmm, would that be an idea to prevent that when `{silence: true}` is passed?

starwed, line 76: @kevinsimper  I think I just need to go through and implement the "Invalidate" event to replace the redraw meaning of "Change".  Shouldn't be too difficult.

bayleedev, line 76: @kevinsimper, possibly but they could potentially overwrite something otherwise, right? It seems worth the risk imo. Attributes and data are different enough to need to be stored in different places.

Right now with them:

```
person = Crafty.e('Person, Collision')
person.attr(hit: 0) # Just overwrote `Collision.hit()`
```

With data, you _should_ only use `get`, `set`, and `data` to access them and they can only collide with your own data.

starwed, line 76: Opened PR #694 to rename all the redraw instances of "Change".  (i.e., almost all of them.)

bayleedev, line 76: I'm not completely sold on putting these attributes on the top level. This component adds 3 top level items:

 `attributes` - Current data for the model
`changed` - Which keys have changed
`default` - Default value for the model

We'd be moving `attributes` to the top level, meaning we couldn't give any data the same name as common methods or attributes like `default`, `changed`, `hit`, etc. It feels like it's polluting the top level object.

Because they aren't triggering a change on the attributes it makes sense the trigger name should also be different. `Change` to `DataChange`.

If these changes are required for this to be merged in I'll be happy to update the pr, but just throwing out my 2¢

kevinsimper, line 76: @blainesch It is cool to hear what you think. I am just thinking about those who is starting out with Crafty and have not wrote any code with Crafty, how they would think it would work. I do understand that polluting the top level object is a bad thing and I think you are totally right, but that may not makes sense for beginners and people just making games in their spare time and do not have time to understand what the difference between the .attr function and the set and get function which seems to do the same job.

I like the idea and would really like that you could do something like this, because it would make the game code a lot simpler and very easy to read:

```
var hero = Crafty.e('2D, Canvas, HeroComponent')
    .attr({
        health: 100,
        mana: 100
    })
    .bind('Change[health]', function (health) {
        healthBar.update(health);
        if (health === 0) {
            hero.destroy();
            // Play animation, sound, etc..
        }
    })
    .bind('Change[mana]', function (mana) {
        manabar.update(mana);
        // add data.mana to totalusage
        if (mana === 0) {
            // Alarm the user
        }
    });
```

This makes the code very easy to read and understand instead of now you have to do one big change function with a lot of pseudo code, where you have you to check yourself if health has changed and then do something.

```
var hero = Crafty.e('2D, Canvas, HeroComponent')
    .attr({
        health: 100,
        mana: 200
    })
    .bind('Change', function (data) {
        if (data.health) {
            healthBar.update(data.health);
            if (data.health === 0) {
                hero.destroy();
                // Play animation, sound, etc..
            }
        }
        if (data.mana) {
            manabar.update(data.mana);
            // add data.mana to totalusage
            if (data.mana === 0) {
                // Alarm the user
            }
        }
    });
```

bayleedev, line 76: Yes, I agree the `Change[key]` format is a lot nicer, but that could be an improvement of the `attr` method in just a few lines, right? It seems like something that is separate from this pr.

``` coffeescript
Crafty.component 'ChangeTrigger', 
    init: ->
      @bind 'Change', @triggers
    triggers: (data) ->
      for key of data
        @trigger("Change[#{key}]", data[key])
```

bayleedev:
Any update on getting this merged?

buddylindsey:
I have been investigating craftyjs, and it seems like this would be a good addition. If this isn't merged is there another "standard" way to do this?

starwed:
@buddylindsey If you'd like to use this code _now_, you should be able to pretty much just copy paste the component into your own code (maybe as a separate file). It's an extension to crafty, as opposed to a change to the underlying engine.

> Any update on getting this merged?

Well, it's a big patch!  I think I'd like to cut the next release today, and then this can probably be merged after that.  That'll give it time to be properly tested ~~before~~ after landing.

One remaining meta point I'm not certain about is the `Crafty.m` shorthand. It only saves a tiny amount of typing, and might obfuscate that this really is just another component, and that using `.m` returns something you can treat as a regular entity.

bayleedev:
@starwed that was actually something I was thinking of doing myself. I'm using this component now and just requiring it in the component init.

Thank you for the update. I'll get the `Crafty.m` removed.

bayleedev:
As per @kevinsimper request, this change has not been squished.

mucaho:
Very nice idea! I would like to see that functionality in Crafty. 

As of now the model functionality is added to entities by adding the "Model" component. Would it be useful & possible to add a global model to the Crafty instance itself? (e.g. `Crafty.model(...)`). Every entity could then listen to global model changes.

bayleedev:
Thanks for the support! I'm glad others can find uses for it [:

I'm not sure that's a great idea, since models should also have methods and we have no way of defining those methods on the "global model". It's possible but I think that feature would be more of an implementation on your specific game. For example:

``` coffeescript
Crafty.component 'Board',
  init: ->
    @requires('Model')
  # additional methods

Crafty.extend
  model: Crafty.e('Board')

Crafty.model.on 'PlayerMove', ->
  # foo!
```

src/model.js
```diff
@@ -0,0 +1,252 @@
...
+ * Component for data/model behavior.
+ * @trigger DataChange - When any data on the model has changed.
+ * @trigger DataChange[key] - When the specific key on the model has changed.
+ * @trigger DataChange[key.key] - The nested key value has changed.
+ * @example
+ * ~~~
+ * Crafty.c('Person', {
+ *   defaults: {name: 'Fox'}
+ * });
+ * person = Crafty.m('Person').setup({name: 'blainesch'});
```
kevinsimper: You removed the `.m()` function :)

src/model.js
```diff
@@ -0,0 +1,252 @@
+var Crafty = require('./core.js'),
+    document = window.document;
+
+/**@
+ * #Model
+ * @category Model
+ * Component for data/model behavior.
```
kevinsimper: It should something about that it created and depends on a whole other attributes, that will be accessed through `.setup`, `.get`, `.set`.

Maybe also something about, when to use the model component and why it is a good idea.

"Model is a component you can add to your Component to get a new kind events that is more specific. This makes it possible to only get updates when certain properties change.

src/model.js
```diff
@@ -0,0 +1,252 @@
...
+ * @trigger DataChange - When any data on the model has changed.
+ * @trigger DataChange[key] - When the specific key on the model has changed.
+ * @trigger DataChange[key.key] - The nested key value has changed.
+ * @example
+ * ~~~
+ * Crafty.c('Person', {
+ *   defaults: {name: 'Fox'}
```
kevinsimper, line 27: In the example, if the component already has a variable which is only expected by the model component, it should require the model compent too.

Maybe it should be called `modelDefaults`, because the normal `.attr` would not use it as default.

kevinsimper:
@starwed 

> you wouldn't trigger "Change[key='val']", but rather trigger "Change" with a data object that looked like {key:'val', otherstuff:'etc'}. When you bound to an event with that special syntax, Crafty would parse the conditions, and then only trigger the callback when they were met.

That is because that is how Backbone.js does it. Don't you think it is faster to trigger events with no handlers than to do the parsing every time?

@mucaho 

> As of now the model functionality is added to entities by adding the "Model" component. Would it be useful & possible to add a global model to the Crafty instance itself? (e.g. Crafty.model(...)). Every entity could then listen to global model changes.

It could be easily be done, and it original had a `.m()`, but since it is original a component I think it is best to stay simple.

starwed:
> That is because that is how Backbone.js does it. Don't you think it is faster to trigger events with no handlers than to do the parsing every time?

Well, the parsing would only be necessary on binding.  I wrote up a kind of toy plan for what it might look like; it wasn't as complicated as I'd feared.  For sanity it would probably require abstracting out the common code in Crafty.bind/trigger and e.bind/trigger, but that's likely something that would be good in any case.

Thing is, it might be a useful shorthand in places other than the Model component.

kevinsimper:
I still do not know what to think about the new `this.attributes`, a new set of getter and setters (`.set`, `.get`, `.data`, `.setup`), and a new `DataChange` event.

I thought original that is would hook onto the normal ways of setting values with `.attr` and use the normal `change` event, and then just provide a way to only get events when certain values changed instead of every time _every_ value changed (as it is now with `.bind('change')`).

I think it will make the api inconsistent and because of that people would not use it. You would have to think, "is this value in the normal values or in the special model values?", and then think what is the correct get/set and bind events.

We need something like this indeed!, but I think the code is a middleway now where it is not a core port, but tries to look like one.

Either it should use the normal functions like `.attr` and change event 
or
go the whole other way and make `.model` function where it does not look like the core functions, 
and you do not doubt that.

Both would be really cool! No doubt! Else should it just be a plugin that people can get and add to Crafty as they like.

bayleedev:
> go the whole other way and make .model function where it does not look like the core functions, 
> and you do not doubt that.

Can you give an example, @kevinsimper ?

If we do move the data to the top level, we would lose dirty tracking. We could still have defaults setup on the top level. I could simply listen for "Changes" to throw the custom events.

I'll open an alternative pull request later.

bayleedev:
@kevinsimper @starwed  I updated the model documentation.

bayleedev:
Rebased and bug fix.

starwed:
Closing in favor of #703.

PR #694, opened 2014-01-13
Switch redraw event name to "Invalidate"

### LINKED ISSUE #482: "Change" event overused and overloaded

starwed:
Currently we use "Change" to indicate both when an attribute has been changed by `attr`, and to indicate when an entity needs to be redrawn.
This patch switches the name of the latter type of event to "Invalidate".

Since every other instance of "Invalidate" doesn't pass any data, this no longer passes the old coordinates when an entity is invalidated because of a move -- the event "Move" already covers that case.

This fixes #482

### DIFF
--- a/src/2D.js
+++ b/src/2D.js
@@ -71,7 +71,7 @@ Crafty.extend({
  * @category 2D
  * Component for any entity that has a position on the stage.
  * @trigger Move - when the entity has moved - { _x:Number, _y:Number, _w:Number, _h:Number } - Old position
- * @trigger Change - when the entity has moved - { _x:Number, _y:Number, _w:Number, _h:Number } - Old position
+ * @trigger Invalidate - when the entity needs to be redrawn
  * @trigger Rotate - when the entity is rotated - { cos:Number, sin:Number, deg:Number, rad:Number, o: {x:Number, y:Number}}
  */
 Crafty.c("2D", {
@@ -871,7 +871,7 @@ Crafty.c("2D", {
     /**@
      * #.flip
      * @comp 2D
-     * @trigger Change - when the entity has flipped
+     * @trigger Invalidate - when the entity has flipped
      * @sign public this .flip(String dir)
      * @param dir - Flip direction
      *
@@ -886,7 +886,7 @@ Crafty.c("2D", {
         dir = dir || "X";
         if (!this["_flip" + dir]) {
             this["_flip" + dir] = true;
-            this.trigger("Change");
+            this.trigger("Invalidate");
         }
         return this;
     },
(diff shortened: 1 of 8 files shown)


starwed:
This passes `grunt check`.

The event "Change" doesn't seem to be used anywhere in the test suite, so I don't think any changes are needed there.

sbyrnes321:
This might be a stupid question, but there's something I never quite understood in the drawing system source code. If a sprite is moved (instantaneously) across the screen from (x1, y1) to (x2, y2), the drawing engine needs to know that BOTH of those areas need to be redrawn in the next frame -- both the area around the old location AND the area around the new location. Right? How does it keep track of that?

(I'm not questioning that the code has been working correctly, just wondering how it works.) :-D

starwed:
> This might be a stupid question, but there's something I never quite understood in the drawing system source code. If a sprite is moved (instantaneously) across the screen from (x1, y1) to (x2, y2), the drawing engine needs to know that BOTH of those areas need to be redrawn in the next frame -- both the area around the old location AND the area around the new location. Right? How does it keep track of that?

When the canvas is rendered, Crafty first calculates all the areas of the screen that need to be redrawn.   The first step is to collect an array of rectangles that have changed.  For each entity that's changed, it adds both their old position and their new one to the array.  

It does some minimal optimization of merging overlapping areas together, though we could probably be smarter about that.

Then for each dirty region, it clips to that region, finds all the entities that overlap with that area, sorts them by z index, and then draws them in that order.

**After an entity is drawn to the canvas,  the coordinates of its MBR are stored in an entity property called `.staleRect`, ready to be used the next time that entity changes.**

You can see the parts of the code that handle the old areas in the functions [createDirty](https://github.com/craftyjs/Crafty/blob/develop/src/drawing.js#L302) and [clean](https://github.com/craftyjs/Crafty/blob/develop/src/drawing.js#L281).

kevinsimper:
:+1: 

Also nice explanation, that should be saved somewhere.

PR #695, opened 2014-01-13
Fix shared coord issue

starwed:
The recent changes (#665) to how "Sprite" entities  are generated accidentally caused them to share a `__coord` array. 

This one-line patch fixes the issue.

I was going to add a testcase for this, but we don't seem to have any "Sprite" tests at all?  We should fix that, but probably no point in waiting on that for this particular PR.  :)

### DIFF
--- a/src/extensions.js
+++ b/src/extensions.js
@@ -236,6 +236,7 @@ Crafty.extend({
             this.requires("2D, Sprite");
             this.__trim = [0, 0, 0, 0];
             this.__image = url;
+            this.__coord = [this.__coord[0], this.__coord[1], this.__coord[2], this.__coord[3]];
             this.__tile = tile;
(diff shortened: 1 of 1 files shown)


kevinsimper:
@starwed No we do not have any sprite tests, other than the sprite animation tests.

Why is the Sprite function not in the sprite.js file? That were the first place I tried looked at it.

But yes, that would be nice with test for the sprite function, it would also make it easier to understand.

starwed:
> Why is the Sprite function not in the sprite.js file? That were the first place I tried looked at it.

Yes, every time I look at that, I think it should be moved!  Just haven't gotten around to it yet.

(I think at some point in the past every use of Crafty.extend was confined to extensions.js. )

PR #698, opened 2014-01-16
Multiway speed docs

kevinsimper:
Someone at the forum asked if you could change the speed of multiway and saw that there were no docs for that function.

Should there be way to get the current speed? Other than accessing the private variable this._speed?

### DIFF
--- a/src/controls.js
+++ b/src/controls.js
@@ -812,6 +812,19 @@ Crafty.c("Multiway", {
         return this;
     },
 
+    /**@
+     * #.speed
+     * @comp Multiway
+     * @sign public this .speed(Number speed)
+     * @param speed - The speed the entity has.
+     *
+     * Change the speed that the entity moves with.
+     *
+     * @example
+     * ~~~
+     * this.speed(2);
+     * ~~~
+     */
     speed: function (speed) {
         for (var k in this._keyDirection) {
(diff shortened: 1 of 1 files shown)


src/controls.js
```diff
@@ -812,6 +812,19 @@ Crafty.c("Multiway", {
...
+     * Change the speed that the entity moves with.
+     *
+     * @example
+     * ~~~
+     * this.speed(2);
+     * ~~~
+     */
     speed: function (speed) {
         for (var k in this._keyDirection) {
             var keyCode = Crafty.keys[k] || k;
```
bayleedev, line 830: Unrelated to the PR, but is there a reason we are using `var` in the loop instead of at the top of the method?

starwed, line 830: Possibly someone didn't like the idea of an extra line?  It doesn't _really_ make a difference when it's executed, but I'm a bit surprised jshint doesn't complain about it.  It does in some similar but more bug-prone cases, so maybe this particular pattern isn't considered harmful.

bayleedev, line 830: I was always under the impression it was a performance gain because you aren't redeclaring it multiple times. However this test proves otherwise.

[JSHint FTW](http://jsperf.com/declaring-variables-inside-loops/6)

Issue #147, opened 2011-11-26
Collision Events get implicitly cancelled


robkuz:
When a collision event is created and one of the listeners to that event
reverts the condition that lead to that collision event in the first place
the event gets silently cancelled out for all following listeners (callbacks).

This leads to the effect that one has to take care in which order listeners
to onHit are added and what those listeners do. At the end it makes for
very brittle code.

please see the fiddle over here http://jsfiddle.net/robertj/przxN/7/

Only if the order of the two onHit callbacks is reversed the debug message
is ever printed.

grozen:
I am so on this.

starwed:
This isn't really configured as a Crafty event in the first place -- and that would probably be the best way to fix it.

Currently `onHit` sets up a function that, on each "EnterFrame"  watches for hit events, and uses the callbacks when the hit state changes.  Each call to `onHit` sets up a new watcher.

If the watcher instead triggered  "OnHit" and "OnHitOff" events (or whatever it makes sense to call them), then any number of handlers bound to those events would get called.  Of course, these might then need to handle the "cancelled collision" case themselves.

grozen:
That actually sounds considerably better than what I had in mind (consolidate all the individual callbacks, and call them as a unit).

Not very backwards compatible, but you know that I favor improvement over preservation. I do believe your solution would also fix this bug (all registered handlers will be called once the collision takes place, regardless of what side effects they cause).

However, like many aspects in Crafty collision could use some tests, so I set out to do that as well while I'm at it. Thing is, I tried setting up three entities that have the `Collision` and `SolidHitBox` components for said tests, only to end up with the following flicker-fest: http://jsfiddle.net/mCdUX/145/

Am I doing something wrong, or is this yet another bug? (Which I will gladly squish along with this, if it doesn't prove too much of a mess)

**EDIT:** I must not have saved my fiddle properly... but also the issue vanished after rebuilding Crafty. Must have been using an old version. Onward with solving this, then.

grozen:
@starwed can we consider this issue closed? This bug was actually what caused me to introduce the event-based hit detection, following your sage advice.

Of course the problem persists if you keep using the old style `OnHit`, but using events instead solves the problem, as I demonstrated in the fiddle I linked to in my pull request.

mucaho:
closed this, as it has been resolved with #699 .

PR #699, opened 2014-01-16
Collision events

grozen:
This pull request exposes an alternative method for handling collision by binding to events that fire as collisions take place. The events in question are "HitOn" for when collisions occur and "HitOff" for when they cease.

This pull request was born as an effort to solve issue #147, which it does (as you can see in the following fiddle: http://jsfiddle.net/nm2nL/5/).

I've added some tests as well, and will more than welcome your scrutiny.

### DIFF
--- a/src/collision.js
+++ b/src/collision.js
@@ -5,18 +5,38 @@ var Crafty = require('./core.js'),
 /**@
  * #Collision
  * @category 2D
+ * @trigger HitOn - Triggered when collisions occur. Will not trigger again until collisions of this type cease, or an event is requested once more (using `resetHitChecks(component)`). - { hitData }
+ * @trigger HitOff - Triggered when collision with a specific component type ceases - { componentName }
+ *
  * Component to detect collision between any two convex polygons.
+ *
+ * If collision checks are registered for multiple component and collisions with
+ * multiple types occur simultaniously, each collision will cause an individual
+ * event to fire.
+ *
+ * **Note:** All data received from events is only valid for the duration of the event's callback.
+ * If you wish to preserve the data, make a copy of it.
+ *
+ * For a description of collision event data (hitData above), see the documentation for
+ * `.hit()`.
+ *
+ * @see .hit, Crafty.init, .resetHitChecks
  */
 Crafty.c("Collision", {
     /**@
      * #.init
      * @comp Collision
-     * Create a rectangle polygon based on the x, y, w, h dimensions.
+     * Set up collision handling.
      *
      * By default, the collision hitbox will match the dimensions (x, y, w, h) and rotation of the object.
+     *
+     * **Note:** If the entity this component is applied to does not have its
+     * dimensions set the default hit area will not be set properly.
      */
     init: function () {
         this.requires("2D");
+        this._collisionData = {};
+
         this.collision();
     },
 
@@ -35,17 +55,15 @@ Crafty.c("Collision", {
      * @trigger NewHitbox - when a new hitbox is assigned - Crafty.polygon
      *
      * @sign public this .collision([Crafty.polygon polygon])
-     * @param polygon - Crafty.polygon object that will act as the hit area
+     * @param polygon - Crafty.polygon object that will act as the hit area.
      *
      * @sign public this .collision(Array point1, .., Array pointN)
-     * @param point# - Array with an `x` and `y` position to generate a polygon
+     * @param point# - Array of [x, y] coordinate pairs to generate a hit area polygon.
      *
-     * Constructor takes a polygon or array of points to use as the hit area.
+     * Constructor that takes a polygon or array of points to use as the hit area,
+     * with points being relative to the object's position in its unrotated state.
      *
-     * The hit area (polygon) must be a convex shape and not concave
-     * for the collision detection to work.
-     *
-     * Points are relative to the object's position and its unrotated state.
+     * The hit area must be a convex shape and not concave for collision detection to work properly.
      *
      * If no parameter is passed, the x, y, w, h properties of the entity will be used, and the hitbox will be resized when the entity is.
      *
@@ -67,7 +85,7 @@ Crafty.c("Collision", {
         this.unbind("Resize", this._resizeMap);
         this.unbind("Resize", this._checkBounds);
 
-        
+
 
         if (!poly) {
             // If no polygon is specified, then a polygon is created that matches the bounds of the entity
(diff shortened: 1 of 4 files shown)


starwed:
I've only glanced at it so far, but I think you're checking for hits with "EnterFrame"?

There's no guarantee that these events happen in a particular order, so it might be that an object moves _after_ the collision check occurs.  So if you try to use this to prevent overlaps, for instance, it might fail.  It would only lag by a single frame at worst, and so not be very noticeable to the eye -- but if you want to treat "no overlaps" as an invariant it might cause issues.

One solution would be to bind to "Move" and "Rotation" events instead of "EnterFrame".  But these could conceivably trigger multiple times in a physics frame, and thus be more expensive to calculate.   (I've noticed that in practice, the collision detection code is the most CPU intensive thing that happens outside of rendering.)  And they wouldn't work when another entity move on top of _you_.

I'm not really sure what the best approach is.  Maybe the best use of `onHit` isn't to prevent collisions, but just to detect when you pass over powerups, switches, and the like, in which case it doesn't really matter if it might occasionally lag by a single frame.

starwed:
It's possible that a long term solution is to have code in the hashmap implementation that deals with this when an entities position updates.  It already does half the work in any case; it wouldn't be too crazy for it to maintain the collision status of every "Collision" element.

grozen:
I am indeed checking for hits on _EnterFrame_, @starwed, but this is no different than what "onHit" currently does.
While it does not make your point less true, the real question is whether or not it is actually a problem. If you check out the fiddle I made then everything seems in order, as in, you'd never even suspect that the problem you describe exists. Then again, it might be pure luck.

Also, is your second comment more of a note to yourself? I am afraid I did not follow it, exactly.

starwed:
> I am indeed checking for hits on EnterFrame, @starwed, but this is no different than what "onHit" currently does.

That's true.  As long as it's documented in some way, sounds good to me!

> Also, is your second comment more of a note to yourself? I am afraid I did not follow it, exactly. 

Sorry, just thinking out loud.  As an aid for collision detection and rendering, Crafty keeps track of the location of 2D entities in course-grained spatial index.  Every time an entity changes positions, we update it's position in that index; if we also wanted to update the collision status of entities every time they moved, that would be the obvious place to put such code.

grozen:
Thanks for the clarification. I've gone ahead and enhanced the documentation.

src/collision.js
```diff
@@ -313,6 +336,196 @@ Crafty.c("Collision", {
...
     },
 
+
+    /**@
+     * #._createCollisionHandler
+     * @comp Collision
+     * @sign public void .checkHits(String component, Object collisionData)
+     * @param component - The name of the component for which this handler
+     * checks for collisions.
+     * @param component# - Collision data object used to track collisions with
```
starwed: Should be collisionData the second time?

starwed:
Ok, I've gone through and the API seems sensible to me.  I haven't tested it in action yet, though!

Two things:
- Could you squash down all the commits at some point before we land it?  There are a _lot_.  :)
- The existing implementation of `hit` is inefficient if you care about multiple components.  It fetches a list of all potential collisions, then discards any potential matches that don't have the right component.  That means if you call it multiple times per frame with the same component, it'll repeat that first stage multiple times.  That said, it won't waste time on the more time consuming sort of collision check, so maybe that doesn't really matter.

Anyway, once the commits are squashed down, this can go ahead and land.  It'll probably get more of the sort of testing it needs once it's  actually in the development branch.  :)

grozen:
Neat. Keep in mind that like my previous change the test page contains a playground that demonstrates the feature. However, it displays some very odd behavior as all the draggable component refuse to get dragged until the window is resized... (I am at a loss as to why this happens, at the moment. The problem is even more severe on chrome).

I will be sure to squash my commits before long.

Regarding the inefficiency of 'hit', I suspect this is outside the scope of this change, not that I would mind tackling this independently. Specifically, I suspect the way to go here is to have hit receive multiple components so that it can find all the interesting collisions in one pass.

grozen:
Ok, I **think** I've performed a proper squash. I consider rebasing to be 50% dark magic.

starwed:
Oh hmm, wasn't thinking about the tests.    I think sometime after you opened this PR, we switched to calling them as .js files for everything except the animation tests.  For the automated tests, does this need to be a standalone html file, or can it just be called from `index.html` like the others are now?

I do like the idea of the playground, though -- they're a nice addition to the automated tests.

grozen:
All the tests need to pass is a Crafty stage, so they aren't really tied to the playground in any way. Still, I consider the playgrounds a complementary part to the tests I produce, as they allow you to get a more tangible sense of the feature being tested.

Not to mention they serve as an example of just how the feature is used outside of a test.

Seeing how we can list individual test files to be run in the `Gruntfile` instead of listing them all under index.html, is this such a bad thing?

I'd hate to be a troublemaker, obviously, and if need be I'll move the tests.

src/collision.js
```diff
@@ -206,32 +221,40 @@ Crafty.c("Collision", {
...
      * }]
      * ~~~
-     * `MBR` is your standard axis aligned rectangle intersection (`.intersect` in the 2D component).
-     * `SAT` is collision between any convex polygon.
+     * **obj:** The entity with which the collision occured.
+     * **type:** Collision detection method used. One of:
+     * + *MBR:* Standard axis aligned rectangle intersection (`.intersect` in the 2D component).
+     * + *SAT:* Collision between any two convex polygons.
+     * **overlap:** If SAT collision was used, this will signify the overlap percentage between the colliding entities.
      *
```
mucaho, line 256: Could you add a line here saying: "`SAT` collision can only occur if both colliders have the `Collision` component."?

grozen:
@mucaho That's a very good point. Added.

starwed:
@grozen the only issue with having multiples files is that it takes longer to run them, since apparently a big chunk of the time is spent setting up the framework.  (On my machine each additional test html file adds ~5s of time to `grunt check`, which obviously adds up with a lot of tests.)

Since everything else is good I'll just go ahead and merge this, and deal with the tests in a followup PR.  :)

grozen:
@starwed How very surprising. I would not expect such a performance penalty for merely having additional test files.

I'll do my best to detach my tests from their respective playgrounds without sacrificing the playgrounds in the process.

starwed:
@grozen  Apparently running the tests happens pretty quickly, and most of the time is spent by grunt setting up the testing framework separately for each file?

Having a separate directory for playgrounds and demo-like files would be useful.  In general, some behavior is easier to check visually than through unit testing, so it would be nice to do so before official releases.  :)

PR #700, opened 2014-01-20
add example to Crafty.background documentation

sbyrnes321:


starwed:
Simple improvement to documentation, so merging right away.

Thanks!

PR #701, opened 2014-01-20
fix bug in Crafty.DOM.translate

sbyrnes321:
Arithmatic error: The translation calculation was incorrect when the viewport scale is not 1 AND the viewport x or y is not 0. I fixed it. (I also improved the documentation.)

Example:

http://jsfiddle.net/B53CY/ -- Before this change

http://jsfiddle.net/B53CY/1/ -- After this change

### DIFF
--- a/tests/stage.js
+++ b/tests/stage.js
@@ -243,6 +243,11 @@ test("DOMtranslate", function() {
   craftyxy = Crafty.DOM.translate(clientX, clientY);
   strictEqual(craftyxy.x, -Crafty.viewport._x + (Crafty.viewport._width / Crafty.viewport._scale));
   strictEqual(craftyxy.y, -Crafty.viewport._y + (Crafty.viewport._height / Crafty.viewport._scale));
+  
+  // clean up
+  Crafty.viewport.scale(1);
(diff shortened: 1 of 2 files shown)


kevinsimper:
Nice :+1: 

Is it something we could write a test for?

sbyrnes321:
@kevinsimper  -- I have added a test

tests/stage.js
```diff
@@ -220,6 +220,32 @@ test("centerOn", function() {
   Crafty.viewport.scroll('y', 0);
 });
 
+test("DOMtranslate", function() {
+  var clientX, clientY, craftyxy;
+
+  Crafty.viewport.scale(1.7);
+  Crafty.viewport.x = 38;
+  Crafty.viewport.y = 94;
```
kevinsimper, line 228: You should set them back to normal at the bottom of the test as changed values will effect the other tests, since Crafty does not reset for each test.

sbyrnes321:
@kevinsimper -- Gotcha, thanks. I added a new commit to make the test clean up after itself.

starwed:
Looks good, thanks!

PR #702, opened 2014-01-21
viewport improvements

sbyrnes321:
- Get rid of `Crafty.viewport._zoom` which is synonymous with `Crafty.viewport._scale`. There is no point in having two properties that are always equal to each other.
- Eliminate temporary-object-creation in Crafty.viewport.rect(), by making a persistent object. (To avoid unwanted interactions, this object should be read-only, which required me to edit diamondiso.js.)
- Improve documentation

### DIFF
--- a/src/diamondiso.js
+++ b/src/diamondiso.js
@@ -99,21 +99,25 @@ Crafty.extend({
         area: function (offset) {
             if (!offset) offset = 0;
             //calculate the corners
-            var vp = Crafty.viewport.rect();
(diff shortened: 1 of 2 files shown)


starwed:
I think I started to rename the variable because of the confusion with `viewport.zoom` and `viewport.scale`, which did different things, but never finished the job.

At any rate, the `viewport.scale` method explicitly sets the `_scale` property, so lets definitely get rid of zoom!

Also, nice improvements to the documentation.

PR #703, opened 2014-01-23
Create the Crafty model component with core `attr` updates.

bayleedev:
This is an alternative to #693

The main differences are:
- Saves data to the root of the component.
- Triggers `Change[*]` events, compared to `ChangeData` events.
- Updates on `Crafty.attr` to work with recursive setting/getting. BACKWARDS COMPATIBLE! [:

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,8 @@
     "src/math.js",
     "src/time.js",
     "src/DebugLayer.js",
-    "src/keycodes.js"
+    "src/keycodes.js",
+    "src/model.js"
   ],
   "dependencies": {
     "grunt": "~0.4.1",
--- a/src/core.js
+++ b/src/core.js
@@ -419,6 +419,8 @@ Crafty.fn = Crafty.prototype = {
      * @sign public this .attr(String property, * value)
      * @param property - Property of the entity to modify
      * @param value - Value to set the property to
+     * @param silent - If you would like to supress events
+     * @param recursive - If you would like merge recursively
      * @sign public this .attr(Object map)
      * @param map - Object where the key is the property to modify and the value as the property value
      * @trigger Change - when properties change - {key: value}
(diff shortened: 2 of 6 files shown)


starwed:
Sorry for taking a while to look at this.

I don't really understand the use case of `isDirty`.  Is it ever set back to clean again?  Or does it only track whether it has changed ever?

bayleedev:
Right now it simply just sets then to dirty upon change. Ideally once you "save" your model back to the server it'll change then back to "clean". I never added a "save" method because making cross browser Ajax requests is a little above my head lol

Because the data is on the top level, I'm not entirely sure how we would knew what to send to the server, unless only changed data. I guess every attribute that can be serialized or isn't a function. We can think about that when we get there.

bayleedev:
The problem with the `blacklist` is that other component might have other top level attributes. So sending all the correct data fields might get messy. In the example below I successfully got `name` in the console and nothing else. This would definitely be the biggest con of this pr compared to the other.

``` javascript
Crafty.c('Person', {
    name: 'blainesch',
    init: function() {
        this.requires('Model');
    }
});
var blaine = Crafty.e('Person');
var value, key;
for (key in blaine) {
    value = blaine[key];
    // Starts with underscore / protected attributes
    if (key.match(/^_/)) {
        continue;
    }
    // Functions
    if (value.constructor.name === 'Function') {
        continue;
    }
    // Blacklist
    if (key.match(/^(0|length|changed)/)) {
        continue;
    }
    console.log('Good key: ', key);
}
```

starwed:
I feel like I can't really judge whether this PR or the other provides better support for Model type behavior -- I've never used them, so don't know what sort of problems/traps/etc can occur.

Anyway, some general notes/questions:
- `_get` and `_set` might be slightly too generic; someone might end up using those methods on their own components.  You could either qualify them or add an extra underscore?
-  I'm not sure I like that setting a value through the form `attr("keyname", value)` creates a temporary object.  My first contribution to the project was to go through and kill object creation everywhere I could, and it actually made a surprising difference in game behavior.  (GC pauses are kind of the ultimate enemy of browser games right now.)  I guess a solution would be to have a pool of reusable objects (like we now do in 2D.js), but that might get a little complicated.
- Same with `_set` creating an options object every time it's called.  Options objects are a really nice syntax, but unfortunately something we should probably avoid in Crafty.  :(
- I don't quite follow why all the functions are being called through `apply` -- why is that necessary?
- Docs use `.get`, actual function name is `._get`?  (Same for set.)

bayleedev:
@starwed 
1) How about "namespacing" them? l could update them to be `_attr_get` and `_attr_set`.
2) This is how the current [attr](https://github.com/blainesch/Crafty/blob/897e1131aba0b27cee52ce068eadc96a7a3f388f/src/core.js#L453-L455) works now. So it's not any worse.
3) I'm not sure how to get around not using `options`, I assume we don't want a ton of extra params and end up like [php](http://www.php.net/manual/en/function.imageellipse.php), and I also assume we don't want to remove those options since both seem necessary. Please advise on what's necessary to to get this merged.
4) You're right, I don't need to use `apply` on all of them. I'll update accordingly.

starwed:
@blainesch 

> How about "namespacing" them? l could update them to be _attr_get and _attr_set.

That seems fine to me; however you want. 

> This is how the current attr works now. So it's not any worse.
> Well, that's true.  :)  It kind of defeats the point of calling `attr` without an object as an argument, so the docs should mention it if they don't already.
> 
> I'm not sure how to get around not using options
> If silent and recursive are the only options, I think we just bite the bullet and have both as parameters.  `attr` is the sort of method that might get called multiple times _per frame_, so we have to balance the affordances of the API with performance concerns.

(In general, another option is to change the state of the entity like this:

```
e.silent(true).attr({x: 300}).silent(false).attr({y:600});
```

But that doesn't seem like a natural fit here.)

I guess they could also be special-cased properties of the attribute object that's passed in?  Seems like that could be a little clumsy, but something like `.$silent` and `.$recursive` are pretty unlikely to be already in use.

bayleedev:
Fixed @starwed. I didn't squash them since it was mentioned not to do that in the alternative pull request, but let me know when you have  reviewed the changes and I'll  squash this up for you.

kevinsimper:
Hi, back from ski holiday again :)

@blainesch This looks really great! Simple and easy to use and i would use this.
Nice with the tests too.

bayleedev:
Check it out, I have the travis green flag :smiley:

If this is all good should I squash and get merged? Or are we waiting on something still?

starwed:
> If this is all good should I squash and get merged? Or are we waiting on something still?

The technical side of it looks fine to me; it doesn't look like it'll impact anyone who uses `attr` without caring about the Model stuff. Like I said earlier, I don't feel like I know enough about how the system is supposed to work to provide feedback on that part, though.

@kevinsimper: you seem to be more familiar with this stuff, if it looks good to you go ahead and merge.  :)

bayleedev:
Squashed this bad boy.

starwed:
I'll take kevin's silence to mean he has no problems, I guess?  :)  So lets merge this.

starwed:
I suspect there'll be a couple of follow-up PRs -- as a random example, the docs don't explain the silent and recursive parameters.  But now that it's landed, anyone can submit a quick PR instead of us leaning on @blainesch.  :)

bayleedev:
w00t

kevinsimper:
@starwed Thanks, yes! You are awesome! :+1:

PR #705, opened 2014-01-27
Resolve #552: events stick under certain conditions

### LINKED ISSUE #552: fourway() event gets stuck ?
When I test the [Getting started](http://buildnewgames.com/introduction-to-crafty/) code ([demo here](http://buildnewgames.com/assets/article//introduction-to-crafty/tut_bng/index.html)), and hold CTRL + UP_ARROW + E_KEY , the character keeps moving upwards infinitely.

Notes:
- Getting started guide uses crafty v0.5.3
- I have tested on FF 24 and chrome 29.0 on a Windows machine
- Same behavior observed when i press WINDOWS_KEY + (any_arrow).
- Similar behavior observed when i press CTRL + (any arrow key) + (A or S or W or E or D or Q)
- Same behavior observed when I test with local code that uses v0.5.4 

Not sure what's going on but if someone points out what might be causing the problem I'm happy to investigate and submit a patch

ashes999:
Note: this only works with the following HTML mark-up:

`<body onblur='Crafty.resetKeyDown();'>`

This is despite the DOM event bindings for `blur`. Please provide feedback on how to fix this (it should work, but doesn't).

### DIFF
--- a/src/controls.js
+++ b/src/controls.js
@@ -27,18 +27,34 @@ Crafty.extend({
      * ~~~
      * @see Keyboard, Crafty.keys
      */
-
     detectBlur: function (e) {
         var selected = ((e.clientX > Crafty.stage.x && e.clientX < Crafty.stage.x + Crafty.viewport.width) &&
             (e.clientY > Crafty.stage.y && e.clientY < Crafty.stage.y + Crafty.viewport.height));
 
-        if (!Crafty.selected && selected)
+        if (!Crafty.selected && selected) {
             Crafty.trigger("CraftyFocus");
-        if (Crafty.selected && !selected)
+        }
+        
+        if (Crafty.selected && !selected) {
             Crafty.trigger("CraftyBlur");
-
+        }
+        
         Crafty.selected = selected;
     },
+    
+    resetKeyDown: function() {
+        // Tell all the keys they're no longer held down
+        for (var k in Crafty.keys) {
+             if (Crafty.keydown[Crafty.keys[k]]) {
+                 this.trigger("KeyUp", {
+                     key: Crafty.keys[k]
+                 });
+             }
+        }
+		
+        Crafty.keydown = {};
+    },
+    
     /**@
      * #Crafty.mouseDispatch
      * @category Input
@@ -56,11 +72,11 @@ Crafty.extend({
      * Notable properties of a MouseEvent e:
      * ~~~
      * //(x,y) coordinates of mouse event in web browser screen space
-     * e.clientX, e.clientY	
+     * e.clientX, e.clientY
      * //(x,y) coordinates of mouse event in world/viewport space
-     * e.realX, e.realY		
+     * e.realX, e.realY
      * // Normalized mouse button according to Crafty.mouseButtons
-     * e.mouseButton			
+     * e.mouseButton
      * ~~~
      * @see Crafty.touchDispatch
      */
@@ -336,6 +352,7 @@ Crafty.bind("Load", function () {
     Crafty.addEvent(this, Crafty.stage.elem, "mousedown", Crafty.mouseDispatch);
     Crafty.addEvent(this, Crafty.stage.elem, "mouseup", Crafty.mouseDispatch);
     Crafty.addEvent(this, document.body, "mouseup", Crafty.detectBlur);
+    Crafty.addEvent(this, window, "blur", Crafty.resetKeyDown);
     Crafty.addEvent(this, Crafty.stage.elem, "mousemove", Crafty.mouseDispatch);
     Crafty.addEvent(this, Crafty.stage.elem, "click", Crafty.mouseDispatch);
     Crafty.addEvent(this, Crafty.stage.elem, "dblclick", Crafty.mouseDispatch);
@@ -366,6 +383,7 @@ Crafty.bind("CraftyStop", function () {
     }
 
     Crafty.removeEvent(this, document.body, "mouseup", Crafty.detectBlur);
+    Crafty.removeEvent(this, window, "blur", Crafty.resetKeyDown);
 });
 
 /**@


src/controls.js
```diff
@@ -27,18 +27,39 @@ Crafty.extend({
...
+             }
+        }
+        Crafty.keydown = [];
+        
+        // Stop moving
+        this._movement = {x: 0, y: 0};
+                
+        // Trigger direction change event
+        this.trigger('NewDirection', this._movement);   
+    },
```
starwed, line 56: Looks like there's some leftover code from the multiway fix here.  That should all be taken care of by triggering the keyup events.

src/controls.js
```diff
@@ -27,18 +27,39 @@ Crafty.extend({
...
+    resetKeyDown: function() {            
+        // Tell all the keys they're no longer held down
+        for (var k in Crafty.keys) {
+             if (Crafty.keydown[Crafty.keys[k]]) {
+                 this.trigger("KeyUp", {
+                     key: Crafty.keys[k]
+                 });
+             }
+        }
+        Crafty.keydown = [];
```
starwed: Use `Crafty.keydown.length=0` to quickly wipe an array without creating a new one.

starwed:
> This is despite the DOM event bindings for blur. Please provide feedback on how to fix this (it should work, but doesn't).

You probably want to use `window` rather than `document.body` when binding that event, although I haven't gone in and tested that!  :)

The basic idea here looks good!    Some comments:
- If a user holds down a key while the window loses focus, and then lets up when focus is returned, it'll cause the opposite problem.  So in `keyboardDispatch` we need to not fire a keyup event if Crafty doesn't know about the key being down.  We already do the equivalent check for keydown.
- I think you added some extraneous spaces at some line endings.  If you've installed the grunt toolchain, run `grunt check` and it'll lint the code -- that'll tell you which lines need fixing up.  (And also, run unit tests.  To just run the linter by itself, use `grunt jshint`.)
- If you can, please squash the commits down so that the revision history is cleaner.  ([Instructions here](https://github.com/craftyjs/Crafty/wiki/Rewriting-History) if you're not sure how.)

Thanks for the PR!

src/controls.js
ashes999: It's interesting. Changing this to `.keydown.length = 0` creates a side-effect: the first case of `up+ctrl+e` halts correctly, but it remembers the state of the up key being held down, and subsequent calls still make the fourway stick.

I'll have to leave it as-is for now.

src/controls.js
ashes999, line 56: You're right. I'll integrate this change.

ashes999:
I can't get `Crafty.addEvent` to to work; neither `document.body` nor `window` actually bind the event. I'm pretty sure I want to bind to `document.body`, because I want the equivalent of `<body onblur="...">`.

Can you please investigate this? It seems like another bug.

starwed:
> I can't get Crafty.addEvent to to work; neither document.body nor window actually bind the event. I'm pretty sure I want to bind to document.body, because I want the equivalent of <body onblur="...">.

I checked out your branch -- changing `document.body` to `window` when adding the blur event worked fine for me in Firefox and Chrome.  Maybe doublecheck that it wasn't working?  For instance, maybe your browser was caching the old version when you tested it.   (As I understand it, the standard is that the window fires a blur event, and that's what we actually care about -- as long as the window is focused, it's catching the keyup/down events.)

> It's interesting. Changing this to .keydown.length = 0 creates a side-effect: the first case of up+ctrl+e halts correctly, but it remembers the state of the up key being held down, and subsequent calls still make the fourway stick.

Ah, that's because the keydown map starts as an object.  If you change it so that it's initalized as an array, then setting the length to 0 should work fine.  :)

starwed:
Oh, and the reason adding `onblur` to the body element works can be found [in the HTML5 spec](http://www.w3.org/html/wg/drafts/html/master/sections.html#the-body-element):

> The body element exposes as event handler content attributes a number of the event handlers of the Window object. It also mirrors their event handler IDL attributes.
> 
> The **onblur**, onerror, onfocus, onload, onresize, and onscroll event handlers **of the Window object**, exposed on the body element, replace the generic event handlers with the same names normally supported by HTML elements.

In other words, the onblur _attribute_ of `<body>` attaches the event to the window object.  Using `addEventListener` apparently doesn't work quite the same way.   

> A regular event listener attached to the body using addEventListener(), however, would be run when the event bubbled through the body and not when it reaches the Window object:

(I only found this out through this [stackoverflow question](http://stackoverflow.com/questions/6939047/does-html-body-tag-have-an-onblur-event) )

ashes999:
You're right. It seems to work as expected now.

For the object, since it's an object (not an array), reseting it to a new object (`{ }`) should be sufficient.

Please verify. I think we're done here :)

ashes999:
Just to clarify:
- I updated, squashed, and re-pushed
- There is still an occasional weird issue with focus. But it's not related to this (just like `CTRL+H` still breaks `fourway`)

ashes999:
Any comments @starwed ?

starwed:
@ashes999  Nah, I think this looks good.  Just want to merge develop into testing before this lands, since it's a bigger change.

starwed:
We still need to fix the CMD key issue on macs, but this patch is nicely compatible with that.  (We can just call the reset function and do whatever else will happen.)

starwed:
Oh, and thanks for the PR!   :)

ashes999:
Glad to help :)

Two open issues are still:
- CMD key (Macs)
- CTRL+H (Windows at least)

Since #552 is closed, I suggest opening one (or two) new issues to track these two.

Thanks for the code review!

starwed:
I'll just reopen 552.  (Since there's already discussion of the issues there.)

PR #707, opened 2014-01-28
Fix DOM coordinates translation on Chrome 32

potomak:
It happened that on the latest Chrome (32.0.1700.77) both
`document.body.scrollTop` and `document.documentElement.scrollTop` were
valorized, this caused a wrong computation of translated position of mouse
pointer with a missing trigger of a `Click` event on a DOM entity.

Note: I setup a separate test page for this test case to be used only with
_real browsers_ as I found really difficult to handle page scrolling with
phantomjs.

More details about it at https://github.com/ariya/phantomjs/issues/10162
and https://github.com/gruntjs/grunt-contrib-qunit/issues/34

### DIFF
--- a/src/DOM.js
+++ b/src/DOM.js
@@ -394,9 +394,12 @@ Crafty.extend({
          * where the stage is within the screen, what the current viewport is, etc.
          */
         translate: function (clientX, clientY) {
+            var doc = document.documentElement;
+            var body = document.body;
+
             return {
-                x: (clientX - Crafty.stage.x + document.body.scrollLeft + document.documentElement.scrollLeft) / Crafty.viewport._scale - Crafty.viewport._x,
-                y: (clientY - Crafty.stage.y + document.body.scrollTop + document.documentElement.scrollTop) / Crafty.viewport._scale - Crafty.viewport._y
+                x: (clientX - Crafty.stage.x + ( doc && doc.scrollLeft || body && body.scrollLeft || 0 )) / Crafty.viewport._scale - Crafty.viewport._x,
+                y: (clientY - Crafty.stage.y + ( doc && doc.scrollTop  || body && body.scrollTop  || 0 )) / Crafty.viewport._scale - Crafty.viewport._y
             };
         }
     }
--- a/tests/dom_translate.html
+++ b/tests/dom_translate.html
@@ -1,0 +1,33 @@
+<!DOCTYPE html>
+<html class="no-js">
+  <head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+
+    <!-- Note to developers: this test should be run only in real browsers -->
+
+    <!-- DEPENDCIES -->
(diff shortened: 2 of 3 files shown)


starwed:
So we were relying on the fact that previously, only one or the other of `body | documentElement` had a scrollTop attribute?  This fix is needed becuase Chrome now returns values for both?

~~Does it still work in less recent versions of Chrome?~~ Ok, 32 is the stable release.

starwed:
I don't quite understand the condition `(document.documentElement || document.body)`.  AFAIK documentElement will exist in every browser, so there's no need for the fallback to body.

Perhaps it should test directly for whether `document.documentElement.scrollTop` is nonzero instead?

sbyrnes321:
For what it's worth, here is some jquery code that I think does a related function --https://github.com/jquery/jquery/blob/c2aca17d457d302cb1683f925b9e5ee93f0984ea/src/event.js#L486

In Chrome 31 console right now, I see document.documentElement.scrollTop = 0 but document.body.scrollTop = 829. Don't ask me why ... (But I think that this means that this pull request will make Chrome 31 not work right.)

(Update: Here is a page describing the change -- https://www.shieldui.com/blogs/chrome32-update-causes-visual-issues )

starwed:
Both jQuery and shieldUI pages suggest that the term should be like something like

``` javascript
(documentElement.scrollTop || body.scrollTop || 0)
```

which would prevent breakage in old versions of Chrome (and maybe other browsers?)

potomak:
@starwed 

> only one or the other of `body | documentElement` had a scrollTop attribute?

Yes, this is how it should be.

> This fix is needed becuase Chrome now returns values for both?

Yes, version 32 of Chrome returns a value for both `document.body.scroll<X>` and `document.documentElement.scroll<X>`.

> I don't quite understand the condition `(document.documentElement || document.body)`. AFAIK `documentElement` will exist in every browser, so there's no need for the fallback to body.

You're right, I thought `documentElement` was missing in old browsers.

@sbyrnes321 

> In Chrome 31 console right now, I see `document.documentElement.scrollTop = 0` but `document.body.scrollTop = 829`. Don't ask me why ... (But I think that this means that this pull request will make Chrome 31 not work right.)

Yes, this was the old behavior and yes, it would brake old browsers, I'm going to fix it, thanks for reporting it! If you open `tests/dom_translate.html` test case you should see the failing test.

potomak:
@starwed @sbyrnes321 I updated the PR, thanks for your feedback. It's already rebased and ready to be merged if you think it's ok.

starwed:
Thanks!  I'll go ahead and merge this now, so I can post a release candidate.

Since the tests aren't automated, we'll have to remember to run them before future release?  Better than no tests at all, of course.

potomak:
> Since the tests aren't automated, we'll have to remember to run them before future release?

Unfortunately yes, I didn't find a way to automate that test, implying page scrolling, with phantomjs.

PR #710, opened 2014-01-31
Process main docblock separately from trigger block to work around bug

starwed:
This fixes an issue where the doc generator wasn't parsing fenced codeblocks correctly in some cases.

It seems like it might have been caused by something sitting in the event trigger blocks -- this works around the issue by just processing the blocks separately. 

This is a pretty shallow fix, but we can figure out the real issue after the 0.6.2 release.


starwed:
Merging right away, since it's a low risk change we need to support the next release.

PR #713, opened 2014-02-02
Throw an error if duplicate component function name

kevinsimper:
This pullrequest tried to solve the problem when you accidently make a function with the same name as another component already defined.

If you do define a component with a for an example move function and then add the 2D component to your entity, depending on the order of the components added, you would get weird error depending on what is in the move function, which does not look like your move function. This is not userfriendly.

It should not be possible to overwrite other component functions in a entity.

This is not the solution, as this prevents any two components of having the same object function name. This would make it more errorsafe, but would also force you to come up with new names, eventhrough the components would newer be used together, like a component for a hero and a enemy.
I also tried modifing the `Crafty.fn.extend`, but that made it complex as that function is used to extend the Crafty global namespace and not just the local entity.
I then tried checking every time if the entity had a `.getId` key in the object

Another solution would be to create a extend function only to extend entities with components. This way the normal `Crafty.fn.extend` would stay simple, and we would use a `Crafty.fn.extendEntity` function.

There is a couple of things to keep in mind when solving this problem:
- The init function has to ignored. It gets overwritten multiple times, but only used once every time.
- It should only be functions, as x, y, w, h and suchs variables get declared multiple times in initialization.

What does you others think?

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1313,7 +1313,23 @@ Crafty.extend({
      *
      * @see Crafty.e
      */
+    commonComponentFunctions: ['init', 'draw', 'remove', 'matchHitBox'],
+    componentFunctions: [],
     c: function (compName, component) {
+
+        for(var key in component){
+
+            if(typeof component[key] == "function"){
+                // If the function has not been in a component before, add it
+                if(this.componentFunctions.indexOf(key) === -1){
+                    this.componentFunctions.push(key);
+                } else if(this.commonComponentFunctions.indexOf(key) !== -1){
+                    // If the function name is a common name like init, do not throw an exception
+                } else {
+                    throw('Function already defined in another component: ' + key);
+                }
+            }
+        }
         components[compName] = component;
     },
 
--- a/src/core.js
+++ b/src/core.js
@@ -1313,17 +1313,17 @@ Crafty.extend({
      *
      * @see Crafty.e
      */
-    commonComponentFunctions: ['init', 'draw', 'remove', 'matchHitBox'],
-    componentFunctions: [],
+    _commonComponentFunctions: ['init', 'draw', 'remove', 'matchHitBox'],
+    _componentFunctions: [],
     c: function (compName, component) {
 
         for(var key in component){
 
             if(typeof component[key] == "function"){
                 // If the function has not been in a component before, add it
-                if(this.componentFunctions.indexOf(key) === -1){
-                    this.componentFunctions.push(key);
-                } else if(this.commonComponentFunctions.indexOf(key) !== -1){
+                if(this._componentFunctions.indexOf(key) === -1){
+                    this._componentFunctions.push(key);
+                } else if(this._commonComponentFunctions.indexOf(key) !== -1){
                     // If the function name is a common name like init, do not throw an exception
                 } else {
                     throw('Function already defined in another component: ' + key);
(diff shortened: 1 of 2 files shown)


mucaho:
Nice idea, but crafty should also throw exceptions when someone tries to override ordinary properties.

> It should only be functions, as x, y, w, h and suchs variables get declared multiple times in initialization.

You have included a list of functions that are allowed to be overriden (called `_commonComponentFunctions` in your code). Would it be difficult to also add a list of properties that are allowed to be overriden - like the x,y,... you mentioned ?

mucaho:
Keep in mind that even if we check the properties of newly added components, the user can still go ahead and overwrite properties with ease (e.g. a component could overwrite properties/functions in its **init** method).

The current _property override-checking_ you propose is definitely better than having no checking whatsoever. 
A **safer** way is to only define properties with [Object.defineProperty()](http://craftyjs.com/api/) - or analog methods. This would require adapting the **whole** crafty code to explicitly use `Object.defineProperty`. and it could incur performance loss.

kevinsimper:
> Would it be difficult to also add a list of properties that are allowed to be overriden - like the x,y,... you mentioned ?

The common properties is basicly a list of ignore properties, it also only checks functions, so `strings`, `arrays` and `objects` will not throw an exception.

https://github.com/craftyjs/Crafty/pull/713/files#diff-7eb52b366866677666470e019283c8eaR1322

starwed:
I don't think it's necessary to be 100% safe.  If someone _wants_ to override a function, we should let them.

I'd lean towards not even throwing an error -- just adding a line in extend that warns you when a component is overridden.  It could just log a message when the target property already exists.

kevinsimper:
>  If someone wants to override a function, we should let them.

Totally, 

> just adding a line in extend that warns you when a component is overridden.

If you do that it will log about 100 times, not i viable solution, both because of init which constantly gets overwritten. But then we should make a custom extend function which is only for entities.

Funny i saw this after i made this: 
http://stackoverflow.com/questions/20905686/override-function-in-craftyjs

mucaho:
> If someone wants to override a function, we should let them.

That's the beauty of `Object.defineProperty()`: You can define a function that can't be overwritten, but can be reconfigured. 
That means you can use `Object.defineProperty()` on an existing function in order to reconfigure it so that it can be overwritten. Then you can go ahead and overwrite the function. After that you can go ahead and protect it again by making it unmodifiable.
So there are additional steps someone has to go through in order to override/delete a property. There will be no overriding by accident :) . Here is a [JSFiddle](http://jsfiddle.net/Zm6Rv/4/) showing the process. On the other hand, I agrree, this might be a bit of an overkill.

> it also only checks functions, so **strings, arrays and objects** will not throw an exception.

Yes I understand that :) But why don't you throw an exception for strings, arrays and objects also?
What's your reasoning behind it?

starwed:
> But then we should make a custom extend function which is only for entities.

That would make a lot of sense.  (Aside: why does `extend` not copy self-references?  It's not a deep copy, so that shouldn't really matter...)

mucaho:
> why does extend not copy self-references

Does extend copy the reference to a property object or the object itself? I will take a wild guess here if it copies the reference: We have some kind of storage, right? When you try to do `JSON.stringify` you have to explicitly exclude circular (superset of self-) references. If you do not do that you get [circular reference exception](http://stackoverflow.com/questions/10392293/stringify-javascript-object-with-circular-reference).

kevinsimper:
Not relevant anymore

PR #716, opened 2014-02-02
Update dependencies - removed tilde to ensure same version

kevinsimper:
Updated dependencies since both the qunit now use a two newer version of PhantomJs.

I did check all the changelog and nothing breaking seem introduced in any of them.

Also removed tilde to ensure same version at all contributers.

The new version of JsHint also found a new bug, which properly make this PR look broken.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,17 +63,16 @@
     "src/keycodes.js"
   ],
   "dependencies": {
-    "grunt": "~0.4.1",
-    "grunt-contrib-uglify": "0.2.0",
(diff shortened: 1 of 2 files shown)


starwed:
Nice, and removing the tilde is a good idea.  (I think that was causing me npm conflicts at one point.)

Looks like jshint will be satisfied if you change the comma on  L49 to a semicolon.  Could you add a commit for that so the tests pass?

kevinsimper:
@starwed I fixed it now.

starwed:
Thanks, merging.

PR #721, opened 2014-02-05
New documentation generator

kevinsimper:
This takes all the comments from all the files and in the most unopinionated way generate one json file.

The json file can then be used to generate the docs the way you want.

I am currently working on a angular app which loads the json file, and thereby do not need any compilation.

Here is the whole file it generates:
https://gist.github.com/kevinsimper/8832896


bayleedev:
Angular :+1:

I found a github flavored markdown parser written in js a bit ago. Please hold........Found again [here](https://github.com/chjj/marked)!

Good luck!

starwed:
We should maybe include the file names and the line numbers of the docs as metadata.

@blainesch  Marked is actually what we use right now; it seems to be the only decent parser for github flavored stuff.

kevinsimper:
> We should maybe include the file names and the line numbers of the docs as metadata.

The filename is there already, line number will be easy to include.

kevinsimper:
Just added the angular app. It is very simple and only categories the content just as the old did.

But i need help setting up the content on the page. Now it just shows the json which should be shown on the page.

How to get rolling:
´grunt api´
´grunt connect´
Go to ´localhost:8000/build´

kevinsimper:
I updated the docs, because it was wrong some places and it caused the angular app not to show it.

There is still some way to go, but it is not long.

The cool thing with it being one json file, it that you can make a search function in Angular, which then would work offline!

Here is the current state, try click at the "Crafty Core", that is the one that is most optimized.

http://kevinsimper.github.io/craftydocs/

starwed:
Cool, good to see work happening here.  :)

Just glancing at it, a few questions/comments:
- Do we actually want a dynamically built doc set in browser?  There was actually a very long, noticeable pause while it loaded the json.  
- Does the craftydocs link support linking to a particular page?
-  Either the initial parsing needs to be more aware of the formats, or there needs to be a second pass.  For instance, there's a particular format that a `@param` tag is supposed to follow: there's the type+name, and the description.  That should be dealt with at some point before feeding into angular or whatever other generator we use.
- I'm confused why you changed references in the docs to `.HashMap` when the actual code still uses `.map`.

kevinsimper:
> Do we actually want a dynamically built doc set in browser? There was actually a very long, noticeable pause while it loaded the json.

The page takes half a second to load on my machine. The `docs.json` is only 160 kb, that is the size of a javascript library. 

> Does the craftydocs link support linking to a particular page?

Yes, it will be possible to link to the individual "pages". I am aiming for something like this: http://backbonejs.org/ where it is very easy to read all the docs and understand it

> Either the initial parsing needs to be more aware of the formats, or there needs to be a second pass. For instance, there's a particular format that a @param tag is supposed to follow: there's the type+name, and the description. That should be dealt with at some point before feeding into angular or whatever other generator we use.

I can see what you mean, I setup the layout wrong, thanks for noticing!

> I'm confused why you changed references in the docs to .HashMap when the actual code still uses .map.

I just changed it fast so that the docs generator would generate the correct docs. It should be called HashMap because that is what the function is called. That Crafty then initialize a Crafty.map function is a other thing, and the docs should then just tell that Crafty.map is a function of HashMap

starwed:
> It should be called HashMap because that is what the function is called. That Crafty then initialize a Crafty.map function is a other thing, and the docs should then just tell that Crafty.map is a function of HashMap

The docs should refer to it by the same name someone calling it from code would use -- HashMap is like the name of a private variable, and `Crafty.map` is the public interface.  (We only ever create on hashmap.)  The docs shouldn't care about the implementation details.:)

kevinsimper:
Then we should rename all of them to be called `.map`.

The .map documentation is very spares and is only pointing to .HashMap.
http://craftyjs.com/api/Crafty-map.html

The documentation comments should document the code following the comment, else It will make it very difficult to figure out. (Luckily this is the only example.)

PR #724, opened 2014-02-06
unselectable text should have the default arrow cursor, not text cursor, on mouse-over

sbyrnes321:


starwed:
Makes sense.  (There's no way to reverse unselectable, right?)

kevinsimper:
@starwed What do you mean? There is nothing to reverse the unselectable method.

starwed:
Heh, just a typo.  (Should have been "no way to"...)

sbyrnes321:
Correct, there is no way to reverse unselectable. I doubt anyone would want to do that. If someone requests it, it's obviously an easy feature to add.

Issue #711, opened 2014-01-31
Canvas not being cleared properly while zoomed


starwed:
Reported on [stackoverflow](http://stackoverflow.com/questions/20491607/white-lines-at-zooming)

A jsfiddle using 0.6.2 [reproduces the issue](http://jsfiddle.net/ysDtQ/48/)

mucaho:
I can confirm the lines in Chrome; no problems in FF though

starwed:
I see lines in both Chrome and FF, but in slightly different ways.

After reading over the code again, I don't see anything obvious we're doing wrong.  I'll try to reproduce the issue outside of crafty, in case it's just some canvas bug when you set a transform matrix.

starwed:
Hmm, possibly related is [this report](https://groups.google.com/forum/?fromgroups#!topic/craftyjs/h5MHTMsQcAI).

starwed:
I'm a bit confused why something like this would _start_ appearing in 0.6.2 -- I don't think we even changed the canvas drawing code?

I guess I'll do some sort of bisection to figure out where the problem was introduced.

starwed:
Ok, bisection indicates that it's from [this commit](https://github.com/starwed/Crafty/commit/0ad5c872ba1166533f3e4d17872982a4933c1b5b).

Nothing there directly touches the drawing code -- but notably, Furniman calls `centerOn`, which _is_ affected.  So like the problems with `zoom`, I'd guess it's something about viewports coordinates.

starwed:
Yup, before that commit the viewport coordinates are integers in the testcase, afterwards they're fractional.

Presumably this is the problem with the `zoom` function, too.  When the viewport transformation offsets our pixels from "real" pixels some sort of error creeps in.

starwed:
Ok, so we can half fix this: clearing the canvas area _after_ clipping to it fixes the problem for me in Chrome.

Unfortunately, Firefox behavior here seems broken.  [Here](http://jsfiddle.net/VAXrL/1059/) is a jsfiddle that reproduces the issue.  I'm not really sure how we can work round this when the canvas is zoomed in.

However, we can at least round the actual viewport coordinates to integers when we aren't scaled at all.  That should help mitigate the issue.

ZeRo2150:
[...]

PR #725, opened 2014-02-07
Round translation in canvas setTransform

starwed:
The translation in setTransform should be rounded to the nearest pixel, since canvas implementations deal with fractional coordinates inconsistently.

This helps out #711, but zoomed canvases will still have these types of artifacts.

_e:_ I guess I should note that this doesn't actually round the pixels, but I think all we care about is that they're integral.

### DIFF
--- a/src/drawing.js
+++ b/src/drawing.js
@@ -517,7 +517,7 @@ Crafty.DrawManager = (function () {
 
             if (dirtyViewport) {
                 var view = Crafty.viewport;
-                ctx.setTransform(view._scale, 0, 0, view._scale, view._x*view._scale, view._y*view._scale);
+                ctx.setTransform(view._scale, 0, 0, view._scale, Math.round(view._x*view._scale), Math.round(view._y*view._scale) );
 
             }
             //if the amount of changed objects is over 60% of the total objects


kevinsimper:
> The translation in setTransform should be rounded to the nearest pixel

But is this nearest pixel?:

```
14.25|0
14
14.65|0
14
```

mpetrovic:
Val + 0.5 | 0 is round

starwed:
> Val + 0.5 | 0 is round

Only for positive values.  (And the most common viewport values would actually be negative.)

Proper (fast) rounding is something like `val>0 ? val+0.5 | 0 :  val-0.5 | 0`, which is why I went the easy route.  Might be better to just opt for legibility and use `Math.round` -- it only happens once per frame.

mucaho:
is this related to #638?

starwed:
Ok, changed it to use Math.round.

@mucaho  No, this only affects the viewport, not the position of entities.  (And would be necessary even if entities _were_ coerced to integral positions.)

_e:_ Although maybe it was what was causing the reporter to see white lines?

starwed:
Relatively small change that fixes a nasty bug, so lets merge.

PR #727, opened 2014-02-07
Reorganize init/removal of DOM component

starwed:
DOM wasn't really cleaning up after itself.  This moves all the logic to the special `remove` function, and unbinds all the events properly.  To unbind them, the event handlers are now _methods of the DOM component.

There was an also an issue where undraw would throw when the entity's element wasn't actually attached directly to the stage.  It's now just removed from its parentNode.

This should resolve an issue that was reported [in the forum](https://groups.google.com/forum/?fromgroups#!topic/craftyjs/nl6OngFj5eA).

### DIFF
--- a/src/DOM.js
+++ b/src/DOM.js
@@ -40,44 +40,19 @@ Crafty.c("DOM", {
         this._element.style.position = "absolute";
         this._element.id = "ent" + this[0];
 
-        this.bind("Invalidate", function () {
-            if (!this._changed) {
-                this._changed = true;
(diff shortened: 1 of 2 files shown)


starwed:
I'm not really sure about whether we have good test coverage of DOM stuff, so I might try to add some for this.

starwed:
Added some simple tests.

_e:_ And fixed linting errors in them -- forgot that running the tests in browser will not check for that.  :)

Issue #581, opened 2013-10-25
CI with testling-ci


potomak:
To do:
- [ ] update package.json (see http://ci.testling.com/#package.json)
- [ ] update tests to produce TAP output using https://github.com/substack/tape (see http://ci.testling.com/#write)
- [ ] add a webhook to http://git.testling.com (see http://ci.testling.com/#hook)
- [ ] add badge in README.md (see http://ci.testling.com/#badge)

See also #540

kenany:
Looks like QUnit can use a [plugin](https://github.com/twada/qunit-tap) to produce TAP output. Might be easier than converting over to tape (don't get me wrong, I love tape).

mucaho:
@starwed and @sorenbs have owner rights now. They can add the webhook.

If either of you two (potomak/KenanY) feel up for it, you can go ahead and start coding :)

kevinsimper:
@mucaho Yes, that is a start, but the test has to be converted from QUnit to either Mocha and Tape before we can use Testling. (sorenbs has been inactive for a long time unfortunatly).

mucaho:
@kevinsimper Can't we use the [Qunit TAP plugin](https://github.com/twada/qunit-tap) as suggested by KenanY? No need to convert any test, just some preprocessing/postprocessing, right?

mucaho:
There is another alternative called [Sauce Labs](https://saucelabs.com/opensauce/), which is also free for OSS. [Rumored](https://github.com/Financial-Times/ft-origami/issues/218) to be faster than testling.

EDIT:
Yes, testling is effectively [down](https://github.com/substack/testling/issues/88#issuecomment-67369563). The required modifications to the source files (if testling will work again) can be found in my [testling branch](https://github.com/mucaho/Crafty/commit/1f4104e8e112138dd1cf01f4b3d147c25bd64d75)

EDIT2:
Found alternative [zuul](https://github.com/defunctzombie/zuul#cross-browser-testing-via-saucelabs). It works out of the box with `qunit`, and runs the cross-browser tests in sauce labs, after it is initiated on travis ci through a new build.

EDIT3:
Zuul would be quite complicated to integrate with grunt. Luckily there is [`grunt-saucelabs`](https://github.com/axemclion/grunt-saucelabs#integration-with-a-ci-system) - a grunt plugin that we can use to run the tests on sauce labs servers (be it from a local initiated test run or by travis initiated test run)

mucaho:
[...]

PR #728, opened 2014-02-07
Switch to defineProperty; set enumerable flags to be consistent with private/public properties

### LINKED ISSUE #722: Collision hitbox - wrong coordinates - Crafty.e().clone
The issue was originally posted on the [forum](https://groups.google.com/forum/#!topic/craftyjs/8-e5HKiHkJg).

I tracked the bug down, here is the scenario:
- Create an entity with Collision component and place the entity somewhere (x != 0, y != 0)
- Clone the entity; here is what happens:
  - All of the original enitity's components get added to the cloned entity
  - All of the copy-by-value properties (primitive types like numbers) get copied to the cloned entity (except special ones like the entitiy id)  
- The collision hitbox of the cloned entity is wrongly offset by (-originalEntity.x, -originalEntity.y)

The problem occurs because the x, y of the original entity get copied to the cloned entity **after** the Collision component was added to the cloned entity.
Internally, the Collision component attaches it's hitbox to the entity, so that it moves alongside this parent entity. The problem is the hitbox's initial position is wrong and any later x/y changes through attach method are **relative**. That gives us a constant offset between the cloned entity and its hitbox **(-originalEntity.x, -originalEntity.y)**

starwed:
__defineGetter and company are deprecated; this switches completely to defineProperty, and uses the enumerable flag to fix #722.

I've tested in Firefox, Chrome, and Safari.  It still needs testing in IE and mobile browsers.

### DIFF
--- a/src/2D.js
+++ b/src/2D.js
@@ -195,8 +195,11 @@ Crafty.c("2D", {
             get: function () {
                 return this._x;
             },
-            configurable: true
+            configurable: true,
+            enumerable: true
         });
+
+        Object.defineProperty(this, '_x', {enumerable:false});
 
         Object.defineProperty(this, 'y', {
             set: function (v) {
@@ -205,8 +208,10 @@ Crafty.c("2D", {
             get: function () {
                 return this._y;
             },
-            configurable: true
+            configurable: true,
+            enumerable: true
         });
+        Object.defineProperty(this, '_y', {enumerable:false});
 
         Object.defineProperty(this, 'w', {
             set: function (v) {
@@ -215,8 +220,10 @@ Crafty.c("2D", {
             get: function () {
                 return this._w;
             },
-            configurable: true
+            configurable: true,
+            enumerable: true
         });
+        Object.defineProperty(this, '_w', {enumerable:false});
 
         Object.defineProperty(this, 'h', {
             set: function (v) {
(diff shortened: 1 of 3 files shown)


mucaho:
:+1:
You could also remove the enumerability for viewport private properties (for consistency's sake, although it makes no difference)
Another thing to keep in mind is that `Crafty.e().setter()` also uses `Crafty.support.setter`, but that gets refactored in #708 , so I will remove that in #708 once it lands (but I'm sure that was your intention anyway :) )

mucaho:
I was worried that the property flags could decrease the performance, but that [is not the case](http://jsperf.com/js-access-to-not-enumarable-property).

kevinsimper:
@starwed 
I can test it on IE and mobile, can this be done trough the tests?

What about the #581, it will run tests in IE easy and fast :)

starwed:
Yeah, if anything goes wrong from this it should be pretty catastrophic, so the tests should catch it!  :)

I guess also check the [fiddle example from 722](http://jsfiddle.net/AqU7E/4/) -- you should see both a green and a red square.

(581 looks useful.)

starwed:
Seems to work fine on Android in the mobile versions of Firefox and Chrome.

starwed:
@kevinsimper  or anyone else check in IE?

mucaho:
I can confirm it works for [IE](http://www.browserstack.com/screenshots/680457b4a05c92c5a0f7804691fe2e211485beb4)

starwed:
Thanks!  I'll go ahead and merge then.

PR #730, opened 2014-02-09
if no color is assigned, the Color component should do nothing whatsoever

sbyrnes321:
Before this change, in canvas mode, it still does the fillRect command but without specifying a fillStyle. So it fills the rectangle with whatever random color was last used.


kevinsimper:
Looks fine, good catch! :+1:

PR #731, opened 2014-02-12
Pause *before* firing "AnimationEnd"

starwed:
Currently, if you try to play a new animation by binding to "AnimationEnd", it'll be paused right after the event fires.  This just switches the order, and adds a test for animation chaining.  (Which we really need, because I'm pretty certain there have been a bunch of related bugs over time.)

This was reported [on the forums](https://groups.google.com/forum/?fromgroups#!topic/craftyjs/FfVTSBfJXVM).


mucaho:
:+1:

grozen:
You are a patron saint of code.

PR #733, opened 2014-02-13
Improve "Color" init/remove routine

starwed:
- Allows the "Color" component to be properly removed from entities.
- Properly invalidates the entity when "Color" is added.

In response to [this post](https://groups.google.com/forum/?fromgroups#!topic/craftyjs/oTPqijDiz9I).

### DIFF
--- a/src/drawing.js
+++ b/src/drawing.js
@@ -11,16 +11,28 @@ Crafty.c("Color", {
     ready: true,
 
     init: function () {
-        this.bind("Draw", function (e) {
-            if (!this._color) { return; }
-            if (e.type === "DOM") {
-                e.style.backgroundColor = this._color;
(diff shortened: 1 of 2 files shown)


sbyrnes321:
Nice! Thanks!

kevinsimper:
Don't you need to bind the remove function to the event: `.bind("RemoveComponent"`?

starwed:
Nah, `remove` is a special method like `init`.  It gets run when that specific component is removed or the entity is destroyed.

Most of Crafty's built in components were written before that existed, so they don't take advantage of it, and have to bind to the "RemoveComponent" event and check whether it's the component they care about.

starwed:
(I should add a test to make sure this is working correctly, of course.)

PR #734, opened 2014-02-14
Crafty.math.Vector2D improvements

mucaho:
this PR was split from #708
- crossProduct method added
- adds optional argument to save the result into instead of instantiating a new object
- various minor test fixes

### DIFF
--- a/src/math.js
+++ b/src/math.js
@@ -364,6 +364,21 @@ Crafty.math.Vector2D = (function () {
     }; // dotProduct
 
     /**@
+     * #.crossProduct
+     * @comp Crafty.math.Vector2D
+     *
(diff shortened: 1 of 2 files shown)


starwed:
This looks good to me, although I honestly have _no idea_ what the (existing) `getNorm` function would be used for?  (Also, it doesn't normalize the resultant vector in one of the use cases, which seems suspicious.)

Anyway, that's a problem with the existing code, so I'll go ahead and merge this.

mucaho:
`getNormal()` returns two different values (depending on whether an optional parameter was filled or not):
- In the trivial case it returns the perpendicular vector to this vector (maybe add documentation for that?)
- As you mentioned in your other post, the second return value represents an unit vector perpendicular to the line between two vectors

PR #735, opened 2014-02-14
Crafty.math.Vector3D

mucaho:
this PR was split from #708

adds the Crafty.math.Vector3D feature and tests

### DIFF
--- a/src/math.js
+++ b/src/math.js
@@ -644,6 +644,505 @@ Crafty.math.Vector2D = (function () {
     return Vector2D;
 })();
 
+Crafty.math.Vector3D = (function () {
+    /**@
+     * #Crafty.math.Vector3D
+     * @category 2D
+     * @class This is a general purpose 3D vector class
+     *
+     * Vector3D uses the following form:
+     * <x, y, z>
+     *
+     * @public
+     * @sign public {Vector3D} Vector3D();
+     * @sign public {Vector3D} Vector3D(Vector3D);
+     * @sign public {Vector3D} Vector3D(Number, Number, Number);
+     * @param {Vector3D|Number=0} x
+     * @param {Number=0} y
+     * @param {Number=0} z
+     */
+
+    function Vector3D(x, y, z) {
+        if (x instanceof Vector3D) {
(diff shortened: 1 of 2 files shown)


kevinsimper:
Looks cool to me

src/math.js
```diff
@@ -644,6 +644,505 @@ Crafty.math.Vector2D = (function () {
...
+     * @param {Vector3D} vecRH
+     * @param {Vector3D} [result] the vector to store the result in
+     * @returns {Vector3D} the new normal vector
+     */
+    Vector3D.prototype.getNormal = function (vecRH, result) {
+        result = result || new Vector3D();
+        result.setValues(this);
+        return result.crossProduct(vecRH).normalize();
+    }; // getNormal
+
```
starwed, line 881: It's not clear to me that this is the equivalent of the 2D function of the same name -- I think that actually finds a unit vector that is normal to the line between the two input vectors, as opposed to being normal to both of them.

mucaho, line 881: So 2D `getNormal` ==

> an unit vector that is perpendicular to the line between two vectors

and 3D `getNormal` ==

> an unit vector that is perpendicular to the plane defined by the two 3D vectors

Do you want me to rename one of those methods? They both kind of have similar semantics though

starwed, line 881: Honestly just looking at the name, I'd expect the behavior you implemented here -- maybe the 2D one should be renamed?  They probably shouldn't have the same name, as that might be misleading.

mucaho, line 881: Well I can't think of a better name than getNormal for both 2D and 3D :)
I think we can still leave the names as they are, but rather improve the documentation.

mucaho, line 881: Ah I thought of one: rename to `getPerpendicular` for 2D. See #745

mucaho:
adjusted the documentation for `Vector3D.getNormal`

mucaho:
@starwed 
Is the inclusion of this feature worth the extra space in `crafty-min.js`? If not, is there a way to place it into the commit history and remove it right away (it may come in handy in the future)?

PR #736, opened 2014-02-20
Update the documentation for the Mouse component

kevinsimper:
It needed to be more clear that you have to include the Mouse component before you can bind to Mouse events, else you could expect it to be like jQuery and wonder why i do not trigger anything.
Removed the text "the entity" from the triggers because it makes it more easier to read and you do understand that it talks about the entity.
Included a better link for the MouseEvent details.
Updated with a working example you can copy and paste.

### DIFF
--- a/src/controls.js
+++ b/src/controls.js
@@ -390,24 +390,34 @@ Crafty.bind("CraftyStop", function () {
  * #Mouse
  * @category Input
  * Provides the entity with mouse related events
- * @trigger MouseOver - when the mouse enters the entity - MouseEvent
- * @trigger MouseOut - when the mouse leaves the entity - MouseEvent
- * @trigger MouseDown - when the mouse button is pressed on the entity - MouseEvent
- * @trigger MouseUp - when the mouse button is released on the entity - MouseEvent
- * @trigger Click - when the user clicks the entity. [See documentation](http://www.quirksmode.org/dom/events/click.html) - MouseEvent
- * @trigger DoubleClick - when the user double clicks the entity - MouseEvent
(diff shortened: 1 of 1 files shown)


src/controls.js
```diff
@@ -390,24 +390,34 @@ Crafty.bind("CraftyStop", function () {
...
+ * @trigger MouseOut - when the mouse leaves - MouseEvent
+ * @trigger MouseDown - when the mouse button is pressed on - MouseEvent
+ * @trigger MouseUp - when the mouse button is released on - MouseEvent
+ * @trigger Click - when the user clicks - MouseEvent
+ * @trigger DoubleClick - when the user double clicks - MouseEvent
+ * @trigger MouseMove - when the mouse is over and moves - MouseEvent
+ *
+ * To be able to use the events on a entity, you have to remember to include the Mouse component, else the events will not get triggered.
+ *
+ * You can read more about the MouseEvent, which are the parameter passed to the callback.
```
mucaho: You can read more about the MouseEvent, which **is** the parameter passed to the callback ;)

mucaho:
That clears up the confusion about the mouse events, thanks!
Could you also add that link to the MouseEvent to the [mouseDispatch](http://craftyjs.com/api/Crafty-mouseDispatch.html) documentation?

kevinsimper:
Done :+1:

mucaho:
Great, thanks! Merging, as it's docs fix.

kevinsimper:
@mucaho I also updated the craftyjs.com website  with more getting started. Could you look at that also? That would be awesome! https://github.com/craftyjs/craftyjs.github.com/pull/35

PR #740, opened 2014-03-01
Better documentation on Storage to explain the limitations

ashes999:
I dug around a lot to discover exactly what the limitation is. This is important, because many users will likely host multiple Crafty applications on the same domain (which means storage is split across all games).


kevinsimper:
Fine addtion.

I think you made a spelling error. Will = Well

ashes999:
Good eye. Fixed/updated.

PR #741, opened 2014-03-02
Improve attr documentation

### LINKED ISSUE #738: Update .attr docs
Right now the docs for `.attr` does not say anything about that you can also use it to get values.

But it should say that you should use the direct way instead.

```
this.attr('name', 'Hero');
this.attr('name'); // Hero
```

Somebody in the forum was confused about it:
https://groups.google.com/forum/#!topic/craftyjs/ChukZ7SvG40

mucaho:
This commit fixes #738, it also adds new signatures and more
detailed explanation to the inner workings of attr.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -416,28 +416,45 @@ Crafty.fn = Crafty.prototype = {
     /**@
      * #.attr
      * @comp Crafty Core
-     * @sign public this .attr(String property, * value)
+     * @trigger Change - when properties change - {key: value}
+     *
+     * @sign public this .attr(String property, Any value[, Boolean silent[, Boolean recursive]])
      * @param property - Property of the entity to modify
      * @param value - Value to set the property to
      * @param silent - If you would like to supress events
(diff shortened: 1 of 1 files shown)


kevinsimper:
Nice :+1: 

I do not how ever think that the internal api's should be documented public, as they are not meant to be used directly.

mucaho:
Ok, I removed the public documentation for the internal api.

kevinsimper:
I can still se the documentation for `._attr_set` amd `._attr_get`?

If you remove the `@` params, it will just be internal documentation, so that it will not be published.

mucaho:
The first line of the comment block indicates whether the docs will be parsed (on my machine at least). Only if it's `/**@` it will be parsed.

mucaho:
Ok, thanks. I removed all `@`. Better be safe than sorry :+1:

starwed:
This looks good to me, I'll go ahead and merge.

PR #743, opened 2014-03-02
Particles offset & scale

johnHackworth:
This pull request address a small issue with particles and scaling and also adds a new option to allow you to define where the origin of the particles should be placed.

When the component was calculating where it should render the particles and the size of them, the viewport current _scale wasn't being taked into account, so if you zoomed the viewport, the particles place & size were incorrect.

Also, I've added a small option to allow users to define where the origin of the particles should be, allowing them to add a offset to the position of the particles' parent object.

### DIFF
--- a/src/particles.js
+++ b/src/particles.js
@@ -25,35 +25,38 @@ Crafty.c("Particles", {
      * @example
      * ~~~
      * var options = {
-     *	maxParticles: 150,
-     *	size: 18,
-     *	sizeRandom: 4,
-     *	speed: 1,
-     *	speedRandom: 1.2,
-     *	// Lifespan in frames
-     *	lifeSpan: 29,
-     *	lifeSpanRandom: 7,
-     *	// Angle is calculated clockwise: 12pm is 0deg, 3pm is 90deg etc.
-     *	angle: 65,
-     *	angleRandom: 34,
-     *	startColour: [255, 131, 0, 1],
-     *	startColourRandom: [48, 50, 45, 0],
-     *	endColour: [245, 35, 0, 0],
-     *	endColourRandom: [60, 60, 60, 0],
-     *	// Only applies when fastMode is off, specifies how sharp the gradients are drawn
-     *	sharpness: 20,
-     *	sharpnessRandom: 10,
-     *	// Random spread from origin
-     *	spread: 10,
-     *	// How many frames should this last
-     *	duration: -1,
-     *	// Will draw squares instead of circle gradients
-     *	fastMode: false,
-     *	gravity: { x: 0, y: 0.1 },
-     *	// sensible values are 0-3
-     *	jitter: 0
-     * }
-     *
+     * maxParticles: 150,
+     * size: 18,
+     * sizeRandom: 4,
+     * speed: 1,
+     * speedRandom: 1.2,
+     * // Lifespan in frames
+     * lifeSpan: 29,
+     * lifeSpanRandom: 7,
+     * // Angle is calculated clockwise: 12pm is 0deg, 3pm is 90deg etc.
+     * angle: 65,
+     * angleRandom: 34,
+     * startColour: [255, 131, 0, 1],
+     * startColourRandom: [48, 50, 45, 0],
+     * endColour: [245, 35, 0, 0],
+     * endColourRandom: [60, 60, 60, 0],
+     * // Only applies when fastMode is off, specifies how sharp the gradients are drawn
+     * sharpness: 20,
+     * sharpnessRandom: 10,
+     * // Random spread from origin
+     * spread: 10,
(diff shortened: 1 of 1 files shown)


src/particles.js
```diff
@@ -184,6 +187,10 @@ Crafty.c("Particles", {
 
             this.emissionRate = this.maxParticles / this.lifeSpan;
             this.positionRandom = this.vectorHelpers.create(this.spread, this.spread);
+            this.originOffset = {
```
mucaho, line 190: I would not provide a default `originOffset` here. 
You can add the default values to the end of the [presets object](https://github.com/johnHackworth/Crafty/blob/develop/src/particles.js#L135-L175).

src/particles.js
```diff
@@ -25,35 +25,38 @@ Crafty.c("Particles", {
...
+     * sharpnessRandom: 10,
+     * // Random spread from origin
+     * spread: 10,
+     * // How many frames should this last
+     * duration: -1,
+     * // Will draw squares instead of circle gradients
+     * fastMode: false,
+     * gravity: { x: 0, y: 0.1 },
+     * // sensible values are 0-3
+     * jitter: 0
```
mucaho, line 54: `,` at end of line missing

src/particles.js
```diff
@@ -25,35 +25,38 @@ Crafty.c("Particles", {
...
+     * // X Offset for the origin of the particles
+     * originOffsetX: 0
```
mucaho, line 56: `,` at end of line missing

src/particles.js
```diff
@@ -25,35 +25,38 @@ Crafty.c("Particles", {
...
+     * // Y Offset for the origin of the particles
+     * originOffsetY: 0
```
mucaho, line 58: `,` at end of line missing

src/particles.js
```diff
@@ -25,35 +25,38 @@ Crafty.c("Particles", {
...
+
```
mucaho, line 59: end the `options` object with `};` properly

mucaho:
Thanks for the PR!
I added some line comments to your code. Additionally:
- The originOffset logic looks good to me.
- I am not sure that scaling the particle position & size **directly** is the best way to do it:  
  The particles are drawn by an additional canvas. The default Crafty canvas uses [a transformation matrix](https://github.com/johnHackworth/Crafty/blob/develop/src/drawing.js#L533) to scale entities automatically. Maybe it would be better to use a transformation matrix on the [canvas of the particles](https://github.com/johnHackworth/Crafty/blob/develop/src/particles.js#L82) also.

(Ad Particle Canvas: The viewport offsets/ viewport scale usage should be refactored so that these offsets/scale apply to the **canvas itself**, rather than modifiying **each particle seperately**: The particle canvas should listen to viewport events and modify itself appropriately.)

mucaho:
@starwed what do you say? is the direct `particle scale / offset` manipulation ok for now? refactor to `canvas.context.transform scale / offset` in a seperate PR?

johnHackworth:
That would make a lot more sense, in fact... if it's using another canvas (sorry! I didn't see the whole forest, just the few trees of particle.js :) ), viewport.zoom and so should probably affect it too!

(I'll make a pull separated pull request for the offset parameters with the comment/doc fixed, by the way)

mucaho:
Your **current** particle scale implementation is actually good, as it's streamlined with the rest of the code. It's preferable for now. Don't change it :)

In the future we should make a complete change to the `Particles.js`, where the all the offsets and scales get refactored.

mucaho:
@johnHackworth 
Do the particles still appear correctly if you shift the viewport (e.g. `Crafty.viewport.x += 1000`) in conjunction of scaling it?

johnHackworth:
@mucaho  yep... check this gif, for example : 



That viewport is scaled and is being manually displaced to follow keep the rocket at the center of the viewport, and the particles just move along.

mucaho:
> That viewport is scaled and is being manually displaced to follow keep the rocket at the center of the viewport, and the particles just move along

Great (nice looking game btw!).

Could you add a commit with the offset parameters and comment/doc fixes and maybe squash the commits, so we can merge it?

(If you are wondering how to squash commits, this _should_ work on your local repo:)

```
git checkout develop
git rebase -i <put the first letters of the revision just before your first commit here>
<pick the first commit, squash the following commits in the editor that comes up, save & exit editor>
<consolidate all messages into one meaningful one in the 2nd editor that comes up, save & exit editor>
git push --force <this command overwrites the branch on your github repo with your local branch>
```

johnHackworth:
ok, done! (sorry for the long delay!)

mucaho:
Great! Added some suggestions via  johnHackworth/Crafty#1

mucaho:
merged with #881

PR #755, opened 2014-03-14
Add more info to Component docs

### LINKED ISSUE #752: Improve Component documentation
A recent [forum post](https://groups.google.com/forum/#!topic/craftyjs/LujX5Fqocds) indicates confusion about the component - entity system. I had my troubles too, until I wrapped my head around it.

As a start, I propose we add documentation to `Crafty.c`:

> When a component method gets called, the `this` keyword will refer to the current entity the component was added to

mucaho:
fixes #752 .

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1386,6 +1386,12 @@ Crafty.extend({
      * Creates a component where the first argument is the ID and the second
      * is the object that will be inherited by entities.
      *
+     * Specifically, each time a component is added to an entity, the component properties are copied over to the entity. 
+     * * In the case of primitive datatypes (booleans, numbers, strings) the property is copied by value.
+     * * In the case of complex datatypes (objects, arrays, functions) the property is copied by reference and will thus reference the components' original property.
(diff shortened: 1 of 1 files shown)


kevinsimper:
Good idea to add that! Can't we link to some place that explains shallow and deep copy, if a person want to ready more? (Could not seem to find any good ones..)

But the `this` context and behavior is the more difficult part in javascript.

I changed "primitive datatypes", because I think that could confuse people.

```
* In the case of strings and numbers, the value is copied directly.
* In the case of complex datatypes (objects, arrays, functions) the property is copied by reference and will thus reference the components' original property
* (See the two examples below for further explanation).
```

mucaho:
Thanks, added your suggestion. The entity always gets a shallow copy of the component, if I understand correctly.

mucaho:
I suggest we add deep & shallow copy explanation to the `Crafty.e().clone()` method. I wil go ahead and merge this as it's docs and no further suggestions so far :)

PR #758, opened 2014-03-19
Delay modified

heniotierra:
At its current state, the "Delay" component do not do anything when the delay ends. This modification fixes that, allowing you to pass a "callbackOff" method that will be called after all iterations of delay are executed.

Also, now it's possible to cancel a delay with the method .cancelDelay, passing as argument a method reference, the same method reference passed to .delay.

Finally, fixed a minor inconsistency in the the comments.

### DIFF
--- a/src/2D.js
+++ b/src/2D.js
@@ -1012,6 +1012,8 @@ Crafty.c("2D", {
 /**@
  * #Gravity
  * @category 2D
+ * @trigger Moved - When entity has moved on y-axis a Moved event is triggered with an object specifying the old position {x: old_x, y: old_y}
+ * 
  * Adds gravitational pull to the entity.
  */
 Crafty.c("Gravity", {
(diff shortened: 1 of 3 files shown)


heniotierra:
(And I thought I had done everything right this time...)

Well, I'm sorry, it wasn't my intention to include these three files into this PR. It was supposed to be one file only:   time.js. My bad. But, I don't know why this happened. Is there a way to exclude this PR?

heniotierra:
closing

starwed:
From your branch, reset to the develop branch, then commit only the files you want.  Something like:

```
git reset <commit-reference>
git add src/time.js
git commit
```

In place of <commit-reference> you'll put something like `HEAD~3`, which is 3 commits back from the current HEAD.  (You could also maybe just refer to something like `upstream/develop`.)

heniotierra:
My fork was giving me too much trouble, so I just got rid of it, and then forked again. Not the best thing to do, but I'll be more careful from now on, since I have a little bit more notion on how git works. Doing a new PR right now.

BTW, thank you! :)

PR #759, opened 2014-03-19
Improved Delay component

heniotierra:
At its current state, the "Delay" component do not do anything when the delay ends. This modification fixes that, allowing you to pass a "callbackOff" method that will be called after all iterations of delay are executed.

Also, now it's possible to cancel a delay with the method .cancelDelay, passing as argument a method reference, the same method reference passed to .delay.

Finally, fixed a minor inconsistency in the the comments.

### DIFF
--- a/src/time.js
+++ b/src/time.js
@@ -14,7 +14,7 @@ Crafty.c("Delay", {
             while (--index >= 0) {
                 var item = this._delays[index];
                 if (item.start + item.delay + item.pause < now) {
-                    item.func.call(this);
+                    item.callback.call(this);
                     if (item.repeat > 0) {
                         // reschedule item
                         item.start = now;
@@ -24,6 +24,8 @@ Crafty.c("Delay", {
                     } else if (item.repeat <= 0) {
                         // remove item from array
                         this._delays.splice(index, 1);
+                        if(typeof item.callbackOff === "function")
+                            item.callbackOff.call(this);
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


heniotierra:
Now I think it's good. :)

src/time.js
```diff
@@ -45,12 +47,15 @@ Crafty.c("Delay", {
     /**@
      * #.delay
      * @comp Delay
-     * @sign public this.delay(Function callback, Number delay)
-     * @param callback - Method to execute after given amount of milliseconds
-     * @param delay - Amount of milliseconds to execute the method
+     * @sign public this.delay(Function callback, Number delay, Number repeat, Function callbackOff)
```
mucaho: I would emphasize the optionality of the last parameter (similar to rest of Crafty code)

```
@sign public this.delay(Function callback, Number delay, Number repeat[, Function callbackOff])
```

src/time.js
```diff
@@ -68,15 +73,39 @@ Crafty.c("Delay", {
...
+     * #.cancelDelay
+     * @comp Delay
+     * @sign public this.cancelDelay(Function callback)
+     * @param callback - Method reference passed to .delay
+     *
+     * The cancelDelay method will cancel a delay set previously.
+     *
+     * @example
+     * ~~~
+     * this.cancelDelay( reference );
```
mucaho: could you add a short example here with an actual reference?
e.g.

``` javascript
var log = function() {
    console.log("100ms later");
};

console.log("start");
var ent = Crafty.e("Delay").delay(log, 100, 10);

... // sometime later in the future
ent.cancelDelay(log);
```

mucaho:
Thanks for this PR. I added some minor suggestions, but other than that this looks good.

Would you also [squash your commits into one](https://github.com/craftyjs/Crafty/wiki/Workflow#squashinq-commits), please? This helps keep a clean history of the codebase.

heniotierra:
@mucaho Done! Phew! And thanks for the suggestions! :))

mucaho:
Great, merging.

PR #760, opened 2014-03-21
Improve Delay component

mucaho:
- improve documentation
- add tests

follow-up of #759


mucaho:
I will go ahead and merge this, as it fixes a bug from #759 which is already merged.

Issue #761, opened 2014-03-22
Delay bugged. :(


heniotierra:
Hello! I noticed Delay has a problem, problem that was present prior to the modifications I made recently. In fact, I had noticed this problem some months ago, but did circumvent it with an ugly hack at the time. What happens is that the delay acts in a weird manner when there's more than one simultaneous delay set in the same entity, making the same callback be called twice (or thrice) when it should be called only once. I tried really hard to figure out why this happens, but I was unable to.

The function .cancelDelay that I made makes the problem more obvious. My only guess is that when a delay is removed with splice, it messes with the logic of the loop in some way.

That said, I found a way to fix the problem, setting the given delay as undefined instead of using .splice to remove it from the array. Then, I used another loop just to use splice, and that worked pretty well.

 EDIT: Here's the fiddle showing the bug: http://jsfiddle.net/Dj48n/9/. Also, I'll do a new PR with my fix for that.

mucaho:
I just made another PR adding some tests #760 , so I could include the fix in there, once we find the bug :)

Good that you catched that: the current testcases do not test setting up multiple delays, I will add those testcases now.

One thing that boggled my mind is that `Delay` uses property enumeration on the `_delays` array instead of array iteration. Initially I thought it doesn't matter, as the functionality does not depend on the order of iteration, but maybe it does matter after all.

mucaho:
I have added additional testcases. If there is a bug then I think it's local to the cancelDelay method:

``` javascript
cancelDelay: function (callback) {
    for (var index in this._delays) {
        var item = this._delays[index];
        if(item.callback == callback) {
            this._delays.splice(index, 1);
        }
    }
    return this;
}
```

First of all index will be randomly selected in the for loop between `0` and `this._delays.length - 1`.
When you splice an element out of the array, any old indices won't be valid any more (indices coming after the spliced element have to be shifted `1` to the left). So I am assuming the iteration over the `indices` is erroneous.

mucaho:
Yeah, that was the issue. Adding the fix to #760 .

mucaho:
I merged #760, so I will assume the bug is fixed. If you encounter another bug related to `Delay`, just reopen the issue.

heniotierra:
Unfortunately that didn't solve the bug I refer to. Here's the fiddle: http://jsfiddle.net/Dj48n/9/. One alert should appear at the end. 

Shall I do a PR with my fix?

mucaho:
Sure, that would be great!

[...]

PR #762, opened 2014-03-23
fix to issue #761

heniotierra:
Small fix for bug that occurs in very specific situations in Delay component, issue https://github.com/craftyjs/Crafty/issues/761

### DIFF
--- a/src/time.js
+++ b/src/time.js
@@ -13,7 +13,10 @@ Crafty.c("Delay", {
             var index = this._delays.length;
             while (--index >= 0) {
                 var item = this._delays[index];
-                if (item.start + item.delay + item.pause < now) {
+                if (item === false) {
+                    // remove canceled item from array
+                    this._delays.splice(index, 1);
+                } else if (item.start + item.delay + item.pause < now) {
                     item.callback.call(this);
                     if (item.repeat > 0) {
                         // reschedule item
@@ -22,7 +25,7 @@ Crafty.c("Delay", {
                         item.pauseBuffer = 0;
                         item.repeat--;
                     } else if (item.repeat <= 0) {
-                        // remove item from array
+                        // remove finished item from array
                         this._delays.splice(index, 1);
                         if(typeof item.callbackOff === "function")
                             item.callbackOff.call(this);
@@ -84,7 +87,7 @@ Crafty.c("Delay", {
      *   console.log("delay finished");
      * });
      * ~~~
-     * 
+     *
      */
     delay: function (callback, delay, repeat, callbackOff) {
         this._delays.push({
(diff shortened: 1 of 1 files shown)


heniotierra:
@mucaho Sorry, seems like I didn't merge your last modifications. Will try to fix that.

heniotierra:
I think this is it.

heniotierra:
Please, look at the last comment there.

mucaho:
Thanks for this PR! I still don't fully understand the bug though. `index`, `splice` and `push` _should_ play along together (even when nesting them inside each other in loops).

mucaho:
Yeah, the problem was that inside the EnterFrame loop (which operates on decreasing index), the array could have been shrunk even further by a callback that internally invokes the `cancelDelay` method (which shrinks the array with `splice`). The results is incorrect iteration in the EnterFrame loop.

The natural way would be to queue up delays to be deleted in a seperate array, which will then be processed at the end of the loop. This PR does it even better, by using the already existing array to invalidate entries, thus saving performance.

The code does a get a bit more complicated, but this is the best way to deal with "concurrent modifications" on an array (unless we figure out a better way to do this).

mucaho:
I searched for an easier, not so intrusive approach and I found one. It's essentially the same as yours, you can just remove the item at the start of the next "EnterFrame" instead of removing it at the end of the current frame (thus saving one extra loop).

``` diff
diff --git a/src/time.js b/src/time.js
index 02a6acb..e9544bf 100644
--- a/src/time.js
+++ b/src/time.js
@@ -13,7 +13,9 @@ Crafty.c("Delay", {
             var index = this._delays.length;
             while (--index >= 0) {
                 var item = this._delays[index];
-                if (item.start + item.delay + item.pause < now) {
+                if (item === false) {
+                    this._delays.splice(index, 1);
+                } else if (item.start + item.delay + item.pause < now) {
                     item.callback.call(this);
                     if (item.repeat > 0) {
                         // reschedule item
@@ -125,7 +127,7 @@ Crafty.c("Delay", {
         while (--index >= 0) {
             var item = this._delays[index];
             if(item.callback == callback) {
-                this._delays.splice(index, 1);
+                this._delays[index] = false;
             }
         }

```

The [new JSFiddle](http://jsfiddle.net/Dj48n/17/) works for me with this code.

heniotierra:
Nice! :) I'll update the PR with these changes. Thanks!

heniotierra:
But wait, correct me if I'm wrong. Did you put 2 (two) <i>.splice</i>'s in the loop? One at line 17 and another at line 28? The first one(line 17) for canceled delays only, right? That works marvels!
I did just update the PR.

mucaho:
Awesome, with combined effort we managed to tackle this issue :)

Since this commit is linked to the previous, merged commits for the "Delay" component, I'm going ahead and merging this.

heniotierra:
Really awesome! :dancers:

PR #763, opened 2014-03-23
All tests invoked from a single file

grozen:
Playgrounds are now in their own independent directory, and the tests they were once intertwined with are executed through `tests/index.html`.

Also, all test files no longer bleed their innards into the global namespace (which is a wise idea when you include them all from the same place).

There are a few things I am not entirely happy with - especially the fact that the playgrounds use the test folder's lib subdirectory, but that's something I would not want to solve without consulting others (so do feel free to recommend an appropriate course of action).

P.S. Do yourselves a favor and look at this diff while ignoring whitespace. There's plenty of indentation going on here.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -76,9 +76,10 @@ module.exports = function (grunt) {
         },
 
         jshint: {
-            files: ['Gruntfile.js', 'src/**/*.js', 'tests/*.js'],
+            files: ['Gruntfile.js', 'src/**/*.js', 'tests/**/*.js'],
             options: {
                 trailing: true,
+                ignores: ['tests/lib/*.js'],
                 globals: {
                 }
             }
@@ -86,14 +87,12 @@ module.exports = function (grunt) {
 
         qunit: {
             all: [
-                'tests/index.html',
-                'tests/animation/animation.html',
-                'tests/2D/collision/collision.html'
+                'tests/index.html'
             ]
         },
 
         jsvalidate: {
-            files: ['crafty.js', 'tests/*.js']
+            files: ['crafty.js', 'tests/**/*.js']
         },
 
         connect: {
--- a/playgrounds/2D/collision/collision.html
+++ b/playgrounds/2D/collision/collision.html
@@ -1,20 +1,11 @@
 <!DOCTYPE HTML>
 <html>
 	<head>
-		<script src="../../lib/jquery.min.js"></script>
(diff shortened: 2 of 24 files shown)


mucaho:
We should resolve this PR asap, as any merged PR's in the future (that add/modify tests) will break this PR. @starwed should have a look at this when he has time, as it's a pretty big change to the codebase (functionality-wise it's a small change though :) )

starwed:
Mostly looks good; a nice cleanup of the current tests!  :)

One quibble: The setup tests for animation now call `Crafty.sprite` before each test -- is that what we want?  The definition of the sprite won't be destroyed between tests, so is there a way to call it just once?

grozen:
@starwed Done. I was certain that `destroy('*')` kills sprite definitions as well which is why I called it before each test.

Turns out the newer test files were not being linted/validated, so I fixed that as well (sadly now everything is validated including `test/lib`, but the alternative was handpicking the directories to validate. Luckily, the linting task includes an option to ignore specific directories).

mucaho:
I don't agree with disabling a warning globally by default `'-W069': true, // Ignore hash access to object properties` .

I suggest you disable the warning locally for a block of code `/* jshint -W069 */` and reenable it afterwards `/* jshint +W069 */`. If you can fix it to remove the warning in the first place it's even better ;)

grozen:
@mucaho that may have been a little rash, yes. I've localized the warning suppression as you suggested. For that specific case the hash-like access makes a lot of sense, so I'll refrain from "fixing" the code.

mucaho:
>  I've localized the warning suppression

Great!
**EDIT**
On the other hand, why don't you change

``` javascript
equal(ret, spriteAnimation._reels['count'], "...");
```

to

``` javascript
equal(ret, spriteAnimation._reels.count, "...");
```

Wouldn't that fix the problem?

grozen:
Like I said, it would, technically, but _reels makes much more sense as a hash. I understand the code is equivalent, but I like the hash semantics implied by the first option.

starwed:
I guess I'm pretty agnostic on that, so I'll merge as is.[1]   Thanks for cleaning up the tests!
1. But likewise, I wouldn't turn down a patch that switched that behavior.

mucaho:
btw I just figured out, if you want github to show the diff while ignoring white space you can append `?w=1` to the url. So in this case it would be [https://github.com/craftyjs/Crafty/pull/763/files?w=1](https://github.com/craftyjs/Crafty/pull/763/files?w=1)

PR #765, opened 2014-03-26
Decouple Crafty.pixelart from individual renderers

starwed:
Currently `Crafty.pixelart()` explicitly alters the canvas and DOM renderers.  This PR has it fire an event instead.

The canvas and DOM contexts now listen for the event, and make the appropriate changes.  This will make it simpler to add new contexts (like webgl) or to introduce multiple renderers of the same type.

This patch also tracks the current pixelart setting, and ensures that, when a canvas context is inited, that it matches that value.  It doesn't do something similar for the DOM element because the stage is created as soon as you init Crafty.

The WebGL renderer is what prompted me to whip up this patch, of course.  :)  It passes tests, but TBH I haven't explicitly checked that images are appropriately smoothed or not.  So it shouldn't land until that's confirmed -- I just didn't have a convenient test case lying around.

### DIFF
--- a/src/DOM.js
+++ b/src/DOM.js
@@ -311,11 +311,33 @@ Crafty.extend({
                 // Resize the viewport
                 Crafty.uniqueBind("ViewportResize", this._resize);
 
+                // Listen for changes in pixel art settings
+                // Since window is inited before stage, can't set right away, but shouldn't need to!
+                Crafty.uniqueBind("PixelartSet", this._setPixelArt);
             },
 
             _resize: function(){
                 Crafty.stage.elem.style.width = Crafty.viewport.width + "px";
                 Crafty.stage.elem.style.height = Crafty.viewport.height + "px";
+            },
+
+            // Handle whether images should be smoothed or not
+            _setPixelArt: function(enabled) {
+                var style = Crafty.stage.inner.style;
+                if (enabled) {
+                    style[Crafty.DOM.camelize("image-rendering")] = "optimizeSpeed";   /* legacy */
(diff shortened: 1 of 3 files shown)


mucaho:
[Pixelart demo](https://github.com/craftyjs/demos/blob/master/devDemos/example_style.html)

starwed:
Thanks!  There was a silly typo at one part, which I've fixed.

Testing with Canvas and DOM, it works fine, and you no longer need to call canvas.init beforehand.

The only catch is that, if you call pixelart in the middle of a scene, canvas entities won't be drawn in the new style until something else invalidates them.  That would be fixable, but I'm not sure it's necessary?  And someone could always manually invalidate with `Crafty("Canvas").trigger("Change");` if it was needed for a game.

mucaho:
Great, that's fine. You should also adapt the documentation (as it's no longer valid)

> _This method will have no effect for Canvas image smoothing if the canvas is not initialized yet._

to something like

> _If you call pixelart in the middle of a scene, canvas entities won't be drawn in the new style until something else invalidates them. You can manually invalidate all canvas entities with `Crafty("Canvas").trigger("Change")`_

starwed:
Updated the documentation!

mucaho:
> Testing with Canvas and DOM, it works fine, and you no longer need to call canvas.init beforehand .... Updated the documentation!

Looks good, merging.

mucaho:
I tested `pixelart` on my machine, just to be safe, and I can confirm it works as expected.

PR #766, opened 2014-03-26
Enabling multiple sprites for same image

ristes:
Bug description: 
When try to register new sprite name for previously defined (but not yet loaded) image, the previous version was ignoring the new sprite. 

Now we register all the sprite names into the OnImageLoad object, and when onLoad is triggered, all sprites are registered.

### DIFF
--- a/src/sprite.js
+++ b/src/sprite.js
@@ -91,12 +91,32 @@ Crafty.extend({
             img = new Image();
             img.src = url;
             Crafty.asset(url, img);
-            img.onload = function () {
-                //all components with this img are now ready
-                for (var spriteName in map) {
-                    Crafty(spriteName).each(markSpritesReady);
-                }
+            var OnImageLoad = function (map) {
+            	this.map = {};
+            	var self = this;
+            	
+            	
+            	this.onLoad = function() {
+                    //all components with this img are now ready
+            		for (var spriteName in self.map) {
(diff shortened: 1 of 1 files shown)


starwed:
Thanks for the PR!

It looks like this creates a new OnImageLoad object for each sprite map, though, which seems like it defeats the point of designing it like a class.  Maybe just declare it once?  `sharedSpriteinit` is created each time because it needs the arguments of `Crafty.sprite` in its scope, but I don't think that's how you designed `OnImageLoad`.

Also check the linting results that result in the Travis CI failures -- it's mostly complaining about mixed tabs/spaces and other formatting issues, but there are also a couple of global variables created.

ristes:
new `OnImageLoad` is created for each image. The current implementation doesn't support additional sprite maps for image that is added into the assets, but not loaded yet (That was the problem that I faced). I'm new with Crafty, but I suppose that this is not left on purpose. 

The idea behind my patch is to assign one `OnImageLoad` instance for each image (when it is created). Afterwards, if one want to map additional sprites to that same image, the new map will be registered in the corresponding instance from the `OnImageLoad` class for the same image, and when the image is loaded, all the sprites will be marked as ready. 

In the current implementation, the `onLoad` function is registered only when the image is created, and when one want to add new map, it is not marked as ready. At least that was the case in my example.

PR #767, opened 2014-03-27
Zoom scale fix

jgswift:
This fix changes zoom functionality in crafty

The current zoom function uses multiplication when computing the final
zoom amount.  This is fine except when computing zoom scales that are
not whole numbers.

This fix changes that to, instead of multiplication, use simple addition
and subtraction.  Therefore, it is now possible to use fractional scales
(0.1 - 0.9).

I have not fully tested the change across all of crafty but it does seem
to fix a bug in my particular code.

### DIFF
--- a/src/viewport.js
+++ b/src/viewport.js
@@ -300,8 +300,14 @@ Crafty.extend({
                 else
                     v = (1/amount - 1 ) / (1/finalAmount - 1);
 
+                var newZoomScale; // prevents rezooming on pan using fractional scale
+                if (finalAmount === startingZoom)
+                    newZoomScale = startingZoom;
+                else
+                    newZoomScale = startingZoom + ( amount - startingZoom );
+
                 // Set new scale and viewport position
-                Crafty.viewport.scale( startingZoom + ( amount - startingZoom ) );
+                Crafty.viewport.scale( newZoomScale );
                 Crafty.viewport.scroll("_x", startingX * (1-v) + finalX * v );
                 Crafty.viewport.scroll("_y", startingY * (1-v) + finalY * v );
                 Crafty.viewport._clamp();
--- a/src/viewport.js
+++ b/src/viewport.js
@@ -301,7 +301,7 @@ Crafty.extend({
                     v = (1/amount - 1 ) / (1/finalAmount - 1);
 
                 // Set new scale and viewport position
-                Crafty.viewport.scale( amount * startingZoom );
+                Crafty.viewport.scale( startingZoom + ( amount - startingZoom ) );
                 Crafty.viewport.scroll("_x", startingX * (1-v) + finalX * v );
                 Crafty.viewport.scroll("_y", startingY * (1-v) + finalY * v );
                 Crafty.viewport._clamp();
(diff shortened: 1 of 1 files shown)


starwed:
I'm a bit confused -- you can currently zoom by, say, 0.5, and it will zoom out by a factor of 2.  Or you can zoom in by `2.7` and it'll zoom in that much.  What are you wanting to do that is currently impossible?  (It might be that this is a documentation issue.)

(Also, why do you use the expression `startingZoom + ( amount - startingZoom )`?  Isn't that just `amount`?)

jgswift:
Right you are.  I changed the update to reflect your comment.  That was a typo.

jgswift:
Basically the problem was that the zooming bounds would start fluxuating wildly given the following criteria.

the original scale: 1
zoom range: (0.5) - (1.5)
zoom step: 0.5

I dealt with this for 4 hours before I finally resorted to modifying crafty itself.

If you can get that working without this fix, please let me know so I can modify my code accordingly.

starwed:
Zoom currently takes the existing scale into account.  This PR makes it so that it zooms to whatever value you hand in. 

Perhaps if you gave some example code of what you were trying to do, I'd have a better idea of your issue?

jgswift:
Below is a facsimile of the offending code that produced the problem

``` js

game.camera.zoom.step = 2;
game.camera.zoom.scale = 1;
game.camera.zoom.bounds.min = 0.5;
game.camera.zoom.bounds.max = 1.5

Crafty.bind("MouseWheel", function(e) {
    var evt=window.event || e
    var delta= (evt.wheelDelta? evt.wheelDelta/120 : evt.detail)/game.camera.zoom.step;

    var newZoom;

    newZoom = game.camera.zoom.scale+delta;

    var diffZoom = game.camera.zoom.scale-newZoom;

    if(newZoom > game.camera.zoom.bounds.max || 
        newZoom < game.camera.zoom.bounds.min) {
        diffZoom = 0;
    }

    if(diffZoom !== 0) {
        game.camera.zoom.scale = newZoom;
        var target2 = hex.pos2px(Math.round(player.position.x),Math.round(player.position.y));
        Crafty.viewport.zoom(game.camera.zoom.scale,target2.left,target2.top,10);
    } 
});
```

starwed:
I'm still not understanding what is actually going wrong for you.  :(    The above code doesn't make any calls to `zoom()`, so it's not very helpful in getting at the heart of it.

In response to a mouse wheel event, wouldn't you just want to call `Crafty.viewport.scale()` directly?

If you want it to transition smoothly, what's wrong with simply passing in `targetValue/currentValue` as the zoom factor?

jgswift:
Updated the code to use zoom.

jgswift:
Code was originally based on this example - https://github.com/pengyu/Crafty-JS-Examples/blob/master/Crafty/Stage/viewport/zoom/main.html, which you can see uses zoom and not scale.

jgswift:
```
If you want it to transition smoothly, what's wrong with simply passing in targetValue/currentValue as the zoom factor?
```

That is originally what I thought too but the zoom range was still fluxuating out of bounds.

PR #768, opened 2014-03-29
Improve Collision docs

mucaho:
Adds practical examples to the collision features introduced in #699 .


starwed:
Looks good, merging!

PR #769, opened 2014-03-29
Add additional game loop events

### LINKED ISSUE #478: Event for end of frame
I'm implementing basic collision resolution for aligned boxes, and have run into the following problem:

My object applies its movement vector to its position on the `EnterFrame` event. It responds to the `onHit` message from the `Collision` component to perform collision resolution. Unfortunately, it seems that `EnterFrame` occurs after the collision message is sent, so the collision is resolved, but then right before drawing the object moves back into its original position, slightly penetrating the solid object.

Am I missing the proper way of implementing the behavior I want? (I just picked up Crafty the other day). If not, I was able to solve my problem by modifying Crafty to trigger an "ExitFrame" event right after the draw call, which I think is a useful addition.

mucaho:
Fixes #478 .

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1222,11 +1222,18 @@ Crafty.extend({
              * @comp Crafty.timer
              * @sign public void Crafty.timer.step()
              * @trigger EnterFrame - Triggered on each frame.  Passes the frame number, and the amount of time since the last frame.  If the time is greater than maxTimestep, that will be used instead.  (The default value of maxTimestep is 50 ms.) - { frame: Number, dt:Number }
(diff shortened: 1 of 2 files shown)


starwed:
I still feel kind of wishy-washy about adding new events, but there are undeniably cases where it's nice to have things in a guaranteed order, so it's probably a good idea.  You can set up your own custom events, but there are enough built-in components that use "EnterFrame" that this isn't powerful enough.

The docs for the "MeasureXTime" need to be updated to indicate which phases they include, and you should probably add some _really_ simple tests just to make sure the events are being triggered in the right order.

mucaho:
Added test for `simulateFrames`. Updated docs.

starwed:
Ok, looks good!  I'll be interested to see whether we can make more efficient collision resolution using this.

PR #772, opened 2014-04-07
Fix for bug that caused delayed function to stop being called

heniotierra:
Bug described and investigated here: https://groups.google.com/forum/#!topic/craftyjs/8Pwl6qt46y8

### DIFF
--- a/src/time.js
+++ b/src/time.js
@@ -22,7 +22,6 @@ Crafty.c("Delay", {
                         // reschedule item
                         item.start = now;
                         item.pause = 0;
-                        item.pauseBuffer = 0;
                         item.repeat--;
                     } else if (item.repeat <= 0) {
                         // remove finished item from array
@@ -44,6 +43,7 @@ Crafty.c("Delay", {
             for (var index in this._delays) {
                 var item = this._delays[index];
                 item.pause += now - item.pauseBuffer;
+                item.pauseBuffer = 0;
             }
         });
     },


starwed:
Hmm, elsewhere we now handle timing by hooking in to the EnterFrame event's `dt` property, rather than grabbing the time directly.   That way you don't even need to worry about whether the game is paused or not.

heniotierra:
I... don't understand what you mean. That is, I know timing is handled based on EnterFrame, but the issue described in that topic is the fact that in some cases the callback function stops being executed due to timeBuffer being reset when it should not. Am I missing something?

starwed:
Oh, my point was that Crafty's game loop already uses a pausebuffer.   If Delay just accumulated time with each frame, rather than checking the current time, it wouldn't need to handle tracking pauses itself.  That would eliminate a lot of the existing (apparently error prone) code.

It would also ensure that events created using Delay were synced correctly with the timings of, say, animations and tweens.

heniotierra:
I see. That would be pretty nice. I guess we can close this PR, then?

starwed:
Let's at least keep it open until there's an alternate PR!  It might turn out that my idea doesn't work for some reason.  :)

mucaho:
We can also merge this functionality temporarily, until a new PR refactors the whole timing in delay.

heniotierra:
It might be worth it in the case this refactoring will take long to come(?).

starwed:
Ok, I did the refactoring in PR #775.  It seems to work fine, but maybe someone should test it out in a real game?

It actually took longer to refactor the tests, since they previously had to do abusive things to the Date object rather than just using simulateFrame.

heniotierra:
Nice!! :D In my tests so far, it's perfect!

mucaho:
Merged #775, so closing this one. @heniotierra  good job finding and fixing those bugs :)

heniotierra:
'Twas my pleasure! (:

Issue #687, opened 2014-01-09
WebGL support


starwed:
I started toying around with adding a WebGL backend on Monday, and it's actually come together pretty quickly.  The idea here is to render sprites using a webgl context rather than the standard canvas.  So it won't draw 3D models, it'll just do what Canvas does, faster.  (Well, hopefully faster!)

I have the basic viewport and 2D capability working -- I can render colored squares using a WebGL component, rotate and translate them, and translate the viewport.

There's still a lot of work left to do, but I thought I should post here about it.  The few next big things to add are
- [x] Support for alpha settings
- [x] Image textures
- [x] Viewport resize
- [x] Viewport zooming
- [x] "Proper" support for webgl built into the graphical components Sprite and Tint.    (Color, Text, and Image could maybe be added later, though Color and Text are pretty tied to DOM stuff that canvas just happens to support.)  (Still need to actually transplant code from TestColor to Tint, but that'll be easy!)
- [x] Add z-sorting, now that transparency is active.

And I'm sure there's lots of room for optimization, since I'm a complete novice at webgl.

I'm working on the [webgl](https://github.com/starwed/Crafty/tree/webgl) branch of my repo.  All the new code is in [webgl.js](https://github.com/starwed/Crafty/blob/webgl/src/webgl.js).  There's a lot of cruft right now, and I'm testing it with a "TestColor" component that lets you set the color with `color: function (r, g, b)`.

kevzettler:
Awesome!

mucaho:
Nice idea. Let me share my thoughts on the current state of your webgl implementation:
[...]

PR #773, opened 2014-04-07
Implement WebGL backend as an alternative to Canvas

starwed:
This implements a WebGL backend as described in issue #687.
- Works with Color and Sprite components; doesn't work for Image or Text.  To try it, just take an existing game that uses sprites and switch them all to `WebGL` instead of `Canvas` or `DOM`.
- Better docs to come!
- Doesn't currently have any tests; hard to test WebGL directly in our automated tests because phantomJS doesn't support it.
- Inlines the vertex/shader programs via brfs, but that leaves some `require("fs")` lines in the source code; hopefully there's a way around that.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -45,6 +45,9 @@ module.exports = function (grunt) {
             dist: {
                 files: {
                     'crafty.js': ['src/crafty.js']
+                },
+                options: {
+                    transform: ['brfs']
                 }
             },
             debug: {
@@ -52,10 +55,12 @@ module.exports = function (grunt) {
                     'crafty.js': ['src/crafty.js']
                 },
                 options: {
-                    debug: true
+                    debug: true,
+                    transform: ['brfs']
                 }
             }
         },
+
 
         watch: {
             files: ['src/*.js'],
@@ -111,6 +116,8 @@ module.exports = function (grunt) {
     grunt.loadNpmTasks('grunt-jsvalidate');
     grunt.loadNpmTasks('grunt-browserify');
     grunt.loadNpmTasks('grunt-banner');
+
+ 
 
     grunt.registerTask('version', 'Takes the version into src/version.js', function() {
         fs.writeFileSync('src/version.js', 'module.exports = "' + version + '";');
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "grunt-contrib-uglify": "^0.5.1",
     "grunt-contrib-watch": "^0.6.1",
     "grunt-jsvalidate": "^0.2.2",
-    "marked": "^0.3.2"
+    "marked": "^0.3.2",
+    "brfs": "1.0.1"
   }
 }
--- a/src/color.js
+++ b/src/color.js
@@ -133,6 +133,18 @@ Crafty.extend({
 
 
 
+// Define some variables required for webgl
+var fs = require('fs');
+var COLOR_VERTEX_SHADER = fs.readFileSync(__dirname + '/shaders/color.vert', 'utf8');
+var COLOR_FRAGMENT_SHADER = fs.readFileSync(__dirname + '/shaders/color.frag', 'utf8');
+var COLOR_ATTRIBUTE_LIST = [
+    {name:"aPosition", width: 2},
+    {name:"aOrientation", width: 3},
+    {name:"aLayer", width:2},
+    {name:"aColor",  width: 4}
+];
+
+
 
 /**@
  * #Color
@@ -149,6 +161,9 @@ Crafty.c("Color", {
 
     init: function () {
         this.bind("Draw", this._drawColor);
+        if (this.has("WebGL")){
+            this._establishShader("Color", COLOR_FRAGMENT_SHADER, COLOR_VERTEX_SHADER, COLOR_ATTRIBUTE_LIST);
+        }
         this.trigger("Invalidate");
     },
 
@@ -169,6 +184,13 @@ Crafty.c("Color", {
         } else if (e.type === "canvas") {
             e.ctx.fillStyle = this._color;
             e.ctx.fillRect(e.pos._x, e.pos._y, e.pos._w, e.pos._h);
+        } else if (e.type === "webgl"){
+            e.program.writeVector("aColor",
+                this._red/255,
+                this._green/255,
+                this._blue/255,
+                this._strength
+            );
         }
     },
 
--- a/src/crafty.js
+++ b/src/crafty.js
@@ -28,6 +28,8 @@ require('./text');
 require('./time');
 require('./version');
 require('./viewport');
+require('./webgl');
+require('./gl-textures');
 
 if(window) window.Crafty = Crafty;
 
--- a/src/drawing.js
+++ b/src/drawing.js
@@ -2,6 +2,18 @@ var Crafty = require('./core.js'),
     document = window.document;
 
 
+// 
+// Define some variables required for webgl
+var fs = require('fs');
+var IMAGE_VERTEX_SHADER = fs.readFileSync(__dirname + '/shaders/sprite.vert', 'utf8');
+var IMAGE_FRAGMENT_SHADER = fs.readFileSync(__dirname + '/shaders/sprite.frag', 'utf8');
+var IMAGE_ATTRIBUTE_LIST = [
+    {name:"aPosition", width: 2},
+    {name:"aOrientation", width: 3},
+    {name:"aLayer", width:2},
+    {name:"aTextureCoord",  width: 2}
+];
+
 /**@
  * #Image
  * @category Graphics
@@ -12,30 +24,15 @@ Crafty.c("Image", {
     ready: false,
 
     init: function () {
-        var draw = function (e) {
-            if (e.type === "canvas") {
-                //skip if no image
-                if (!this.ready || !this._pattern) return;
+        this.bind("Draw", this._drawImage);
+    },
 
-                var context = e.ctx;
-
-                context.fillStyle = this._pattern;
-
-                context.save();
-                context.translate(e.pos._x, e.pos._y);
-                context.fillRect(0, 0, this._w, this._h);
-                context.restore();
-            } else if (e.type === "DOM") {
-                if (this.__image) {
-                  e.style.backgroundImage = "url(" + this.__image + ")";
-                  e.style.backgroundRepeat = this._repeat;
-                }
-            }
-        };
-
-        this.bind("Draw", draw).bind("RemoveComponent", function (id) {
-            if (id === "Image") this.unbind("Draw", draw);
-        });
+    remove: function() {
+        this.unbind("Draw", this._drawImage);
+        // Unregister webgl entities
+        if (this.program) {
+            this.program.unregisterEntity(this);
+        }
     },
 
     /**@
@@ -53,6 +50,8 @@ Crafty.c("Image", {
      * If the width and height are `0` and repeat is set to `no-repeat` the width and
      * height will automatically assume that of the image. This is an
      * easy way to create an image without needing sprites.
+     *
+     * If set to `no-repeat` and given dimensions larger than that of the image, the exact appearance will depend on what renderer (WebGL, DOM, or Canvas) is used.
      *
      * @example
      * Will default to no-repeat. Entity width and height will be set to the images width and height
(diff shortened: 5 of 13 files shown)


starwed:
There are a few bits of cleanup I missed, like removing some logging statements!  :)

starwed:
I took a stab at writing up an overview of the implementation [on the wiki](https://github.com/craftyjs/Crafty/wiki/WebGL-implementation).  Poke me here if there's obvious stuff missing!  :)

starwed:
Oh, for some reason the Travis CI build is failing. It passes locally just fine for me, but looks like something about the brfs plugin is giving Travis problems?

mucaho:
Yeah Travis sometimes fails to install the required npm modules sometimes. You can force Travis to rebuild the last commit (I achieved it by adding a new commit, or maybe force push the latest commit amended with a blank line, but perhaps you can retrigger the build on Travis' website)

> doesn't work for Image or Text

Will they fallback to DOM / Canvas? `Crafty.e("2D, WebGL, Text")`

starwed:
Ah, `package.json` was specifying the incorrect version of brfs.  I already had the right one installed locally, so I never noticed.  Passes Travis right now.

> Will they fallback to DOM / Canvas? Crafty.e("2D, WebGL, Text")

Nope.  Right now we never do fallback for these sorts of things.  (For instance, before I removed it, Tint worked only with Canvas, not DOM.)

We could maybe at least add a `console.log` statement to them?  Or throw an error?

mucaho:
>  Right now we never do fallback for these sorts of things...

The `Text` documentation is pretty clear it's usable with either DOM or Canvas, so I think it can stay like it is.

What about `Image` though? Wouldn't it have a similar implementation to the `Sprite` component?

starwed:
> What about Image though? Wouldn't it have a similar implementation to the Sprite component?

Yes, it would be pretty similar.  I just need to make sure that the same type of texture can work for both Image and Sprite.

mucaho:
> I just need to make sure that the same type of texture can work for both Image and Sprite

Could you elobarotate, please?

starwed:
> Could you elobarotate, please?

The original WebGL tutorials I encountered turned off texture wrapping, which I believe "Image" needs to function equivalently to its DOM/Canvas counterparts.  i have no idea if this effects performance or anything like that!  So I need to investigate.  :)

mucaho:
>  "Image" needs to function equivalently to its DOM/Canvas counterparts ... it needs texture wrapping turned on .... does it effect performance or anything like that!

Perhaps you know that already, but maybe it helps: 
Texture coordinates are in the range from `0` to `1`. If you make a sampler lookup that is outside of this range, the [texture wrap mode](http://open.gl/textures#Wrapping) will specify how the out-of-range coordinate will be adapted -> the newly adapted texture coordinate will be used to make the sampler lookup. Speculation: performance-wise it's the same no matter where in the texture you lookup the pixel.

`GL_REPEAT` sounds like what you want, but you still have to figure out how to calculate the out-of-range coordinates exactly, like you said.

starwed:
I've implemented WebGL for the Image component now.  There will be some weirdness if you change the size of an entity after creating it, but the Image component is already pretty inconsistent here; I plan on fixing that as a separate patch.  (It's not even clear to me what the behavior should be!)

I also split the PR into logical parts.  :)

starwed:
Oh, and if anyone wants to try the webgl branch out without having to pull+build, [here is the built file.](http://ludonomicon.com/crafty/crafty-webgl.js)

gregtillbrook:
Im working on a space shooter style game atm and have had concerned about the max number of entities I could have on screen, so this webgl backend has been of great interest to me. Ive done some basic tests and results have been really good, although I did have one weird issue (which I'll detail below).

### The results

My test was a simple scene containing randomly placed spinning asteroids. I measured now many roids I could get on screen before framerate dipped below 30fps;

| Platform | canvas | webkit | Delta |
| --- | --: | --: | --: |
| Mac Pro Firefox 29 | 800 | 2000 | +1200 |
| Mac Pro Chrome34 | 1800 | 6500 | +4700 |
| Mac Pro NodeWebkit 0.9.2 | 3500 | 7000 | +3500 |
| Win7 PC Chrome35 | 1100 | 5300* | +4200 |
| Win7 PC NodeWebkit 0.9.2 | 2200 | 5600* | +3400 |

Im making a desktop game using NodeWebkit as a wrapper, hence since Id expect most of my players to be windows based the last result in the table is of most interest to me.

### \* The issue

The one odd issue I hit was that when using WebGL on the windows PC I could never get the asteroids to rotate (i.e. in chrome browser or compiled NodeWebkit exe). Any ideas what may be causing this? Note the same code works fine on my MacPro + the Windows PC is able to render WebGL from other websites fine.

Is there something wrong with my code?

``` Javascript
$(document).ready(function() {
    Crafty.init();
    Crafty.webgl.init();

    Crafty.load(["images/bg.png", "images/asteroid.png"], function() {
        Crafty.sprite(16, "images/asteroid.png", { 
            roidSprite: [0,0] //conforms to pow2 requirement of webgl textures
        });
        Crafty.scene("perf");
    });

    Crafty.scene("perf", function(){
        Crafty.background("url('images/bg.png')");
        Crafty.e('roid');
    }); 

    Crafty.c("roid", {
        init: function() {
            this.addComponent('2D, WebGL, roidSprite')
                .attr({
                    x:Crafty.math.randomInt(0, Crafty.viewport.width),
                    y:Crafty.math.randomInt(0, Crafty.viewport.height)
                })
                .origin("center")
                .bind("EnterFrame", function(params){
                    this.rotation += params.dt * 0.01;
                });
        },
    });
});
```

gregtillbrook:
There also seems to be a confusion between rotation needing to be in degrees or radians atm. i.e. WebGL needs rotation to be set as radians while Crafty still needs degrees for hit box/children position etc.

gregtillbrook:
Is this feedback useful?

mucaho:
Yeah, thank you Greg, we will take a look at it as soon as we have time
(I'm a bit busy at work now).

starwed:
Sorry for the delay on this; I'd been a bit busy with other stuff.  I need to properly investigate the rotation issue mentioned, and check a problem with the "Image" component not working; then this is probably good to land.

> There also seems to be a confusion between rotation needing to be in degrees or radians atm. i.e. WebGL needs rotation to be set as radians while Crafty still needs degrees for hit box/children position etc. 

Yeah, I think we should convert units before setting the rotation attribute.? (Or we could convert in the shader; that might be faster but could also be more confusing to devs...)  I have a test file that renders the same stuff using DOM, Canvas, and WebGL -- I'll include rotation in it to catch this type of issue!

Rotation was definitely working in general, last time I checked.  (I did a very similar test of 100s of sprites rotating to confirm that WebGL is faster!)

heniotierra:
Hi guys!

I'm testing this implementation now. I'll briefly explain my misadventures up til this point. 

[EDIT]: Nevermind the strikethroughs, only now I saw the tip in the fisrt post about Image and Text not being supported. [EDIT END]

First I tried to run a scene which contains a good-sized tilemap, and right away faced an error that stated "Number of entities exceeds maximum limit."; I went to investigate that, and found this: `this.max_size = 300;` which I then changed to `this.max_size = 5000;`, and that worked, the scene was loaded with just some minor problems related to layering, which is probably caused by that spritesheet 'nuisance' @starwed mentioned, which I didn't yet have time to take care of; pretty good performance.

Then I went on to try another scene which also has a tilemap, but much bigger, and many background images (once 2 frigging big background images I split into 40 pieces, 20 each), but the same problem that happened in the previous test happened here too, so I tried to make `this.max_size = 60000;`, but that crashed firefox immediately, then I tried other values, and got it to work with `this.max_size = 30000;`~~, but only work partially since the background images - which use the Image component - didn't appear (the areas where they should be at appear as black) and some other entities which have sprite animations would appear as black squares; I also get messages which I think may be related to the browser not showing the images properly: firts I get messages about successfully loaded textures ("Texture dimensions"), but then, after the 31st texture has been loaded (from 46 total), the console shows a lot of warnings saying "WebGL: INVALID_ENUM: activeTexture: texture unit out of range"; performance was very iffy, but I'd say still acceptable. Maybe I'm trying to take too much out of this, using such huge maps?~~ [EDIT]: After changing back entities with Image component to use Canvas this scene worked fine, except that many of the tiles from the tile map don't show up at all; good performance in chrome, bad performance in firefox. [EDIT END]

Oh, and the tile maps sometimes show up with lines in between tiles.

Then I went on to try yet another scene, this one without a tilemap, but with some entities with sprite animations and other ones with texts, and the sprite animation entities all worked fine in this scene, ~~but none of the text entities appeared; about the texts, I didn't really dug it up to see how this 3D implementation handles it, so it's probably my own fault~~; working fine, pretty good performance.

---

These tests were made with all of these entities set to use WebGL. I'm on Linux Mint 17, using Firefox 31 and Chrome 36.

And then I have some questions: Can this max_size thing be messed with like that? And why such a small limit as 300 entities? And how can this value be changed without directly changing the engine code?

[EDIT]: I have yet to make tests in a mobile. I'll come back later and make a better report. [EDIT END]

starwed:
@heniotierra The max size of 300 was a compromise at the time.  Because it allocates a fixed size buffer, increasing the limit will end up using more memory.

There are several different ways to improve this behavior, but right now I just want to finally land the dang thing.  :)

starwed:
Worked on this a bit, and I think it's actually ready to land quite soon!  When I took games/unfinished toys I had lying around, swapping DOM or Canvas for WebGL worked just fine.  You can check out a build [here](http://ludonomicon.com/crafty/crafty-webgl.js).  (I've mostly tested on Firefox, so there could easily be some nasty browser-specific behavior.)
- I fixed a stupid bug that caused entities created mid-scene to not be displayed.  (The viewport variables weren't being set.)
- It now dynamically changes the size of the entity buffers based on use.  
- It also juggles textures a bit better if you use a lot, rebinding them as necessary if you use more than the max amount.  (I _think_ that's a performance hit, though?)  If you've actually been looking at the code, this involved quite a bit of refactoring/abstracting.

I'll try to put together a demo to make it easier to check that all the functionality works on multiple browsers.

kevinsimper:
@starwed I tried switching my snake demo to webgl, but it does not work and no error is thrown. http://kevinsimper.github.io/craftyjs-snake/

kevinsimper:
You should add this to your package.json so if you require your branch you can run browserify from the commandline.

```
"browserify": {
  "transform": ["brfs"]
 },
```

kevinsimper:
I found out that I had included the components in the wrong order.
One solution would be to `this.require` inside webgl component or/and throw and error.

starwed:
> I found out that I had included the components in the wrong order.

If you mean that you did something like `Crafty.e("2D, Color, WebGL")`, there's not really a nice way around this right now.  I have a vague idea of how to handle that sort of thing in a general way, but it would probably be part of a new PR.  :)

starwed:
I've now tested it across several browsers and on mobile, and it seems to work fine.  It would be nice if some others could check the rendering, especially on IE and iOS.

Here are a couple of simple demos to check out:
- A [simple game](http://ludonomicon.com/crafty/simple-jumper.html) that renders using WebGL
- A [demo](http://ludonomicon.com/crafty/triple-render.html) that renders the same scene simultaneously using WebGL, DOM, and Canvas.  This was a nice way to check that WebGL rendered identically to our existing systems.

sbyrnes321:
Looks great to me!

I have a feature request: Add something like `Crafty.support.webgl` so that developers can give a sensible error message to users without webgl. (This is still pretty common, e.g. older cell phone browsers.) I don't think the code to do this is too hard ... for example view-source at http://doesmybrowsersupportwebgl.com/ 

Even without that, this is very good and should be merged. :-D

starwed:
> I have a feature request: Add something like Crafty.support.webgl so that developers can give a sensible error message to users without webgl.

Oddly, we were [already testing for this](https://github.com/craftyjs/Crafty/pull/773/files#diff-5).  I did update the check, though.

sbyrnes321:
I tried those two demos on windows. Bad news... :-(

In both firefox and chrome, the character doesn't rotate. (But she does flip.) Otherwise it works fine.

screenshot 2014-12-08 11 47 22

In IE 11, it just doesn't work at all. It says that `gl` is undefined in gl-textures line 156.
screenshot 2014-12-08 11 53 25

starwed:
Thanks for testing it out.

> In both firefox and chrome, the character doesn't rotate. (But she does flip.) Otherwise it works fine.

That's pretty weird -- I would have expected that if it rendered at all, it would render the same way.  Especially in Firefox/Chrome.

> In IE 11, it just doesn't work at all. It says that gl is undefined in gl-textures line 156.

I'm probably using some function, somewhere, that isn't supported by IE.  (Maybe some of the typed array stuff?)  

I'll look more carefully at both of these, but it's a bit annoying since I don't have a windows machine to test on.

starwed:
I was able to test on a windows box on campus, which reveleaed the same problems @sbyrnes321 noticed.

The issue with IE11 was just that I was using `Math.log2`, which isn't supported.  Luckily that's an easy fix.

Once I patched that, IE11 had the same issue as Firefox and Chrome -- no rotation.  That seems really, really weird, and I'm not sure what could be causing it!  Maybe something about the shader program which is slightly incorrect, but that only manifests on windows?

starwed:
Using Firefox's shader editor on windows, it appeared that there's some problem setting the attribute `aOrientation` in the shader programs.  If I simply replaced it with hardcoded values, I got a `Color` entity to correctly render rotated.  I suspect those values are somehow ending up as zero on windows.

I'll try to walk through the code and pinpoint places that could happen.  (TypedArrays are something that can apparently have differing implementation details on different machines, so perhaps the issue is there?  It could also be something to do with fetching the pointers to the correct attributes...  but `aExtra` and `aColor` seem to be working ok.)

starwed:
@sbyrnes321  Ok, I think I've figured out the windows bug!  Here are the demos again, with the fix:
- A [simple game](http://ludonomicon.com/crafty/simple-jumper.html) that renders using WebGL
- A [demo](http://ludonomicon.com/crafty/triple-render.html) that renders the same scene simultaneously using WebGL, DOM, and Canvas.

The issue was kind of weird, and lay in the vertex shaders.  If you don't use an attribute, OpenGL will feel free to optimize the program by simply not even allocating memory for that attribute.  Because I used the `aOrientation` attribute  only _outside_ of the `main()` function of the shader, OpenGL apparently was technically free to perform this optimization.  But, I guess only the windows drivers bothered to do so, which is why it worked on Mac/Android/etc.

heniotierra:
Hi @starwed , I tested with the lattest version of the game I was working on (finally launched, amen, though not hardware accelerated due to bugs) and found no errors except for one level where I get 

> Uncaught Number of entities exceeds maximum limit. 

I fixed the bug about log2 by using Math.LN2 instead of Math.log2... and solved the problem of loading the shaders by embedding them in js as string and passing them to que global arrays. A Found no errors except

starwed:
> I fixed the bug about log2 by using Math.LN2 instead of Math.log2.

Yup, the version I'm about to push has the same fix.

> Uncaught Number of entities exceeds maximum limit. 

You should only get this error when the number of alive entities using the same render program exceeds 1024.  (All sprites using the same base texture will use the same program.)  That seemed like a lot to me, but I can certainly up it to something higher.  (Or allow it to be more easily set by the user.)

(If you want to keep lots of entities alive without actually displaying them, you could always use `program.unregisterEntity(e)`.)

heniotierra:
Yeah, I know... but that only works for me if I put it around 16000, and then the game won't break, but some enities won't show up.

Now a little issue I found is that the images are show slighting different. I mean, line around the images (not the frame lines) are a like blurred... But that's not really a big problem.
Here, you can compare:
Canvas
https://drive.google.com/file/d/0B_OhLP63F8vZNWlNODZsellISTQ/view?usp=sharing
WebGL
https://drive.google.com/file/d/0B_OhLP63F8vZV25YYWYzYXVCaG8/view?usp=sharing

starwed:
> Now a little issue I found is that the images are show slighting different.

(Those images look identical to me, and are timestamped just seconds apart...)  Are you using the `pixelart` setting?  I know there are a couple browser bugs that prevent that working in some contexts, I've seen these two at least: [chrome](https://code.google.com/p/chromium/issues/detail?id=134040) and [firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=696630).   (The chrome one got marked fixed just a day ago.)

> 16000, and then the game won't break, but some enities won't show up.

That seems really, really high, and you're presumably running against the actual limits of the hardware.  (There's some maximum size for the buffer used to store entity data.) Are there really that many entities on the screen at once?  It could be there's some bug that's preventing old entities from being destroyed or something.

heniotierra:
Well, I tried with pixelart on and off, and the problem seems to be in both cases.

heniotierra:
Well, yes, there's a fucking load of entites (it's a level with a reeeally big tilemap with 32x32 sized tiles). I'll study trying program.unregisterEntity(e) later.

starwed:
I've personally tested this on:
- Windows: Firefox, Chrome, IE
- OSX: Firefox, Chrome, Safari
- Android (GNex and Nexus 7): Firefox, Chrome
- iOS: None.  Turns out everyone I know with an iOS device is running v7, which doesn't support WebGL.

In any case, I'm just going to go ahead and finally land this!  It'll be easy enough to deal with issues as they come up, and while there are several things that can be improved, at some point it's important to just integrate the base functionality.

mucaho:
:+1:

PR #774, opened 2014-04-09
Fix for propogating 90deg rotations to child entities

FallSe7en:
Hi,

I noticed what I'm pretty sure is a bug in Crafty when rotating an entity by 90 degrees; the entity rotates fine, but the rotation doesn't get propagated to its children.

Here is a fiddle demonstrating the bug: http://jsfiddle.net/f8c4c/7/

Here's my suggested fix and an accompanying test. Strange thing though--when I revert my fix and run the test I wrote in grunt, it passes. Opening up the index.html in a browser, however, makes the test fail as expected (and pass with the fix applied). I'm new to grunt, so maybe I'm just missing something..

Anyways, hope this helps!
Justin

### DIFF
--- a/src/2D.js
+++ b/src/2D.js
@@ -699,7 +699,7 @@ Crafty.c("2D", {
             l = children.length,
             obj;
         //rotation
-        if (e.cos) {
+        if (("cos" in e) || ("sin" in e)) {
             for (; i < l; ++i) {
                 obj = children[i];
                 if ('rotate' in obj) obj.rotate(e);
--- a/tests/2d.js
+++ b/tests/2d.js
@@ -224,7 +224,32 @@
     parent.rotation = 100; // Rotation by 90 degrees from initial position
     strictEqual(Round(child.x), -10, "Child moved around parent upon rotation (x).");
     strictEqual(Round(child.y), 10, "Child moved around parent upon rotation (y).");
+  });
 
+  test("child rotate 90deg", function () {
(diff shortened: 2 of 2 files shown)


sbyrnes321:
Nice!!

But I suggest to change it instead to `if (e.cos !== undefined)` or `if ('cos' in e)` or something like that.

I think that would make it easier for people to understand what the code is doing. What you did is fine too, just a bit confusing for people who are not familiar with the code, I think.

I bet that `e.cos === 0` in the browser, and `e.cos === 3.602852987e-17` or whatever in grunt. A random floating-point-arithmetic discrepancy.

starwed:
> I bet that `e.cos === 0` in the browser, and `e.cos === 3.602852987e-17` or whatever in grunt.

This has come up a couple times before with Chrome specifically -- at some point they slightly changed how they round these results.

starwed:
@FallSe7en  Thanks for the PR!  The fix looks good, but could you remove the "Color" and "DOM" components from the test entities?  We potentially want to be able to run the 2D unit tests in an environment that doesn't include the DOM (see #603), and I don't think the test actually depends on them.

(Squash things down to a single commit when you're done; see [here](https://github.com/craftyjs/Crafty/wiki/Rewriting-History) for more info if you're unsure.)

FallSe7en:
Alright, changes made!

starwed:
Thanks!

PR #775, opened 2014-04-09
Refactor "Delay" to use dt from frameData

starwed:
Rather than explicitly requesting the current time, the "Delay" component should just accumulate the time passed in by the "EnterFrame" event.  This way it doesn't need to worry about tracking the pause state of the game; that's all handled by the main game loop.

This meant drastically simplifying the tests -- they no longer need to overwrite the Date object or anything like that.  They can just use `Crafty.timer.simulateFrame` like we do elsewhere.

This is an alternative to #772.

### DIFF
--- a/src/time.js
+++ b/src/time.js
@@ -8,24 +8,23 @@ var Crafty = require('./core.js'),
 Crafty.c("Delay", {
     init: function () {
         this._delays = [];
-        this.bind("EnterFrame", function () {
-            var now = new Date().getTime();
+        this.bind("EnterFrame", function (frameData) {
             var index = this._delays.length;
             while (--index >= 0) {
                 var item = this._delays[index];
                 if (item === false) {
                     // remove canceled item from array
                     this._delays.splice(index, 1);
-                } else if (item.start + item.delay + item.pause < now) {
-                    item.callback.call(this);
-                    if (item.repeat > 0) {
-                        // reschedule item
-                        item.start = now;
-                        item.pause = 0;
-                        item.pauseBuffer = 0;
+                } else {
+                    item.accumulator+=frameData.dt;
+                    // The while loop handles the (pathological) case where dt>delay
+                    while(item.accumulator >= item.delay && item.repeat >= 0){
+                        item.accumulator -= item.delay;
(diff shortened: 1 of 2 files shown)


src/time.js
```diff
@@ -8,24 +8,23 @@ var Crafty = require('./core.js'),
...
-                    item.callback.call(this);
-                    if (item.repeat > 0) {
-                        // reschedule item
-                        item.start = now;
-                        item.pause = 0;
-                        item.pauseBuffer = 0;
+                } else {
+                    item.accumulator+=frameData.dt;
+                    // The while loop handles the (pathological) case where dt>delay
+                    while(item.accumulator >= item.delay){
```
mucaho: ~~the `while` block looks fishy to me~~
~~what does the comment mean?~~
Ah know I get it: you are worried about the case where `dt > 2*delay` (let's say for delay like 10ms or less). In that case the `while` loop will help, but ultimately the number of repetitions will be wrong (as the check against the number of repetitions is after the `while` which may trigger the callback too often).

To deal with this corner case correctly (even if it's unlikely to happen), I suggest we move the repetition check inside of the `while` loop somehow:

``` javascript
                    item.accumulator+=frameData.dt;
                    // The while loop handles the (pathological) case where dt > 2*delay
                    while(item.accumulator >= item.delay && item.repeat >= 0){
                        item.accumulator -= item.delay;
                        item.repeat--;
                        item.callback.call(this);
                    }
                    // remove finished item from array
                    if (item.repeat < 0){
                        this._delays.splice(index, 1);
                        if(typeof item.callbackOff === "function")
                            item.callbackOff.call(this);
                    }
```

And maybe add a test case with 5 ms delay and 1 repetition, triggering simulateFrame with 20ms.

starwed: > In that case the while loop will help, but ultimately the number of repetitions will be wrong (as the check against the number of repetitions is after the while which may trigger the callback too often).

Ah, nice catch.

starwed:
Ok, updated with suggested change, and added a test.

mucaho:
Great, looks good, so merging

PR #777, opened 2014-04-11
Refactored loader, and created new function 'removeAssets'

heniotierra:
Refactored `Crafty.load`, making it recieve an object with assets definitions, including audio, images and sprites.
Also, created new function `Crafty.removeAssets`, which receives an object similar to the one that's passed to `Crafty.load`, or even the exact same object, in which case it will delete all assets preloaded in that way.


starwed:
From a quick look, this seems like a good approach.  Couple things:
- It might be better to have the removeAsset function trigger a "RemoveAsset" event, and allow components do destroy themselves.  That way we wouldn't need to hardcode the details about which components store images/etc.
- The path variables should probably be private, and then we could have a `Crafty.setPaths` function that takes an object like `{audio:"/sounds", images:"/pics"}` to set them all at once.
- We probably don't need to actually loop over types and then use a series of if statements (`for (var type in data) ...`); just handle each of the three cases separately?  Slightly easier to see what's going on.

Something that I don't think the current code does, but that we should support:
- if the url starts with `protocol://` we shouldn't append the folder path to it.

These are just some thoughts about the design; I haven't tested the code yet, though it looks ok from reading over it.

starwed:
This is failing Travis CI, though, so you might want to figure out which test is going fubar.

heniotierra:
> removeAsset function trigger a "RemoveAsset"

Hum, I'll try that.

> The path variables should probably be private, and then we could have a Crafty.setPaths function that takes an object like {audio:"/sounds", images:"/pics"}

What if one wants no "assets" folder, like, from main page, path/to/images/ ?

> just handle each of the three cases separately?

I thought of doing that, but decided not to so that everything would be inside the "same" loop(not really however). But yeah, doing it separately would make it easier to read.

> I haven't tested the code yet, though it looks ok from reading over it.
> 
> This is failing Travis CI, though, so you might want to figure out which test is going fubar.

The code works, I guess it fails the test because I didn't touch the tests files...

Thanks for your comments!

starwed:
> What if one wants no "assets" folder, like, from main page, path/to/images/ ?

Sorry for being unclear, in the example those would have been setting the whole path; it wouldn't default to including assets.  :)  

> I thought of doing that, but decided not to so that everything would be inside the "same" loop(not really however). But yeah, doing it separately would make it easier to read.

Hmm, that's true.  Just go with whichever you think is clearer, then!

starwed:
Unsurprisingly, `tests/loader.js` will need some changes before the build passes.  :)

Might be good to have tests for the new functionality, too.

heniotierra:
I fixed the test thing, but will wait for doing the other stuff before sending it. Will also try to add a test for `removeAssets`. 

But there's one thing I just found: `Crafty.audio.remove` does indeed have a bug. While it successfully deletes the sounds (in `Crafty.audio.sounds`), it fails to delete the assets because it uses as reference the full file path (from `s.obj.src`), but the assets object has relative paths as references. ~~Maybe we should pass a second argument to `Crafty.audio.remove` with relative file path?~~ Passing a second argument wouldn't be the best thing to do, since even the first argument is optional. :-/ Have an idea!

EDIT: Fixed this bug.

> path variables should probably be private

By private do you mean using prototyping and stuff, or just like Crafty._paths.audio Crafty._paths.images, and then using the methods to set/get them? For now I'm going for the later.

EDIT2: 

> if the url starts with protocol:// we shouldn't append the folder path to it.

Now the code supports this, for http://, https:// and file:// protocols. I tried doing that in one line with a regular expression, but I'm no good with regex, and the ones I found looking around didn't work. But then again, I was looking for a regex that'd work for any protocol.
Anyway, here's the test (self executing) function I wrote that's responsible for checking if there a protocol in the file definition, where f is the file. That's inside a getFilePath function.

```
var test = (function() {
        var protocols = ['http://','https://','file://'], 
            res = false;
        for (var i = 0, l = protocols.length;i<l;i++)
            if (f.search(protocols[i]) === 0) {
                res = true;
                break;
            }
        return res; 
    })()
```

Do you think that's too "dirty"?

heniotierra:
I think I did everything you pointed out, except for this:

> It might be better to have the removeAsset function trigger a "RemoveAsset" event, and allow components do destroy themselves. That way we wouldn't need to hardcode the details about which components store images/etc.

I was unable to do that, since I didn't find how to make the components listen to an event. An entity with that component can "listen", but can the component object _en si_ do that too?

starwed:
> By private do you mean using prototyping and stuff, or just like Crafty._paths.audio Crafty._paths.images, and then using the methods to set/get them? For now I'm going for the later.

Yup, that's what I was thinking of.  

> Now the code supports this, for http://, https:// and file:// protocols. I tried doing that in one line with a regular expression, but I'm no good with regex, and the ones I found looking around didn't work. But then again, I was looking for a regex that'd work for any protocol.

I'll have a look at coming up with a regex, because that would be a lot cleaner.
_e:_ I think just checking for whether the string `://` is present would be good enough -- no one would have that in a relative file path!

> I was unable to do that, since I didn't find how to make the components listen to an event. An entity with that component can "listen", but can the component object _en si_ do that too?

I was thinking that the init function of the component should bind a function to that event.

heniotierra:
> I think just checking for whether the string :// is present would be good enough

Done. As clean as possible.

> I was thinking that the init function of the component should bind a function to that event.

Done. :+1: 

And I think that's all? :)

heniotierra:
Fixed some errors in the docs, and made some improvements to them too.

heniotierra:
Now I was able to make the loader loop cleaner('if else' instead of 'if else if else if'), and easier to read, I think, as well as removeAssets's. Please, have a look. Also more changes to documentation: Added more info to the function descriptions, and made image_whitelist example 'work' for new Crafty.load.

heniotierra:
Now I removed the changes made to sprite.js, cos they were not necessary, and components are being removed without touching it.

starwed:
Hmm, deleting the components without destroying related entities first could cause some weird issues in the future.  Some components rely on the special `remove` function to cleanup after themselves.  This isn't a property of an entity, but the component itself.  (Same as with `init`.)  If you delete a component, then the remove function will never be called on entity destruction.  This isn't something the named-sprite components do right now, but it easily could be in the future!  :)

It of course also won't actually free the resources just yet; the components need to be destroyed first, since they each hold references.  And custom user created components might also want to hold on to references to assets; this is one reason I was suggesting a "RemoveAsset" event.

Thanks for all the work!  I'll go ahead and merge this, but I'll likely work on a followup patch that changes how `removeAssets` works.

heniotierra:
Thank you too! :wine_glass: 

> If you delete a component, then the remove function will never be called on entity destruction.

Oh, didn't know that.. Anyway, it's awesome that you will implement that fix!

sbyrnes321:
Just a reminder: We need to emphasize in the release notes that this breaks back-compatibility.

For example, I had to change this:

```
Crafty.load(['foo.png', 'bar.png']);
```

to this:

```
Crafty.paths({'images': './' });
Crafty.load({'images': ['foo.png', 'bar.png']});
```

(It's a nice change though.)

PR #778, opened 2014-04-12
Add audio_isPlaying.

mucaho:


starwed:
Pretty simple change, so merging.  Thanks again!

Pity we can't have (automated) tests that use audio.

PR #785, opened 2014-05-12
Update sound.js, Crafty.audio.add not work like example says 

XiongLiding:
add type detect for `id[i]`, otherwise code like `coin: "sounds/coin.mp3` in example won't work! but only `coin: ["sounds/coin.mp3"]` works

### DIFF
--- a/src/sound.js
+++ b/src/sound.js
@@ -166,7 +166,7 @@ Crafty.extend({
             if (arguments.length === 1 && typeof id === "object") {
                 for (var i in id) {
                     if (typeof id[i] === "string") {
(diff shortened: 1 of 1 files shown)


kevinsimper:
Good catch! :+1: 

I do think through that we should change the documentation instead, because you can already use `Crafty.audio.add("jump", "sounds/jump.mp3");`, where you can pass it as a string, also considered that you call also instead of a string in the object just pass it as a array, as you showed. Better keep it simple :)

Anybody else though on this?

kevinsimper:
I removed the wrong documentation in #800 

Hope that is okay @surmind

Issue #793, opened 2014-07-09
Pressing direction keys simultaneously causes fourway to move multiple times


cwgorman:
If you press, for example, `W` and `Up Arrow` at the same time with something that uses `fourway`, it moves twice as fast. With `Z` and `Q` you can move three times as fast.

I do not expect this behavior. Instead, it should move at the same speed as if only one of those keys had been pressed.

kevinsimper:
Closing this as we are discussing in #794

PR #794, opened 2014-07-09
Have multiway keep track of the number of keys pressed for each direction

cwgorman:
So if W and UP_ARROW both correspond to the same direction, pressing both at the same time will have no effect, instead of doubling the speed. Resolves issue #793.

### DIFF
--- a/src/controls.js
+++ b/src/controls.js
@@ -706,18 +706,32 @@ Crafty.c("Multiway", {
     _speed: 3,
 
     _keydown: function (e) {
-        if (this._keys[e.key]) {
-            this._movement.x = Math.round((this._movement.x + this._keys[e.key].x) * 1000) / 1000;
-            this._movement.y = Math.round((this._movement.y + this._keys[e.key].y) * 1000) / 1000;
-            this.trigger('NewDirection', this._movement);
+        var dir = '' + this._keyDirection[e.key];
+        if (dir) {
+            var n = this._activeDirections[dir] || 0;
+            if (!n) { //This is the first keydown for this direction
+                if (this._directionSpeed[dir]) {
+                    this._movement.x = Math.round((this._movement.x + this._directionSpeed[dir].x) * 1000) / 1000;
+                    this._movement.y = Math.round((this._movement.y + this._directionSpeed[dir].y) * 1000) / 1000;
+                    this.trigger('NewDirection', this._movement);
+                }
+            }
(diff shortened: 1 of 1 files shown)


kevinsimper:
Hey @cwgorman 

You are right, that is not the intended behaviour. That should be fixed!

I looked at the code and have some comments:

You should avoid creating new variables since that will force the garbace collecter to work and we will want to avoid that.
https://github.com/craftyjs/Crafty/pull/794/files#diff-006f862972d60d6e4ffefc1397f44328R711
https://github.com/craftyjs/Crafty/pull/794/files#diff-006f862972d60d6e4ffefc1397f44328R725

Is type conversion also needed here? (Have not runned the code myself, so can not immediately tell)
`'' + this._keyDirection[k]]` and `var dir = '' + this._keyDirection[e.key];`

cwgorman:
@kevinsimper I've made the changes you've requested. Sorry it took so long; I haven't had much time for personal work lately.

I am curious about the garbage collection though. Is it actually slower than performing the additional lookups? I assumed the JS engines in today's browsers would be smart enough to optimize that.

Chaosed0:
I ran into this one over the last weekend. Patch looks good to my unexperienced eye - anyone else still looking at this?

mucaho:
How will this code perform in the following hypothetical case:  
Imagine a space shooter where `Q` will move the player 1 square left and 1 square up (direction of `-90 - 45 == -135`), and `E` will move the player 1 square right and 1 square up (direction of `0 - 45 == -45`). Now if `Q` and `E` are pressed together, player should just move 1 square up.

Maybe we could split the `activeDirections` map into `activeDirectionsLeft`, `activeDirectionsRight`, `activeDirectionsUp` and `activeDirectionsDown`; then increase/decrease the `activeDirection__` counters by partitioning the key direction on an incoming key event:
- (`-180 < direction < -90` `||` `90 < direction < 180` ) => LEFT
- (`-0 < direction < -90` `||` `0 < direction < 90`) => RIGHT
- (`-0 < direction < -180`) => UP
- (`0 < direction < 180`) => DOWN

This would solve the space shooter example, but still isn't precise enough for non-symmetric keys, but I think even this is overkill.

MakerJake:
No more sending me emails please.

mucaho:
ty cwgorman the proposed changed was integrated alongside other improvements to `Multiway` in  #876

PR #798, opened 2014-07-15
Make a main file that requires all the files

kevinsimper:
This makes it possible to require crafty without compiling it first.
Also by defining main in package.json require can automatically
pick it up from the root folder if you are using browserify.


kevinsimper:
Does anybody else have a opponion about this else I will go ahead and merge this, because I do think this will be on the path of making Crafty more organized.

PR #805, opened 2014-07-23
Made textColor work again with new Crafty.assignColor - fixed #804

### LINKED ISSUE #804: Text code calls Crafty.toRGB() which doesn't exist anymore
The function `Crafty.toRGB()` is called here:

https://github.com/craftyjs/Crafty/blob/develop/src/text.js#L173

but the function doesn't exist anymore. It was deleted in this pull request: https://github.com/craftyjs/Crafty/pull/770

Thanks in advance!

kevinsimper:
Fixed the textColor function mentioned in #804

Made a test which would have caught the error.

### DIFF
--- a/src/text.js
+++ b/src/text.js
@@ -152,11 +152,12 @@ Crafty.c("Text", {
     /**@
      * #.textColor
      * @comp Text
-     * @sign public this .textColor(String color, Number strength)
-     * @param color - The color in hexadecimal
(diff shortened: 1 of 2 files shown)


sbyrnes321:
Cool, my code works again after that change! Thanks!

kevinsimper:
Cool, I will then merge it later tommorow.

PR #807, opened 2014-07-24
Use jsdoc for generating documentation

kevinsimper:
I am using jsdoc for creating documentation.

It can be installed via npm by `npm install jsdoc``

I then run this command `jsdoc src/color.js src/text.js src/time.js` because I have only changed the 3 files.

I then get a finished html site like this:

http://kevinsimper.github.io/craftyjsdoc/Components.Color.html

It would be pretty fast to implement.

And you can get the JSON by doing

`jsdoc src/color.js src/text.js src/time.js -X < MYJSON:json`

### DIFF
--- a/src/color.js
+++ b/src/color.js
@@ -131,6 +131,10 @@ Crafty.extend({
 
 
 
+/**@
+ * My Components
+ * @namespace Components
+ */
 
 
 
@@ -138,6 +142,8 @@ Crafty.extend({
  * #Color
  * @category Graphics
  * Draw a colored rectangle.
+ * @namespace Color
+ * @memberof! Components
  */
 Crafty.c("Color", {
     _red: 0,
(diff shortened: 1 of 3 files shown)


starwed:
A related option is perhaps using [dox](https://github.com/tj/dox), which parses the JSDoc tags, but handles markdown and outputs the json by default.

We'd definitely have to change how we markup some of the documentation to switch to a tool like that.  But that's probably something that could be automated pretty easily.

starwed:
I've been tinkering with the docs, and I now have a working (but not polished or complete) setup:
- I rewrote our custom parser to spit out json instead of html.  It's not that I'm opposed to jsdocs, it just was easier to start from scratch instead of meshing it with our particular doc style, which is _close_ to what jsdocs expects but not a direct match.
- I have a toy script for [React](http://facebook.github.io/react/docs/getting-started.html) that takes the json generated by the previous script and uses it to generate HTML.  I haven't used it before, but it seems like a very straightforward way of going from json --> actual documentation pages.  And it's pretty flexible in terms of whether you generate on the fly in the client, or ahead of time using npm, so that might be a good fit for us.

Just a heads up that I'm working on this.

PR #809, opened 2014-07-30
Optimize entity and addComponent function to create fewer objects

kevinsimper:
We should avoid creating objects that will be garbage collected,
because this slows downs performances and especially if you
create a lot of entities.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -220,45 +220,35 @@ Crafty.fn = Crafty.prototype = {
      * ~~~
      */
     addComponent: function (id) {
-        var uninit = [],
-            c = 0,
-            ul, //array of components to init
-            i = 0,
-            l, comps, comp;
+        var comps = [],
+            c = 0;
 
         //add multiple arguments
         if (arguments.length > 1) {
-            l = arguments.length;
-            for (; i < l; i++) {
-                uninit.push(arguments[i]);
+            var i = 0;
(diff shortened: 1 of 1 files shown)


kevzettler:
This is a side note, but related to performance. Probably a bigger crafty optimization discussion. In general there are a bunch of techniques that you can use to make sure your Javascipt engine's JIT compiler will optimize your code. The general idea is that smaller functions are easier for the JIT to inline. 

I bring this topic up because looking at:
https://github.com/craftyjs/Crafty/pull/809/files#diff-7eb52b366866677666470e019283c8eaR230
One of the key things that keeps the JIT from inlining a function is manipulating the arguments array. 
I wouldn't let that stop this PR but something to consider in the future there might be some JIT techniques to really get some speed out of craftyjs

kevinsimper:
That sounds really cool, I have not read about that yet! :+1: 

But what would you change about that line? How would it become better? :) Line 230 is just a array push?

kevzettler:
You don't have to change that specific line. But that line will de-optimize the rest of the function. So you could move the rest of the `if` `else` logic in to a separate function that will get optimized and keep this function isolated to the arguments manipulation. lodash is an example library that uses some JIT techniques.

kevinsimper:
Can not figure out a better how to improve it, but any pull request to optimize it further is appreciated!

Issue #764, opened 2014-03-25
Crafty.load fails with duplicate image


ashes999:
Tested in Crafty 0.6.2 on Ubuntu/Chrome.

Simplest case:

`Crafty.load(['image.png', 'image.png'], function() { console.log("Done!"); });`

The load-complete function never fires. Crafty fails silently and doesn't ever complete loading. Please at least raise an error highlighting the problem, so that I can more quickly track down the issue (ideally, mention the duplicate-loaded asset).

Another option is just loading and moving on, ignoring the fact that an asset was loaded twice (provided it doesn't actually download twice).

mucaho:
So you don't get an error inside the [error callback](http://craftyjs.com/api/Crafty-loader.html)?

``` javascript
Crafty.load(['image.png', 'image.png'],
    function() { //when loaded
        console.log("Done!");
    },
    function(e) { //progress
    },
    function(e) { //uh oh, error loading
        console.log("Error!", e);
    }
);
```

ashes999:
@mucaho correct, the error call-back never fires.

heniotierra:
Fixed this for the new `Crafty.load` code. If the image(or audio) has already been loaded, it won't be loaded again, and oncomplete will still be called at the end, but no error is thrown in this case.

EDIT: A new PR was sent, this new one enabling onerror function to be called in this situation.

mucaho:
should be fixed via a5fd9872eceb97d5a86e146d6bf2ebc5191f50c0 now

PR #813, opened 2014-08-11
Fixed problem when trying to load already loaded asset

heniotierra:
Hi!

This patch fixes problem with refactored loader function that occurs when trying to (re)download already loaded assets.

Couple months ago I sent a PR with a fix for this, which made sure the file wouldn't be redownloaded, and still would call the onComplete function - but it wouldn't call the onError function in this specific situation. 

This patch now does all that: avoids redownloading the given asset, calls the onError function, and properly calls onComplete when finished. 

I messed that first PR when trying to push (squashing) this updated version, so please forgive me for sending it again.

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -292,45 +292,50 @@ Crafty.extend({
             for(asset in data[type]) {
 
                 current = data[type][asset];
+                obj = null;
 
                 if (type === "audio" && audSupport) {
-                    if (typeof current === "object") {
-                        var files = [];
-                        for (var i in current) {
-                            fileUrl = getFilePath(type, current[i]);
-                            if (!isAsset(fileUrl) && isSupportedAudio(current[i]))
-                                files.push(fileUrl);
+                    if (!Crafty.audio.sounds[asset]) {
+                        if (typeof current === "object") {
+                            var files = [];
+                            for (var i in current) {
+                                fileUrl = getFilePath(type, current[i]);
+                                if (!isAsset(fileUrl) && isSupportedAudio(current[i]))
+                                    files.push(fileUrl);
+                            }
+                            if (files.length)
+                                obj = Crafty.audio.add(asset, files).obj;
                         }
-                        obj = Crafty.audio.add(asset, files).obj;
-                    }
-                    else if (typeof current === "string" && isSupportedAudio(current)) {
-                        fileUrl = getFilePath(type, current);
-                        if (!isAsset(fileUrl))
-                            obj = Crafty.audio.add(asset, fileUrl).obj;
-                    }
+                        else if (typeof current === "string" && isSupportedAudio(current)) {
+                            fileUrl = getFilePath(type, current);
+                            if (!isAsset(fileUrl))
+                                obj = Crafty.audio.add(asset, fileUrl).obj;
+                        }
 
-                    //addEventListener is supported on IE9 , Audio as well
-                    if (obj && obj.addEventListener)
-                        obj.addEventListener('canplaythrough', pro, false);
+                        //addEventListener is supported on IE9 , Audio as well
+                        if (obj && obj.addEventListener)
+                            obj.addEventListener('canplaythrough', pro, false);
+                    } else {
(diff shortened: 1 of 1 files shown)


heniotierra:
Fixes issue https://github.com/craftyjs/Crafty/issues/764 which I know is related to the loader previous to refactoring, but the problem is also present in the refactored function.

heniotierra:
I found there was still a small problem, **which is now definitely fixed**, which was the err function informing the wrong path to a given sound (worked fine for images) in case it had already been loaded. It would be nice if somebody could test this patch.

kevinsimper:
Looks good, is there a easy way to test it? :)

heniotierra:
Yes, you could just call `.load` and then, after it was complete, call it again passing the same object. I'll try to put up a fiddle.

src/loader.js
```diff
@@ -292,45 +292,50 @@ Crafty.extend({
...
 
-                    //addEventListener is supported on IE9 , Audio as well
-                    if (obj && obj.addEventListener)
-                        obj.addEventListener('canplaythrough', pro, false);
+                        //addEventListener is supported on IE9 , Audio as well
+                        if (obj && obj.addEventListener)
+                            obj.addEventListener('canplaythrough', pro, false);
+                    } else {
+                        var rm = window.location.origin + window.location.pathname;
+                        fileUrl = Crafty.audio.sounds[asset].obj.src.replace(rm, "");
```
kevinsimper, line 320: We do we need this?

heniotierra:
Here the fiddles. The expected behaviour is calling `.load` three times, and alerting "finished" each of these times, with error messages after the first and second calls for each asset.

Here with the failing code
http://jsfiddle.net/heniot/u3d8yxLn/14/

Here with the fix
http://jsfiddle.net/heniot/u3d8yxLn/15/

src/loader.js
heniotierra, line 320: So that we can grab the file name or file URL in case that sound has already been loaded. Without that, `fileUrl` will have the value of the previously processed sound.

kevinsimper, line 320: But it gets never used anywhere, because the audio object do exist and therefore the error handler at last do not use it.

heniotierra, line 320: It's used in line 337:
`err.call({ src: fileUrl });`

kevinsimper, line 320: Yes, but your code gets run if this is true:
`Crafty.audio.sounds[asset]``

and if that is true, then `obj` is try because that is the audio object and then `err` is not called.

kevinsimper:
I like your jsfiddle, but `console.log` feels better :)

src/loader.js
heniotierra, line 320: Do you mean line 298 `if (!Crafty.audio.sounds[asset]) {` ? It's the opposite of what you said: if the asset does not exist in `Crafty.audio.sound`, then the sound is created and `obj` is set, otherwise `fileUrl` is grabed from the already loaded file.

heniotierra:
Sure thing! Here it is:

Failing code
http://jsfiddle.net/heniot/u3d8yxLn/17/

Fixed code
http://jsfiddle.net/heniot/u3d8yxLn/18/

src/loader.js
kevinsimper, line 320: That is because your code looks like this if you flip the if:

```
if (Crafty.audio.sounds[asset]) {
    var rm = window.location.origin + window.location.pathname;
    fileUrl = Crafty.audio.sounds[asset].obj.src.replace(rm, "");
}
```

heniotierra, line 320: Yes. I fail to see where it is wrong. Explaining: if there is a `Crafty.audio.sounds[asset]`, then we will have a `fileUrl`, but not an `obj`, and `err` will be called.

heniotierra, line 320: @kevinsimper Do you still think that's not right? 
The fact that you get the error messages in the fiddle with the fix means `err` is being called in line 337.

mucaho:
Thanks, added some tests and merged via a5fd9872eceb97d5a86e146d6bf2ebc5191f50c0.

PR #815, opened 2014-08-13
Multiway now keeps tracking of keydown's

heniotierra:
Issue: Currently, keydown will only hold keys activated through actual keyboard input. 

This small change allows the user to keep track of keydowns when **KeyDown** event is triggerred directly using `Crafty.trigger("KeyDown", key)`. It thus allows one to also use Keyboard's component `isDown` function. I did this modification in the making of a simple GUI with directional arrow keys for use with mobiles.

### DIFF
--- a/src/controls.js
+++ b/src/controls.js
@@ -318,11 +318,9 @@ Crafty.extend({
 
         if (e.type === "keydown") {
             if (Crafty.keydown[e.key] !== true) {
-                Crafty.keydown[e.key] = true;
                 Crafty.trigger("KeyDown", e);
             }
         } else if (e.type === "keyup") {
-            delete Crafty.keydown[e.key];
             Crafty.trigger("KeyUp", e);
         }
 
@@ -707,6 +705,7 @@ Crafty.c("Multiway", {
 
     _keydown: function (e) {
         if (this._keys[e.key]) {
+            Crafty.keydown[e.key] = true;
             this._movement.x = Math.round((this._movement.x + this._keys[e.key].x) * 1000) / 1000;
             this._movement.y = Math.round((this._movement.y + this._keys[e.key].y) * 1000) / 1000;
             this.trigger('NewDirection', this._movement);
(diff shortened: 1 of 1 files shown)


heniotierra:
The Twoway line that I modified obviously does not really change anything, except that instead of inheriting from Fourway, it now inherits directly from Multiway, since that's what's used instead of `fourway`.

starwed:
Hmm, on the surface this PR doesn't make sense to me.

The `Crafty.keydown` dictionary tracks the global state of the physical keyboard.  "Multiway" isn't the proper place to handle that logic -- you might have other components that consume "KeyUp" and "KeyDown" events.

There's probably a more natural way to fix the issue you're having.  :)

heniotierra:
> you might have other components that consume "KeyUp" and "KeyDown" events.

Yes, you're right... It's just that I had looked for any bindings to KeyDown and KeyUp events other than Multiway's, and there wasn't any other, so I thought Crafty.keydown was only used on multiway context, although it is so obvious that it could be used in many other contexts. And I knew Multiway wasn't the correct place for that, semantically speaking, but it solved my problem(lazy ass me), so maybe it'd help somebody else... Anyway, sorry for making you put your time on this silly thing.

I'll try to think of a better way to deal with this situation of mine.

Thanks for answering, sir! (Y) 

[I should probably close this issue.]

heniotierra:
I already found how to fix this problem in a less "invasive" way, without having to change Crafty's source. Thanks again!

MakerJake:
Can you get me off the email list?!?!?!

starwed:
@MakerJake This issue hasn't been updated in months, so not sure why you're asking here?

Anyway, this isn't an email list -- if you're getting notifications, it's because you've set up your github account to get them.
-  If you look at the right column of a github issue, there might be a button marked `unfollow`.  That'll stop you getting notifications about a particular issue.
- If you're getting notifications for _everything_, you've chosen to watch the entire repo, you need to change those settings instead.  (Perhaps simply the `unwatch` button at the top.)

PR #816, opened 2014-08-16
Add ability to see if two polygons overlap.

bayleedev:
poly_overlap

### DIFF
--- a/src/2D.js
+++ b/src/2D.js
@@ -1186,6 +1186,43 @@ Crafty.polygon.prototype = {
             current[0] = x;
             current[1] = y;
         }
+    },
+
+    /**@
+     * #.overlap
+     * @comp Crafty.polygon
+     * @sign public bool .overlap(Polygon item)
+     * @param item - Polygon you are checking overlap on.
+     *
+     * Determines if two polygons overlap. Do polygon a have points
+     * in polygon b or vise-versa?
+     *
+     * @example
+     * ~~~
+     * var poly1 = new Crafty.polygon([50,0],[100,100],[0,100]);
+     * var poly2 = new Crafty.polygon([0,0],[50,50],[0,50]);
+     * poly1.overlap(poly2); // true
+     *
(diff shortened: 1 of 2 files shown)


bayleedev:
actually this idea wouldn't work if poly2 was `[0,0],[100,50],[0,50]`.

bayleedev:
I suppose I could make an array of points between the given points and check all of those instead. Is there a better way to do this?

bayleedev:
It's not very performant to do this for collision detection. The reason I started this is I have a entity that has a section of itself that is "Solid" the rest is not. I think a better implementation would be to have new collision component that simply used a collidable x/y/h/w instead of the attributes x/y/h/w

starwed:
> I think a better implementation would be to have new collision component that simply used a collidable x/y/h/w instead of the attributes x/y/h/w

Collision between two polygons is already handled by the "Collision" component.  It uses a fairly standard algorithm for checking collisions; it'll work between any two convex polygons.  If you have a partially solid entity, you just explicitly assign a hitbox to it.

(There are probably numerous improvements that could be made to the underlying implementation, but it definitely exists and works as is.)

bayleedev:
ahh, I saw you could define character1 as a polygon, but I didn't realize you could define the other non-moveable entities as such too...
[Example](http://jsfiddle.net/L6agsp1L/)

``` coffeescript
Crafty.scene "Start", ->
  Crafty.e("2D, DOM, Color, Collision, Solid")
    .color("green")
    .attr(w: 100, h: 100, x: 100, y: 200)
    .collision [0, 0], [100, 0], [100,10], [0,10]
  Crafty.e("2D, DOM, Color, Fourway, Collision")
    .fourway(4)
    .color("blue")
    .attr(w: 100, h: 100, x: 0, y: 0)
    .collision([0, 0], [0, 10], [10, 100], [0, 100])
    .onHit "Solid", ->
        @color "red"

Crafty.init 500, 500
Crafty.background "Silver"
Crafty.scene "Start"
```

PR #821, opened 2014-09-25
Prevent buttons other than left cancelling a `Draggable`'s drag on a `mo...

### LINKED ISSUE #820: Mouseup event with right button ends drag
The `Draggable` component only starts a drag on _left_ mouse down, but will end it on either _right or left_ up.

I'd like to be able to right click while dragging something around. It seems natural to be able to do so. If you agree, I can submit a pull request with a fix.

[Relevant code here](https://github.com/craftyjs/Crafty/blob/e7ab075da3460b4d9ebda6004bc44d42d1e9f582/src/controls.js#L521-L535)

rhys-vdw:
...useup` event. Fixes craftyjs/Crafty#820

### DIFF
--- a/src/controls.js
+++ b/src/controls.js
@@ -526,7 +526,7 @@ Crafty.c("Draggable", {
 
     _onup: function (e) {
         // While a drag is occurring, this method is bound to mouseup DOM event
-        if (this._dragging === true) {
+        if (e.mouseButton === Crafty.mouseButtons.LEFT && this._dragging === true) {
             Crafty.removeEvent(this, Crafty.stage.elem, "mousemove", this._ondrag);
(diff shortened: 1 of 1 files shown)


starwed:
Looks good to me, passes existing tests, so merging.  Thanks!

(Though, it looks like we don't really have any tests for mouse behavior -- apparently [phantomjs](https://github.com/ariya/phantomjs/issues/10586) does support simulating drag+drop, I wonder how hard it would be to add that sort of thing to our test suite.)

rhys-vdw:
Thank you.

I'm open to contributing to the project. It's really nice. Perhaps create a
new issue for missing Draggable unit tests and I'll try to see if I can
integrate them in the coming weeks.

PR #822, opened 2014-09-30
Fixed problem saving JSON string with numbers in quotes

acrisioscunha:


kevinsimper:
Thanks for the PR.

The way to use Crafty.storage, is just to pass the javascript object, and then that stringify's it.

When would you stringify the object before passing it to `Crafty.storage`? :)

acrisioscunha:
Sorry my english is bad and I do not know if I understand your question.

When receiving an information as a string and do a check of information processing before saving.

The use of Crafty.storage function has not changed and return the same value as string to Crafty.storage.

kevinsimper:
I think this would confuse users as sometimes you want to save a number as a string and not as a number, and saving a object now works, so I do not understand why you would stringify beforehand when `Crafty.storage` does stringify if it is a object, so no need for stringify before.

Also a test that could be added to show the functionality:

```
test('saving string json object', function(){
    var data = {
        a: 4,
        b: 8.4
    };
    Crafty.storage('data', data);
    var testData = Crafty.storage('data');
    equal(testData.a, 4, 'the values should be equal');
    equal(testData.b, 8.4, 'the values should be equal');
    Crafty.storage.remove('data');
});
```

Issue #757, opened 2014-03-18
Crafty.audio sound looping failing randomly


pawurb:
I am just finishing a work on a game using Crafty: http://siorb.dobreziele.pl/

I came across an issue with looping sounds. Initial launch of 'gameplay' scene works all fine. After coming back from 'gameOver' to 'gameplay' scene, background soundtrack which is infinitely looped 6sec. (mp3 and ogg file) plays only once and then stops. I execute it like this:
        `Crafty.audio.play('gameplay', -1, 1)`
When I execute that sound two times, sometimes one of the channels get looped but it is random. 

I noticed that it breaks depending on how many different sounds got played during the previous gameplay. I stop all the not needed sounds so I don't that this is a problem in the max number of channels. Is there any way i I could 'reset' the channels so that previously played sounds do not affect the present ones ?

I googled for the answer but the only clue I found was that ' implementation of HTML5 Audio is buggy at best.' in Crafty docs. :\ Any clues will be appreciated.

PR #823, opened 2014-10-02
added repeat 'atTime' feature to audio.play

heniotierra:
This PR makes it possible to define a time (audio reproduction time range) at which the sound should be repeated. It allows you to force a seemless/gapless sound loop, even with your browser's limitations (the delay that occurs in most if not all systems between the end of the execution of the audio 'iteration' and the proper restart). Works even if your file has a gap at the end (if an .mp3, for example), but obviously not at the beginning.

Usage:
`// Repeat (forever) when sound reach second 23.2. Should actually repeat somewhere between second 23.2 and 23.4.`
`Crafty.audio.play("backgroundMusic", -1, 1, 23.2);`

### DIFF
--- a/src/sound.js
+++ b/src/sound.js
@@ -105,7 +105,6 @@ Crafty.extend({
             Crafty.asset(path, audio);
             this.sounds[id] = {
                 obj: audio,
-                played: 0,
                 volume: Crafty.audio.volume
             };
             return this.sounds[id];
@@ -187,32 +186,45 @@ Crafty.extend({
             }
             return a;
         },
+
         /**@
          * #Crafty.audio.play
          * @comp Crafty.audio
-         * @sign public this Crafty.audio.play(String id)
-         * @sign public this Crafty.audio.play(String id, Number repeatCount)
-         * @sign public this Crafty.audio.play(String id, Number repeatCount, Number volume)
+         * @sign public this Crafty.audio.play(String id[, Number repeatCount [, Number volume [, Number repeatAt]]])
          * @param id - A string to refer to sounds
-         * @param repeatCount - Repeat count for the file, where -1 stands for repeat forever.
-         * @param volume - volume can be a number between 0.0 and 1.0
+         * @param repeatCount - Repeat count for the file, where -1 stands for repeat forever. Defaults to 1.
+         * @param volume - Volume can be a number between 0.0 and 1.0.
+         * @param repeatAt - Reproduction time in seconds at which the sound should be repeated.
          * @returns The audio element used to play the sound.  Null if the call failed due to a lack of open channels.
          *
-         * Will play a sound previously added by using the ID that was used in `Crafty.audio.add`.
-         * Has a default maximum of 5 channels so that the same sound can play simultaneously unless all of the channels are playing.
-
+         * Will play a sound previously added by using the ID that was used in 
+         * `Crafty.audio.add`. Has a default maximum of 5 channels so that the 
+         * same sound can play simultaneously unless all of the channels are 
+         * playing.
+         * 
+         * When using repeat, and if it's important for you to have a gapless 
+         * loop, try passing the 'repeatAt' argument, which makes the sound 
+         * to be restarted at given reproduction time (in seconds), with 
+         * aproximate accuracy - there might be a discrepancy of some 
+         * milisenconds (perhaps 100-200 ms ahead), but the sound will still 
+         * play seemlessly.
+         *
          * *Note that the implementation of HTML5 Audio is buggy at best.*
          *
          * @example
          * ~~~
          * Crafty.audio.play("walk");
          *
-         * //play and repeat forever
-         * Crafty.audio.play("backgroundMusic", -1);
-         * Crafty.audio.play("explosion",1,0.5); //play sound once with volume of 50%
+         * // Play sound once with volume of 50%
+         * Crafty.audio.play("explosion", 1, 0.5);
+         *
+         * // Repeat (forever) when sound reach second 23.2. 
+         * // Should actually repeat somewhere between second 23.2 and 23.4.
+         * Crafty.audio.play("backgroundMusic", -1, 1, 23.2); 
+         *  
          * ~~~
          */
-        play: function (id, repeat, volume) {
+        play: function (id, repeat, volume, repeatAt) {
             if (repeat === 0 || !Crafty.support.audio || !this.sounds[id])
                 return;
             var s = this.sounds[id];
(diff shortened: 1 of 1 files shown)


heniotierra:
I don't understand why this is failing Travis test... This is the message:

> PhantomJS timed out, possibly due to a missing QUnit start() call.

src/sound.js
```diff
@@ -187,32 +186,39 @@ Crafty.extend({
...
             return a;
         },
+
         /**@
          * #Crafty.audio.play
          * @comp Crafty.audio
-         * @sign public this Crafty.audio.play(String id)
-         * @sign public this Crafty.audio.play(String id, Number repeatCount)
-         * @sign public this Crafty.audio.play(String id, Number repeatCount, Number volume)
+         * @sign public this Crafty.audio.play(String id[, Number repeatCount [, Number volume [, Number repeatAt]]])
```
kevinsimper, line 193: I can see that you try to tell that the a optional, but shouldn't it look more like this:
http://api.jquery.com/jQuery.get/

Or do you have an example where they do it like this?

`@sign public this Crafty.audio.play(String id[, Number repeatCount] [, Number volume] [, Number repeatAt])`

src/sound.js
```diff
@@ -187,32 +186,39 @@ Crafty.extend({
...
+         * @param repeatCount - Repeat count for the file, where -1 stands for repeat forever. Defaults to 1.
+         * @param volume - Volume can be a number between 0.0 and 1.0.
+         * @param repeatAt - Reproduction time in seconds at which the sound should be repeated.
          * @returns The audio element used to play the sound.  Null if the call failed due to a lack of open channels.
          *
          * Will play a sound previously added by using the ID that was used in `Crafty.audio.add`.
          * Has a default maximum of 5 channels so that the same sound can play simultaneously unless all of the channels are playing.
-
+         * 
+         * When using repeat, and if it's important for you to have a gapless loop, try passing the 'repeatAt' argument, which makes
```
kevinsimper: We should try to keep the comments to a 80 column :)

src/sound.js
```diff
@@ -187,32 +186,39 @@ Crafty.extend({
...
-         * @sign public this Crafty.audio.play(String id)
-         * @sign public this Crafty.audio.play(String id, Number repeatCount)
-         * @sign public this Crafty.audio.play(String id, Number repeatCount, Number volume)
+         * @sign public this Crafty.audio.play(String id[, Number repeatCount [, Number volume [, Number repeatAt]]])
          * @param id - A string to refer to sounds
-         * @param repeatCount - Repeat count for the file, where -1 stands for repeat forever.
-         * @param volume - volume can be a number between 0.0 and 1.0
+         * @param repeatCount - Repeat count for the file, where -1 stands for repeat forever. Defaults to 1.
+         * @param volume - Volume can be a number between 0.0 and 1.0.
+         * @param repeatAt - Reproduction time in seconds at which the sound should be repeated.
```
kevinsimper, line 197: What do you mean by "Reproduction"? :)

src/sound.js
```diff
@@ -220,43 +226,52 @@ Crafty.extend({
...
 
+            var a = c.obj;
             a.volume = s.volume;
             a.src = s.obj.src;
 
             if (this.muted)
                 a.volume = 0;
             a.play();
-            s.played++;
+            c.played++;
```
kevinsimper, line 246: Shouldn't this be the sounds object `s`? `var s = this.sounds[id];`

src/sound.js
heniotierra, line 193: Fisrt of all, thanks for answering! :)

I did it like that based on some other functions in Crafty:
.delay in https://github.com/craftyjs/Crafty/blob/develop/src/time.js
.load in https://github.com/craftyjs/Crafty/blob/develop/src/loader.js

Which one is the "most correct" way?

src/sound.js
heniotierra: Oh, ok.

src/sound.js
heniotierra, line 197: Reproduction: the act or process of reproducing
Reproduce: play back, replay - reproduce (a recording) on a recorder; "The lawyers played back the conversation to show that their client was innocent"

Taken from http://www.thefreedictionary.com/reproduction and thesaurus in http://www.thefreedictionary.com/reproducing

Do you think that's confusing?

src/sound.js
heniotierra, line 246: I don't think so. Unless I'm mistaken (which could be the case), no, because keeping it `s` would cause problems, since `s.played` (and `c.repeat`) was being increased continuously, at each play (or repeat), and that makes repeating to break, as well as 'playing once'  after a while (repeating instances, or combinations of using `.play`, `.stop`, `.play` again, etc, with different parameters or not).
Also, `s` is an object shared between all the instances (channels) of the same sound, so if you're playing a sound more than once at the same time, reseting `s.repeat` would change that value for all those channes, which caused more problems, and we wouldn't want that.
Moving those into `c` and then reseting the values when they finish playing is painless and super effective. :)

heniotierra:
@kevinsimper Hey man, I just changed the comments section to stick to 80 column max, as suggested by you.

heniotierra:
There seems to be a very weird bug with this PR... Using  `Crafty.audio.stop()` breaks game logic, skiping the rest of the callback function bound to the event. But works when passing sound ids ->  `Crafty.audio.stop("soundID")` . Also, `Crafty.audio.stop()` works in the console and outside events(?). 

The problem appears in situations like this below, and it won't show any error message, but makes the game to stop working properly (the event below is like a 'scene ending' event). I'm still trying to investigate this, but it doesn't make much sense to me.

```
Crafty.one('SomeEvent', function(){
    console.log('before audio stop'); // executed
    // in this context Crafty.audio.stop only works if sound id argument is passed
    Crafty.audio.stop(); // this breaks stuff
    console.log('after audio stop'); // not executed
    (...)
});

// then, some time later...
Crafty.trigger('SomeEvent'); 
```

EDIT: fixed misinformation in this comment's two paragraphs.
EDIT2: I accidentally closed the PR.

heniotierra:
@kevinsimper  I forgot to mention that the changes I made to 'audio.stop' and 'sound' and 'channel' objects fix issue https://github.com/craftyjs/Crafty/issues/757 .. However, up to this point I was unable to surpass the bug mentioned in the comment above, except by avoiding calling 'Crafty.audio.stop' without passing sound ID. Just doing that makes all run smoothly... But because of the bug, feel free to close this PR. I could also send the fix for the issue linked in a seperate PR...

heniotierra:
I'm closing this, since I couldn't find a solution for the bug mentioned.

PR #825, opened 2014-10-11
Update bower.json

navelpluisje:
Added ignore; This way most parts will not be fetched by bower
Changed main; Added dist in front of the name so the source can be added to a project automated.

This way the file is more conform the defaults.


kevinsimper:
Cool

Issue #828, opened 2014-10-30
SAT implementation bug


VHonzik:
I believe there is a bug in SAT implementation in Crafty. The issue are the default values for minimum and maximum of vertices projection onto axes inside `_SAT: function (poly1, poly2)` (collision.js):

``` javascript
//default min max
min1 = min2 = -1;
max1 = max2 = -1;
```

The dot products few lines down can be negative and what is a problem even -1. See the attached diagram at the bottom and following lines:

``` javascript
dot = points1[j][0] * normal.x + points1[j][1] * normal.y;
dot = points2[j][0] * normal.x + points2[j][1] * normal.y;
```

Suppose the dot products for a rectangle polygon are -1,-1,-50,-50 (which in fact are the results for Poly1 in the attached image) in this order and examine the following min/max if conditions:

``` javascript
if (dot > max1 || max1 === -1) max1 = dot;
if (dot < min1 || min1 === -1) min1 = dot; 
```

Both min1 and max1 will end up being -50 when the result should be `min1=-50; max1=-1`. Thanks to SAT exiting as early as possible, Crafty will return the polygons don't collide but they clearly do.

I humbly suggest setting the defaults directly to the first dot product and later start the projections computation from the second vertex so you can save one assignment yaaay!

![crafty 2](https://cloud.githubusercontent.com/assets/5425854/4842170/953edf66-6022-11e4-959b-8848e10b2cb0.png)

starwed:
Eek, that is some gnarly code.  

An alternate fix would be

``` javascript
//default min max
min1 = min2 = Infinity;
max1 = max2 = -Infinity;
```

The collision routine has long needed some auditing for performance.  (For instance, when the polygons are two rectangles -- there's definitely a faster way!)

starwed:
Fixed by #829

PR #829, opened 2014-10-30
Fix bug in _SAT collision test

starwed:
The SAT function improperly calculated min/max values in some cases.  This fixes that issue by comparing to +/- Infinity instead of special casing the initial values.  (It also gets rid of a now needless comparison, which might help performance.)

This also includes a qunit test that catches the issue.

Fixes issue #828.

### DIFF
--- a/src/collision.js
+++ b/src/collision.js
@@ -596,21 +596,21 @@ Crafty.c("Collision", {
             normal.y /= length;
 
             //default min max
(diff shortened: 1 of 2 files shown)


kevinsimper:
Looks good to me, looks more simple too!

starwed:
Ok, merging.

PR #830, opened 2014-11-07
Do not reset components when doing a hard reset - Fixes #827

### LINKED ISSUE #827: Crafty.stop(true) removes all components with no way to reload them
According to the docs one should be able to call "Crafty.stop(true); Crafty.init();" -- (for example if one wants to launch a new game hosted in SPA).

However, Crafty.stop(true) calls initState() [core.js:940] which resets components to the empty object [core.js::48].  And since the built in components only register themselves when first loaded (e.g. Crafty.c("2D", ... ) [2D.js::77]), they are gone with no hope of returning without a page refresh.

I'm not sure what's supposed to happen here. In a simple fix Crafty.stop(true) would not flush components. Even better maybe, components would register themselves into a 'staging' area so game developers can explicitly cherry-pick only the components they want to register into the engine?

kevinsimper:
Fixes #827

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -39,25 +39,25 @@ var version = require('./version');
 
 var Crafty = function (selector) {
     return new Crafty.fn.init(selector);
-},
+};
     // Internal variables
-    GUID, frame, components, entities, handlers, onloads,
-    slice, rlist, rspace, milliSecPerFrame;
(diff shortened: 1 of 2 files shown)


kevinsimper:
Going to merge this in.

PR #832, opened 2014-11-12
Implement a new strategy for avoiding event queue mutation

### LINKED ISSUE #824: TweenEnd is fired multiple times
I noticed this long ago, but always find a way to circumvent the problem. What the title says, in some situations TweenEnd is fired multiple (sometimes three or four) times, but for me it's obvious it should fire only once. Because of this issue, ugly workarounds are needed.
In most cases using `this.one("TweenEnd",function(){ [...] })` avoids the problem, but in some other situations this won't help so much, forcing me to use `delay` to execute that callback function.
Oh, by the way, I remember the delay bug I helped solve some time ago was caused by the enterframe event being fired a few times after pausing... maybe there is a relation there? Could it be that the function that triggers the tween steps (on enterframe?) is still bound even after tween "ended"? Or something like that... (I don't fully understand Tween inner workings yet)

starwed:
The event callback queue can mutate in ways that mess up the iteration.  There are two cases we need to handle:
- While iterating through an event's callbacks, a new one is added.  This should not be triggered.
- When a callback is removed, we shouldn't actually mutate the callbacks array if this iteration was triggered in the middle of an iteration over the same array.  (We currently kind of half-handle this case, but there's a better way.)

The first is solved by registering the callback array's length at the time iteration starts.  The second is solved by adding a depth counter to each callback array.  It's iterated at the beginning of iteration, and decremented at the end.  Items are removed from the callbacks array only when the counter indicates a depth of 1.

This patch adds a test for the addition of a callback mid-event.  It also adds one for binding/triggering groups of entities at once, since I noticed a typo I introduced while modifying the code wasn't being caught by the existing tests.

This fixes #824.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -637,24 +637,27 @@ Crafty.fn = Crafty.prototype = {
     bind: function (event, callback) {
 
         // (To learn how the handlers object works, see inline comment at Crafty.bind)
+        var h = handlers[event] || (handlers[event] = {}), callbacks;
 
         //optimization for 1 entity
         if (this.length === 1) {
-            if (!handlers[event]) handlers[event] = {};
-            var h = handlers[event];
-
-            if (!h[this[0]]) h[this[0]] = []; //init handler array for entity
-            h[this[0]].push(callback); //add current callback
+            callbacks = h[this[0]];
+            if (!callbacks) {
+                callbacks = h[this[0]] = []; //init handler array for entity
+                callbacks.depth = 0; // metadata indicating call depth
+            }
+            callbacks.push(callback); //add current callback
             return this;
         }
 
         this.each(function () {
             //init event collection
-            if (!handlers[event]) handlers[event] = {};
-            var h = handlers[event];
-
-            if (!h[this[0]]) h[this[0]] = []; //init handler array for entity
-            h[this[0]].push(callback); //add current callback
+            callbacks = h[this[0]];
+            if (!callbacks) {
+                callbacks = h[this[0]] = []; //init handler array for entity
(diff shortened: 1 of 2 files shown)


kevinsimper:
Could we take something from Backbone?

Here is an equal example made in backbone and they do not fire handlers that is bound in the current `trigger`.

http://jsbin.com/bujagavope/1/edit?html,js,output

Here is the code that does it.
http://backbonejs.org/docs/backbone.html#section-28

starwed:
By itself, that wouldn't be safe if one of the events was removed from the array in response to the trigger.  So they must do something else different.

Looking through the code, when removing an event it looks like they [create a new array of events](http://backbonejs.org/docs/backbone.html#section-21), and only push the events to retain onto it.  So the original array is left untouched, and any existing iteration can continue without a problem.  That probably wouldn't work for us because of how much we need to worry about GC?  Generally creating new arrays all over the place is bad for that.

One very simple thing we could steal is the use of `while` loops instead of `for`.

Anyway, this patch prevents the same types of issues.  (One way it behaves differently is that if an event is removed in Crafty, it won't trigger on the current event, assuming it hasn't already.  While that makes sense, it does make things more dependent on the _order_ of iteration.)

kevinsimper:
Looks good to me :+1: 

Is this what we need for Crafty 6.3? we should ship the update :)

starwed:
Yes, you can pull this in to 0.6.3 if you want.  After looking at the event code there are a couple more intrusive changes I want to make, but this one is quite simple and should be pretty low risk.  :)

Issue #746, opened 2014-03-06
Improve performance of methods that return new objects


mucaho:
Improve the performance of `2D.pos()`, `2D.mbr()` and similar methods to allow passing an optional argument in which the result will be stored.
- [x] `2D.pos(), 2D.mbr()`
- [x] `Crafty.math.Matrix2D`
- [x] `Crafty.map.key`

addisonbgross:
Are you looking to add a constructor to say Crafty.math.Matrix2D that will store the newely created matrix in the argument passed? eg: var matrix; Crafty.math.Matrix2D( matrix )

starwed:
@addisonbgross  Yes, I think that's the idea.  It would allow for easy recycling of objects in contexts where that makes sense, such as iterations.

We actually already do this with `map.key`.

addisonbgross:
I'll get my people on it

starwed:
Closing this; since it now works for `pos` and `mbr`.  `key` already supported it, and `Matrix2D.setValues` works for that object.

PR #833, opened 2014-11-17
Adding Matrix2D bonus constructor and tests

addisonbgross:


starwed:
Cool, thanks!  (Especially for adding the tests.)  But it looks like this won't merge cleanly -- you'll probably want to rebase against a more recent version of the develop branch?

A couple of questions/comments:
-  Do we  need qunit as a dependency?  We already include it via `grunt-contrib-qunit`
- Looking at github's diff, I think the whitespace is a bit inconsistent with the surrounding code
- What does the line `Matrix2D.prototype.matrix = [];` do?

addisonbgross:
I got rid of the line "Matrix2D.prototype.matrix = []", and the excess qunit inclusion. Hopefully this resolves the conficts

starwed:
Thanks, looks good!  Just one last thing: it's nice to squash the commits together before we merge.  You can do that easily with an interactive rebase; here, the command would be `git rebase HEAD~3 -i`.  It'll present you with a list of the last three commits; you change the type of the most recent two from `pick` to `fixup`. 

Then you force push (using the flag `-f`) to your own repo, and the changes be reflected in this PR.

The point of this is to keep our git history nice and clean, without showing the intermediate steps that each patch went through.  :)  The only time you shouldn't do this is if someone was working directly off of your branch.  (In other words, if it should be treated as a public branch rather than your own private work.)

I can do this on my end if you don't have time/etc, just let me know.

addisonbgross:
Thanks for the info! Hopefully it is ready now
Edit: Also, two of my peers are completing the other two components of issue #746 right now

addisonbgross:
The final part of this issue should be ready for submission by the end of this week. I will make sure that our commits are organized and final :)

addisonbgross:
This issue is complete! I am unsure how to squash the commits correctly, but it is all in place. If this could be accepted before Dec 3rd, I would be eternally greatful! :)

src/2D.js
```diff
@@ -591,19 +591,31 @@ Crafty.c("2D", {
...
-            _w: (this._w),
-            _h: (this._h)
-        };
+    pos: function (pos) {
+	   if (arguments.length == 1) {
+		  pos._x = (this._x);
+		  pos._y = (this._y);
+		  pos._w = (this._w);
+		  pos._h = (this._h);
+		  return;
```
starwed: For convenience, the method signature should stay the same -- return pos here.  (And the same goes for `mbr`.)

src/2D.js
```diff
@@ -613,14 +625,34 @@ Crafty.c("2D", {
...
-        return {
-            _x: (this._mbr._x),
-            _y: (this._mbr._y),
-            _w: (this._mbr._w),
-            _h: (this._mbr._h)
-        };
+    mbr: function (mbr) {
+		if (!this._mbr) {
+			if (arguments.length == 1) {
+		        mbr._x = (this.pos()._x);
```
starwed, line 631: Just use `return this.pos(mbr)`, rather than calling `pos` four times.

starwed:
I added a couple of notes.  

You could also consider checking whether the argument is undefined, rather than checking the length of the arguments array.

Since in this case we expect it to be an object, you could simply replace `if (arguments.length == 1)` with `if (mbr)` (and ditto for the other, similar functions).  That would also make it slightly easier for `pos` to call in to `mbr`.

It would be slightly more general to use `if (typeof mbr !== "undefined")`, but since the code will fail if someone passes in anything other than an object, it doesn't really matter here!  :)

To squash or reorder the changes, just do an interactive rebase.  The easiest way is just `git rebase -i develop`.

If you don't feel you have time before whatever deadline you're working with, I can just make all these changes locally... just let me know.  :)

addisonbgross:
We are all in place and ready. I'm not sure if the rebase was effective; if not, would you mind doing that? I don't think I fully understand the process

starwed:
Thanks!  But I don't think the rebase was successful.  (Or perhaps the push wasn't?)

I can fix it locally and then merge, but you might want to try figuring out how to do a rebase -- it's a pretty important part of working with git, and not so scary once you get used to it..

addisonbgross:
I definitely want to learn rebasing. I was fearful of ruining my classmate's progress, but now that the semester is over there will be time to learn. Thank you for all of your help :)

addisonbgross:
Is it conceivable for this pull request to be merged tonight or tomorrow? My classmates and I will receive bonus marks if it is accepted ;)

starwed:
Ok, thanks!  I merged half of it. 
- On closer inspection, the changes to Matrix2D were unnecessary (sorry for not noticing earlier) -- the function modified was a constructor, so it just created a new object _and_ assigned the values to the object passed in.  There's already a function `.setValues()` for copying the values of one matrix onto another, and I think it would make more sense to use that in most cases.
- I thought of a way to simplify the logic of `mbr` and `pos` a little, so went ahead and added that as a separate commit.  (There's a [javascript trick](http://addyosmani.com/blog/exploring-javascripts-logical-or-operator/) using `||` as a conditional operator.)
- I added a module declaration to the attached test files, so that they didn't show up under the "Collision" module.

Thanks again for the patch!

PR #834, opened 2014-11-17
Autobind parameter for declaring components

starwed:
Adds a special `events` role when creating components.  This is a shorthand for binding functions to events, with the added benefit that they'll be automatically unbound when the component is removed or destroyed.  You use it like this:

``` javascript
Crafty.c("MyComponent", {
    events: {
        EnterFrame: function(){
            // this function is bound to EnterFrame
        },
        Moved: function(){
            // this function is bound to the "Moved" event
        },
        // We can give a reference to a function instead!
        Change: "_changeFunction"
    },
    _changeFunction: function(){
        // This function is bound to "Change" by its name, but can still be called normally as a method on the component
    }
});

```

I'd probably want to go through and switch some of the built-in components to use this (as well as `remove()`) at some point.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -221,7 +221,7 @@ Crafty.fn = Crafty.prototype = {
      */
     addComponent: function (id) {
         var comps = [],
-            c = 0;
+            comp, c = 0;
 
         //add multiple arguments
         if (arguments.length > 1) {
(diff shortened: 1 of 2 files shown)


src/core.js
```diff
@@ -241,10 +241,18 @@ Crafty.fn = Crafty.prototype = {
...
+            comp = components[comps[c]];
+            this.extend(comp);
             //if constructor, call it
-            if (components[comps[c]] && "init" in components[comps[c]]) {
-                components[comps[c]].init.call(this);
+            if (comp && "init" in comp) {
+                comp.init.call(this);
+            }
+            if (comp && "autobind" in comp){
+                var auto = comp.autobind;
```
kevinsimper: Is this better than just referencing it inside the for loop?

`for (var eventName in comp.autobind){`

kevinsimper:
I really like this!

But can the name be confused with the javascript `.bind` functionality?

In Backbone (not hidden that i like Backbone), they just call it `events`.
How Backbone does it: http://backbonejs.org/#View-delegateEvents

The javascript bind https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind

Facebook React Autobind http://facebook.github.io/react/blog/2013/07/02/react-v0-4-autobind-by-default.html

Somebody doing a autobind function where the bind this to all the functions in a model https://gist.github.com/dotmaster/1070599

starwed:
> In Backbone (not hidden that i like Backbone), they just call it events.

I'd actually thought of just using `events`.  And it looks like they do it _exactly_ the same way I had in mind, so we might as well use the name, too.

src/core.js
starwed: Well, we still have to test that `comp` is defined, buy I guess we don't need to test for autobind.

starwed:
I think this will improve the ergonomics of a pretty common pattern, so merging.

PR #835, opened 2014-11-17
Fix _SAT algorithm, optimize Crafty.polygon

starwed:
Our current implementation of the SAT (Separating Axis Theorem) algorithm is both a bit slow and somewhat incorrect.

To improve the speed, this patch changes the behavior of Crafty.polygon to store points as a single flat array, rather than an array of smaller arrays.  This is a breaking change -- code that interacts with this (e.g. by creating collision regions) will have to be modified.  I'm pretty sure it's worth it, though.  Doing some jsperf tests indicates that dealing with a single array is much faster.

This patch also prevents the creation of temporary objects within each step of the SAT algorithm, which also increases the speed a bit.

The algorithm also seemed somewhat incorrect; it calculated the minimal overlap for each edge of each polygon, but never actually did anything with the first result.  This doesn't matter when dealing with simple rectangles, which is probably why this went unnoticed til now.

I added some tests to make sure our _SAT behavior isn't broken in the future.

### DIFF
--- a/src/2D.js
+++ b/src/2D.js
@@ -1098,21 +1098,21 @@ Crafty.c("Gravity", {
  * #Crafty.polygon
  * @category 2D
  *
- * Polygon object used for hitboxes and click maps. Must pass an Array for each point as an
- * argument where index 0 is the x position and index 1 is the y position.
+ * Polygon object used for hitboxes and click maps. Takes a set of points as an
+ * argument, giving alternately the x and y coordinates of the polygon's vertices in order.
  *
- * For example one point of a polygon will look like this: `[0,5]` where the `x` is `0` and the `y` is `5`.
- *
- * Can pass an array of the points or simply put each point as an argument.
+ * The constructor accepts the coordinates as either a single array or as a set of individual arguments.
+ * If passed an array, the current implementation will use that array internally -- do not attempt to reuse it.
  *
  * When creating a polygon for an entity, each point should be offset or relative from the entities `x` and `y`
  * (don't include the absolute values as it will automatically calculate this).
  *
  *
  * @example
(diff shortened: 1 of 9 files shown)


starwed:
The test files were failing `jshint`, so I force pushed a new version of the patch.

By the way, one advantage of switching to flat arrays for `Crafty.polygon` is that we could use typed arrays in the future, which are at least [a little bit faster](http://jsperf.com/sat3-vs-sat4).

Thanks to @Vhonzik for getting me to look at the _SAT code in #828.

There's still room for improvement here, especially when one of the objects is just an axis-aligned box.  (Which would actually be pretty common!)

kevinsimper:
Impressive pull request, quite a lot of changes but with nice tests.

I do not know what I feel about the new way to define coordinates, it makes it easier to mess up i think, but we can try it (we can just do like your speciel formatting).

starwed:
> I do not know what I feel about the new way to define coordinates, it makes it easier to mess up i think, but we can try it (we can just do like your speciel formatting). 

I should have probably linked to the [jsperf tests](http://jsperf.com/sat1-v-sat2) I ran.  In Firefox and Safari, the version (SAT3) with flat arrays is significantly faster than the version without (SAT2).   In Chrome it doesn't matter so much.   (SAT1 is the current implementation, which is slower in all browsers.)

I get similar results when [comparing containsPoint](http://jsperf.com/flat-vs-nested-arrays); the change doubles the speed in Firefox, and still provides a bump up for Safari and Chrome.

VHonzik:
> Thanks to @Vhonzik for getting me to look at the _SAT code in #828.

I'm glad I could contribute. Loving the speed of deployment.

starwed:
I'm going to go ahead and merge this, since although it breaks compatibility it really does bring a nice performance bump.

PR #838, opened 2014-12-01
Add/remove Sprite components sometimes fails to change background-image

FallSe7en:
When switching between two Sprite components that use different background
images, the second background-image fails to get set properly.

This happens because style.backgroundColor is an invalid value when trying to switch to the second component (it is "initial").

### DIFF
--- a/src/sprite.js
+++ b/src/sprite.js
@@ -193,9 +193,12 @@ Crafty.c("Sprite", {
                 // If needed, we will scale up the entire sprite sheet, and then modify the position accordingly
                 var vscale = this._h / co.h,
                     hscale = this._w / co.w,
-                    style = this._element.style;
+                    style = this._element.style,
+                    bgColor = style.backgroundColor;
 
-                style.background = style.backgroundColor + " url('" + this.__image + "') no-repeat";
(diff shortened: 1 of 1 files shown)


kevinsimper:
Hi FallSe7en

Thanks for the pull request! :+1: 

I was thinking if you have an example? You can do it online here with this boilerplate
http://requirebin.com/?code=375aaba238f0eb3fafe9

FallSe7en:
I hope a JSFiddle is ok (I'm more familiar with it)!

http://jsfiddle.net/f8c4c/15/

The first switch works because it's off the same image, but then the second switch fails to switch the background-image.

starwed:
Aha, the problem simply doesn't manifest in Firefox; that's why I wasn't able to reproduce it.  Thanks for the linked example!

It does occur in Chrome and Safari.  (I can't test IE.)  I guess those browsers revert the background color to "initial" when an image is removed, whereas Firefox doesn't?

This patch looks fine to me, so merging.

(Although, I wonder why we even bother keeping the background color?  I'm not sure we provide a means of setting it in the first place!)

Issue #837, opened 2014-11-30
Flip + Rotate = Trouble


jcnauta:
When rotating and flipping a canvas entity, the entity is not visible when drawn.
The reason is that in the lines following line 2875 (crafty version 0.6.3), the context is saved twice (once for rotation and once for flipping), but after that it is only restored once (line 2907 or so).

This leaves the context in the state after rotation, if I understand the [documentation](https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Drawing_graphics_with_canvas) correctly.

A solution that worked for me is to replace this:

``` javascript
if (this._rotation !== 0 || (this._flipX || this._flipY)) {
  context.restore();
}
```

with this:

``` javascript
if (this._rotation !== 0)
  context.restore();
if (this._flipX || this._flipY)
  context.restore();
```

Another solution would be to always save before transforming and drawing, and restoring afterwards.
I hope someone can fix this for the next release! Please correct me if I'm wrong on anything.

starwed:
A funny bit of synchronicity: I noticed this bug a couple years ago, and kind of thought I'd fixed it back then.  I never ran into it in any code I wrote, until just an hour ago!

It's probably slightly better to fix it at the other end -- that is, to only save the context once if necessary.  (Saving and restoring a context is apparently somewhat expensive when using canvas.)  If you want to submit a PR, go for it, otherwise I'll probably whip something up tomorrow.

jcnauta:
Such a coincidence! I never submitted a pull request before (something for my bucket list I suppose). At the moment I don't have time to figure it out, but hopefully I will in the near future.

I'm glad this will be fixed!

starwed:
Should be fixed by #839

PR #839, opened 2014-12-01
Fix a bug in canvas entity implementation when flipped and rotated

starwed:
Currently, the canvas context was saved separately for flipped and rotated entities, but restored only once.  This totally messes with canvas rendering!

This patch fixes this by only saving the context once if called for.  (I also added a couple of comments to the code.)

Unfortunately, we can't have canvas-specific tests in our test suite as currently implemented.

This fixes Bug #837.

### DIFF
--- a/src/canvas.js
+++ b/src/canvas.js
@@ -103,18 +103,21 @@ Crafty.c("Canvas", {
         co.w = w || coord[2];
         co.h = h || coord[3];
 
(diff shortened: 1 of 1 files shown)


kevinsimper:
Looks good! :+1:

starwed:
Ok, merging.

PR #840, opened 2014-12-05
report error if scene does not exist

fdavis:
The error for borking a scene name was a bit misleading. Its not that hard to correlate, but a more direct message should help people new to crafty like me, with our typos. :)

```
Uncaught TypeError: Cannot read property 'initialize' of undefined
```

error reproduced here: http://jsfiddle.net/z3bbzehv/

### DIFF
--- a/src/scenes.js
+++ b/src/scenes.js
@@ -154,7 +154,12 @@ Crafty.extend({
             oldScene: oldScene,
             newScene: name
         });
-        this._scenes[name].initialize.call(this, data);
+           
+        if (this._scenes.hasOwnProperty(name)) {
+            this._scenes[name].initialize.call(this, data);
+        } else {
+            console.error('The scene "' + name + '" does not exist');
+        }
 
         return;
 


starwed:
Thanks, that looks like a nice improvement.

But is it ok to use [console.error](https://developer.mozilla.org/en-US/docs/Web/API/Console.error)?    MDN suggests not to, but I don't know how up-to-date that advice is.

(Probably we should have some methods on Crafty for raising errors, but that's a separate issue.)

fdavis:
That MDN page suggests all of the console functions are non-standard. Are there any alternatives? It looks like Blackscorp setup Opera logging in [Cron with this line](https://github.com/BlackScorp/cron/blob/gh-pages/assets/js/game.js#L2)

```
if(window.opera){ console = {log:window.opera.postError} }
```

But I'm pretty sure that won't cover all browsers.

starwed:
From doing a grep of the source, I don't think we currently use `console.anything()`.  I'll go ahead and merge this, but I'll probably migrate it to something like `Crafty.error()` before the next release.  (Because it's kind of silly we don't have a standard way of logging errors.)

Thanks again for bringing this up!

(There used to be a fun [heisenbug](https://en.wikipedia.org/wiki/Heisenbug) in IE where `console` would be defined only when you had the dev console open.  That meant that using console.log would cause your program to crash, but _only when you weren't trying to debug it_.)

Issue #842, opened 2014-12-11
Multiway disable control issue


VHonzik:
Good afternoon everyone,
I used Crafty in recent Ludum Dare (was very content with it) and discovered a small annoyance. I would not necessary label it as a bug.

When entity is using `Multiway` component and `.disableControl` is called, the `_movement` attribute is still updated in `_keydown` function when a button is pressed. While the entity is not moved because of the first line in `_enterframe`, user code can still rely on  `_movement`. Prime example of such usage is  [Darren Torpey excellent tutorial](http://buildnewgames.com/introduction-to-crafty/) linked from Crafty website, where `_movement` is used to stop movement on collision.

I believe it would make sense to not update the movement in `_keydown`  using the same check as in `_enterframe`:

``` javascript
if (this.disableControls) return;
```

I would be happy to get my hands dirty and submit pull request with a fix, but I wanted to check if is not an intentional design decision.

starwed:
That seems pretty reasonable to me.

(I guess I should point out that at some point, those components will probably be completely replaced.)

VHonzik:
> (I guess I should point out that at some point, those components will probably be completely replaced.)

Is it worth doing then? 

Also I checked the unit tests and they are counting on the current behavior.

starwed:
> Is it worth doing then? 

As long as the work isn't too involved, I think small improvements still make a lot of sense.

> Also I checked the unit tests and they are counting on the current behavior.

The description of `disableControl` claims that the component will ignore keyboard input, so it should be ok to change the tests.  :)

VHonzik:
Fair enough :) I am on it then.

PR #843, opened 2014-12-11
Multiway with disableControl movement change

VHonzik:
`Multiway` component `_movement.x` and `_movement.y` are no longer updated in `_keydown` and `_keyup` methods when `disabledControls` is `true`. Additionally these `_movement` attributes are reset to zero in `disableControl`. 

Updated "disableControl and enableControl" test to reflect the change. 

Fixes issue #842

### DIFF
--- a/src/controls.js
+++ b/src/controls.js
@@ -706,7 +706,7 @@ Crafty.c("Multiway", {
     _speed: 3,
 
     _keydown: function (e) {
-        if (this._keys[e.key]) {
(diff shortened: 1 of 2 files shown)


starwed:
Looks fine to me, but I'll leave the PR open a day or two in case someone who actually uses those components has an opinion.  :)

starwed:
Ok, merging.  Thanks for the PR!

PR #844, opened 2014-12-14
Remove redundant code in "trigger"

sgnix:
The code executed if _this_ is a single element is exactly the same as the
code that runs if _this_ is multiple elements, thus making the first
condition obsolete.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -774,28 +774,7 @@ Crafty.fn = Crafty.prototype = {
      * Unlike DOM events, Crafty events are exectued synchronously.
      */
     trigger: function (event, data) {
-        var h = handlers[event] || (handlers[event] = {});
         // (To learn how the handlers object works, see inline comment at Crafty.bind)
-        if (this.length === 1) {
-            //find the handlers assigned to the entity
-            if (h && h[this[0]]) {
-                var callbacks = h[this[0]],
-                    i, l=callbacks.length;
-                callbacks.depth++;
-                for (i = 0; i < l; i++) {
-                    if (typeof callbacks[i] === "undefined" && callbacks.depth<=1) {
-                        callbacks.splice(i, 1);
-                        i--;
-                        l--;
-                    } else {
-                        callbacks[i].call(this, data);
-                    }
-                }
-                callbacks.depth--;
-            }
-            return this;
-        }
-
         this.each(function () {
             //find the handlers assigned to the event and entity
             if (handlers[event] && handlers[event][this[0]]) {


starwed:
Thanks for the PR!  Unfortunately, doing it this way means creating an anonymous function every time you trigger an event, which isn't acceptable performance-wise.

The duplication of code _is_ very awkward, though.  There's also a similar duplication of code that occurs when triggering a global event.  I've toyed a few times with ways of simplifying the logic, but it was always noticeably slower when I ran perf tests.

It would be pretty easy to just manually loop through the entities rather than calling `each`, which would mean you could completely avoid creating the anonymous function.

I closed the PR, but the problem you were trying to fix is definitely real... feel free to try another tactic.

sgnix:
Sounds good. I would still like to contribute to simplifying and reusing the code. A couple of quick questions: 
- I am new to canvas. Is creating a new anonymous function a real bottleneck in this case? 
- Do you have a benchmarking or performance testing suite that I can use to validate any further code changes?

sbyrnes321:
I agree with @starwed. But the current situation is only slightly better.

For example, if I call `Crafty('ghost').trigger('melt')` right now, it creates a new anonymous function if there is 0 ghosts, or 2 ghosts, or 3 or 4 etc. The function creation is only avoided if there is exactly one ghost. (With this PR, it would always create a function, instead of "almost always".)

@starwed's suggestion to avoid calling `each()` altogether is a great idea. `each()` is just six lines of code, it is not hiding any fancy logic.

@naturalist - creating new objects and functions over and over causes there to be little pauses in execution while the browser is running garbage-collection to free memory. These pauses are very noticeable when playing a game, no matter how fast and efficient the code is in other respects. Another reason to avoid creating new functions all the time: Browsers can compile and optimize functions, but only if it keeps executing the same function over and over, if I understand correctly.

starwed:
> For example, if I call Crafty('ghost').trigger('melt') right now, it creates a new anonymous function if there is 0 ghosts, or 2 ghosts, or 3 or 4 etc. 

One thing that's not necessarily obvious from the code in isolation -- a reference to a specific entity is treated as a collection of length 1.  So things like `Crafty.e("ghost").trigger("melt)` go through the `l=1` path, which actually makes that particular style of invocation quite common.   (It'll happen anytime you assign `x` and `y`, for instance.)

> Is creating a new anonymous function a real bottleneck in this case? 

The problem is that it creates work for the garbage collector, which is an HTML game's worst enemy.  Browsers are getting better, but it's a common cause of those sporadic pauses you'll see.  Consequently, it's important to not create temporary objects or functions in code that'll be called multiple times per frame.  (And this particular bit of code has its tendrils pretty deep in the internals of the engine -- as I mentioned above, it'll often be called for seemingly trivial operations like updating an entities position.)

> Do you have a benchmarking or performance testing suite that I can use to validate any further code changes?

I do have some code I previously used to measure jank, but I haven't updated it recently.   I'll try to fix it up and post that somewhere.

PR #845, opened 2014-12-15
Remove anonymous function creation in "trigger"

sgnix:
Move common code to a private function, in main scope, where it will be compiled and reused in Crafty.trigger. This removes code repetition and (hopefully) increases performance by removing the creation of anon function when 2 or more entities are found.


starwed:
After looking at this and comparing to how Crafty.trigger works, I've decided to do something similar but different -- abstract out just the part that iterates through the callbacks array, rather than the part that iterates through the handlers.  So closing this for now, but a similar patch will be forthcoming.

Thanks for the PR, though!

sgnix:
Sounds good :+1:

PR #846, opened 2014-12-17
implemented multitouch

heniotierra:
This PR implements basic multitouch functionality. The downside is it breaks intercompatibility with mouse dependent components (but they still work with Mouse ofc), although I believe this can be addressed without any hassles.

The **Touch** component enables (through **Crafty.touchDispatch**) detection of the events "TouchStart", "TouchMove", "TouchEnd" and "TouchCancel". "TouchLeave" events are treated as "TouchEnd"s. These basic events allow for implementing more complex events, like swipe, drag or (with some "juggling") pinch.

This PR does not support multitouch over the same entity. That is, if the same entity is touched by two or more fingers at the same time, only the last finger will be valid.

The changes in this PR include:
1. Refactored **Crafty.touchDispatch**;
2. **Crafty.touchDispatch** sends "TouchStart", "TouchMove", "TouchEnd" and "TouchCancel" events;
3. Added **Crafty.fingers** array, which keeps track of currently touching fingers;
4. Added **Touch** component;
5. Added **AreaMap** component, which is used by Mouse and Touch;
6. Added **Button** component, which just selects either Mouse or Touch for an entity based on device;
7. Added **Crafty.findClosestEntityByComponent**  function, which is used internally by Mouse and Touch.
8. Altered **Crafty.mouseDispatch** to use **Crafty.findClosestEntityByComponent**

### DIFF
--- a/src/controls/controls.js
+++ b/src/controls/controls.js
@@ -6,6 +6,7 @@ Crafty.extend({
     mouseObjs: 0,
     mousePos: {},
     lastEvent: null,
+    touchObjs: 0,
     keydown: {},
     selected: false,
 
@@ -41,6 +42,44 @@ Crafty.extend({
         
         Crafty.selected = selected;
     },
+
+    /**@
+     * #Crafty.multitouch
+     * @category Input
+     * @sign public this .multitouch([Boolean bool])
+     * @param bool - Turns multitouch on and off (default is off - false)
+     * @sign public this .multitouch()
+     * Enables/disables support for multitouch feature.
+     * 
+     * If this is set to true, it is expected that your entities have the Touch component instead of Mouse component.
+     * If false (default), then only entities with the Mouse component will respond to touch.
+     * If no boolean is passed to the function call, it will just return whether multitouch is on or not.
+     * 
+     * Notice that the Touch component (and multitouch feature) is incompatible with the Draggable component or other 
+     * mouse dependent stuff.
+     * 
+     * ~~~
+     * @example
+     * ~~~
+     * Crafty.multitouch(true);
+     * 
+     * var myEntity1 = Crafty.e('2D, Canvas, Color, Touch')
+     *    .attr({x: 100, y: 100, w:200, h:200, z:1 })
+     *    .color('black')
+     *    .bind('TouchStart',function(e){ alert('big black box was touched', e); }),
+     *  myEntity2 = Crafty.e('2D, Canvas, Color, Touch')
+     *    .attr({x: 40, y: 150, w:90, h:300, z:2 })
+     *    .color('green')
+     *    .bind('TouchStart',function(e){ alert('big GREEN box was touched', e); });
+     * 
+     * console.log("multitouch is "+Crafty.multitouch());
+     * ~~~
+     * @see Crafty.touchDispatch
+     */
+    multitouch: function (bool) {
+        if (typeof bool !== "boolean") return this._touchHandler.multitouch;
+        this._touchHandler.multitouch = bool;
+    },
     
     resetKeyDown: function() {
         // Tell all the keys they're no longer held down
@@ -67,11 +106,14 @@ Crafty.extend({
      *
      * This method also sets a global property Crafty.lastEvent, which holds the most recent event that
      * occured (useful for determining mouse position in every frame).
+     * 
+     * ~~~
+     * @example
      * ~~~
      * var newestX = Crafty.lastEvent.realX,
      *     newestY = Crafty.lastEvent.realY;
      * ~~~
-     *
+     * 
      * Notable properties of a MouseEvent e:
      * ~~~
      * //(x,y) coordinates of mouse event in web browser screen space
@@ -82,6 +124,7 @@ Crafty.extend({
      * e.mouseButton
      * ~~~
      * @see Crafty.touchDispatch
+     * @see Crafty.multitouch
      */
     mouseDispatch: function (e) {
 
@@ -89,6 +132,7 @@ Crafty.extend({
         Crafty.lastEvent = e;
 
         var maxz = -1,
+            tar = e.target ? e.target : e.srcElement,
             closest,
             q,
             i = 0,
@@ -96,8 +140,7 @@ Crafty.extend({
             pos = Crafty.DOM.translate(e.clientX, e.clientY),
             x, y,
             dupes = {},
-            tar = e.target ? e.target : e.srcElement,
-            type = e.type;
+            type = e.type;     
 
         //Normalize button according to http://unixpapa.com/js/mouse.html
         if (typeof e.which === 'undefined') {
@@ -109,50 +152,7 @@ Crafty.extend({
         e.realX = x = Crafty.mousePos.x = pos.x;
         e.realY = y = Crafty.mousePos.y = pos.y;
 
-        //if it's a DOM element with Mouse component we are done
-        if (tar.nodeName != "CANVAS") {
-            while (typeof (tar.id) != 'string' && tar.id.indexOf('ent') == -1) {
-                tar = tar.parentNode;
-            }
-            ent = Crafty(parseInt(tar.id.replace('ent', ''), 10));
-            if (ent.has('Mouse') && ent.isAt(x, y))
-                closest = ent;
-        }
-        //else we search for an entity with Mouse component
-        if (!closest) {
-            q = Crafty.map.search({
-                _x: x,
-                _y: y,
-                _w: 1,
-                _h: 1
-            }, false);
-
-            for (l = q.length; i < l; ++i) {
-                if (!q[i].__c.Mouse || !q[i]._visible) continue;
-
-                var current = q[i],
-                    flag = false;
-
-                //weed out duplicates
-                if (dupes[current[0]]) continue;
-                else dupes[current[0]] = true;
-
-                if (current.mapArea) {
-                    if (current.mapArea.containsPoint(x, y)) {
-                        flag = true;
-                    }
-                } else if (current.isAt(x, y)) flag = true;
-
-                if (flag && (current._z >= maxz || maxz === -1)) {
-                    //if the Z is the same, select the closest GUID
-                    if (current._z === maxz && current[0] < closest[0]) {
-                        continue;
-                    }
-                    maxz = current._z;
-                    closest = current;
-                }
-            }
-        }
+        closest = Crafty.findClosestEntityByComponent("Mouse", x, y, tar);
 
         //found closest object to mouse
         if (closest) {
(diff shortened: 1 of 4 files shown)


starwed:
That's pretty cool!  I have a couple of concerns, though:
- every time `touchDispatch` is called, it creates about 10 temporary functions.  And it could easily be called a **lot** when you're dragging.  It's probably better to declare those outside of `touchDispatch`; perhaps create something like a `Crafty.touchHandler` object to store all this stuff in.  You could also then throw some of the state tracking (such as `Crafty.fingers`) into that object.
- this actually changes the behavior of "Mouse".  Currently, we have to consult our collision detection engine to figure out if a Canvas entity is clicked on, but for DOM elements we try to use the existing mechanisms in HTML.  I'll leave some line notes specifically about this.
- It should also be an option to just revert back to the old behavior of just using touch events to simulate mouse clicks.

The first and last points should be pretty simple to address, but the middle one might require some thought.

src/controls.js
```diff
@@ -109,50 +115,7 @@ Crafty.extend({
         e.realX = x = Crafty.mousePos.x = pos.x;
         e.realY = y = Crafty.mousePos.y = pos.y;
 
-        //if it's a DOM element with Mouse component we are done
-        if (tar.nodeName != "CANVAS") {
```
starwed: Here, `tar` is the target of the click.  If it's a canvas element, we might have been trying to click on something on the Canvas renderer.  If it's not, then we probably clicked directly on a DOM entity.  So the next block of code searches for the entity associated with the DOM element target.

src/controls.js
```diff
@@ -109,50 +115,7 @@ Crafty.extend({
...
-            while (typeof (tar.id) != 'string' && tar.id.indexOf('ent') == -1) {
-                tar = tar.parentNode;
```
starwed: We might have clicked on the "inside" part of the DOM element, so ascend the DOM tree looking for a node which Crafty has attached an id like `ent3` to.  If we find one, Crafty will treat entity as the target of the click.

src/controls.js
```diff
@@ -109,50 +115,7 @@ Crafty.extend({
...
-            }
-            ent = Crafty(parseInt(tar.id.replace('ent', ''), 10));
-            if (ent.has('Mouse') && ent.isAt(x, y))
-                closest = ent;
-        }
-        //else we search for an entity with Mouse component
-        if (!closest) {
```
starwed: Now, if for one reason or another we didn't find a target up there, use `Crafty.map.search` to consult the collision engine for possible targets.

src/controls.js
```diff
@@ -259,8 +340,84 @@ Crafty.extend({
...
+     * var coords = { x: 455, y: 267 },
+     *     closestText = Crafty.findClosestEntityByComponent("Text", coords.x, coords.y);
+     * ~~~
+     */
+    findClosestEntityByComponent: function (comp, x, y) { 
+        var tar = Crafty.stage.elem,
+            closest, q, l, i = 0, maxz = -1, dupes = {};
+            
+        //if it's a DOM element with component we are done
+        if (tar.nodeName != "CANVAS") {
```
starwed: This block no longer makes sense; since the target is set to just be the stage element (rather than the target of the event), it'll never find anything in this loop.  I think things like DOM text will now respond to clicks based on their explicitly defined width and height, rather than as normal HTML elements.

src/controls.js
heniotierra: Actually this has just been moved into **Crafty.findClosestEntityByComponent**. (Please, see lines 368 - 375.) ~~However, that tar declaration should have been deleted. Is there a problem about that?~~

heniotierra:
Hi starwed! :)

I thought about doing it that way, but I didn't know whether there would make a difference performance-wise between having those functions as temporary variables or having them inside an object. So, you think it's less demanding having the functions as property of an object?
EDIT: I think I see what you mean... being local variables, they have to be redeclared for every dispatch. Ok, I'll fix that!

I agree there should be a way to use one implementation or the other... I'll do that.

~~And I don't think this changes Mouse behavior at all. Please, see my comment to your note.~~

EDIT: Sorry, I accidentally hit the close button.

src/controls.js
heniotierra: Hummm, really? I guess you noticed the "while" right below that? If that's the case, then I'll have to pass the event as argument too, right? I had thought the target would always be in principle **Crafty.stage.elem**.

heniotierra:
@starwed New changes made:
1. Moved all touch related stuff into new **Crafty.touchHandler** object;
2. Made event target be respected;
3. Added **Crafty.multitouch** function, which makes possible enabling/disabling multitouch;
4. Altered documentation to reflect these new changes;
5. Improved documentation, now informing that **Crafty.findClosestEntityByComponent** is only useful for direct use by the user for searching for Canvas entities, not DOM;
6. Removed unneeded conditional clause.

src/controls.js
heniotierra: This problem has been fixed by passing the target inline (I think?).

src/controls.js
heniotierra: Hum... About the user using **Crafty.findClosestEntityByComponent** directly, in case his are DOM entities, will the functions work if he adds a MapArea to the entities he wants to find? Should that be possible? (testing soon)

starwed:
Thanks, that certainly looks better!    What devices have you checked it on?  Do you maybe have a demo I can play around with, to see how it works?

I think you're probably right that adding an areaMap to a DOM element doesn't work properly right now.  While I think there are ways to work around that, there is no need to address that in this PR; we'll just leave existing behavior as is.

heniotierra:
You're welcome.
And yes, you can try the just launched game 'Doce Amianto' here:
http://doceamianto.com.br/gayme/?lang=en

heniotierra:
Oh, and I have checked it on pretty much my mom's phone.  It's a shitty phone, LG Optimus L5 Dual E615.

starwed:
> And yes, you can try the just launched game 'Doce Amianto' here:
> http://doceamianto.com.br/gayme/?lang=en

Sadly, that doesn't ever load for me.  (I've tried a couple of browsers on a couple of different days, I think it gets stuck on waiting for some resources?  No errors are given, anyway.)

heniotierra:
@starwed That's very weird, since it worked fine in every browser and every computer I tried. But, I think that had something to do with files being loaded with an unwanted delay. Hopefuly that's fixed now... Maybe clear your browser's cache and try once more? And a tip: once you see a diamond on the screen, just touch the screen (or click with the mouse, or press any key on the keyboard).

Besides that, what I think might cause loading problems is some kind of server failure(?). However, I guess there should be a logged message, or an error, in that case.

I've been travelling, left with virtually no access to the internet, that's why I didn't answer this thread sooner. Anyway, if you still have problems loading things, in a few days, when I have access to my mom's phone again, I can make a jsfiddle that works.

starwed:
I tried again and while it still didn't load, it did give an error this time:

> Error: Load timeout for modules: src/entities/base/BaseEntity.js src/components/ProgressBar.js src/components/CustomControls.js src/override/HashMap.js src/override/controls.js 
> http://requirejs.org/docs/errors.html#timeout

If I ping `doceamianto.com.br` it's in the 500-1000 ms range, so the issue might be triggered by having a slow path to the server?

Anyway, if you get a jsfiddle working that would also be great.  :)

heniotierra:
Ok, while creating the jsfiddle, I noticed there was a mistake in **handleMove** function, where it'd trigger a TouchStart instead of TouchMove. That's fixed now.

Here's the jsfiddle:
http://jsfiddle.net/heniot/tfasjt0m/5/

Oh, and your help with the feedback on the game was nice, and you were correct, but it was more of a requirejs issue. Using the config waitSeconds for requirejs seems to solve the problem. That modification hasn't been put online yet though. Thanks! :)

heniotierra:
Just so that you know, the fix that I mentioned for the loading problem related to requirejs in the game Doce Amianto has already been put online. http://doceamianto.com.br/gayme/?lang=en (I hope it works now) You can checkout the controls on a mobile... They must show up at the bottom of the viewport, but can only be seen in fullscreen mode.

starwed:
Cool, thanks!  I'll try to see how this works on an Android tablet.  (Sorry for taking a while to look at it, but I had a lot going on over the holidays.)

starwed:
Ok, got the fiddle working!  (And hooked up the Firefox dev tools to it.)

It looks to me like there's a bug (or at least unexpected behavior) -- if I touch one of the boxes and then hold my finger in place, then touching anywhere with a second finger else also registers as a touch of that box, with coordinates equal to the first finger.

src/controls.js
```diff
@@ -196,71 +195,322 @@ Crafty.extend({
...
+    
+    /**@
+     * #Crafty.touchHandler
+     * @category Input
+     * Manager for the touches in mobile devices. Contains the fingers array, which associates a touch to an
+     * entity. There should be no need to mess with this.
+     *
+     * @see Crafty.multitouch
+     */
+    touchHandler: {
```
starwed: If you'd prefer, you can just prefix this with an underscore to indicate it's private, and just leave regular comments explaining the code.  Like you say, there should be no need to mess with it, so we don't _need_ to expose it to users.

heniotierra:
Hum! Thanks for testing! That problem is probably because of line 278, which should be inside the if clause that follows it, line 280. Doing that and adding a type check in line 284 should fix the problem... What do ya say? Should I make these changes?

starwed:
Go for it!  You might also want to see if you can mock up some simple tests by manually triggering touch events in PhantomJS.  (I haven't looked to see how that works.)

_e:_ Looks like you can [use jQuery](http://stackoverflow.com/a/10406919/460138) for that, since it's included in the testing files.

starwed:
Oh, and to add -- once you fix that bug I noticed (and add tests if you plan on it) we can go ahead and land this.

I suspect that in the future we'd need to change how TouchMove works, to allow it to work properly with Draggable entities.  But that's not something that needs to be addressed right now.  :)

heniotierra:
Alright, I'll do as you say, also trying to add tests.

Do you think TouchMove would have to be changed to allow for Draggable entities to work? Wouldn't we have to just adapt the Draggable component? I never used it or analysed its code much.

heniotierra:
Here's my test code using jquery, but it just doesn't work: no events seem to be triggered, thus all `equal`s fail (except the last one, but that's expected). I'm probably doing something wrong... 
https://gist.github.com/heniotierra/4cf4cbfb49628928f63c

I'm not sure jQuery is the right tool for the job... I thought of trying hammerjs too, but then found hammerjs simulates touch from mouse events.

EDIT: Searching for 'phantomjs touch events' I found this open issue requesting the implementation of touch events https://github.com/ariya/phantomjs/issues/11571; also found this node extension that can simulate a list of touches https://github.com/gardr/mock-phantom-touch-events

EDIT2: Tried mock-phantom-touch-events, but still no touch events are detected. The code using it is here https://gist.github.com/heniotierra/fcfd6cf011a66f11561f

This is the output of the test:

> > Controls - Multitouch simulation
> > Message: Two entities received TouchStart
> > Actual: 0
> > Expected: 2
> > at file:///home/henio/public_html/Crafty/tests/lib/qunit.js:556
> >   at file:///home/henio/public_html/Crafty/tests/controls.js:39
> >   at file:///home/henio/public_html/Crafty/tests/lib/qunit.js:203
> >   at file:///home/henio/public_html/Crafty/tests/lib/qunit.js:361
> >   at process (file:///home/henio/public_html/Crafty/tests/lib/qunit.js:1453)
> >   at file:///home/henio/public_html/Crafty/tests/lib/qunit.js:479
> > 
> > Controls - Multitouch simulation
> > Message: Four fingers currently touching stage
> > Actual: 0
> > Expected: 4
> > at file:///home/henio/public_html/Crafty/tests/lib/qunit.js:556
> >   at file:///home/henio/public_html/Crafty/tests/controls.js:40
> >   at file:///home/henio/public_html/Crafty/tests/lib/qunit.js:203
> >   at file:///home/henio/public_html/Crafty/tests/lib/qunit.js:361
> >   at process (file:///home/henio/public_html/Crafty/tests/lib/qunit.js:1453)
> >   at file:///home/henio/public_html/Crafty/tests/lib/qunit.js:479
> > 
> > Controls - Multitouch simulation
> > Message: One entity received TouchEnd
> > Actual: 0
> > Expected: 1
> > at file:///home/henio/public_html/Crafty/tests/lib/qunit.js:556
> >   at file:///home/henio/public_html/Crafty/tests/controls.js:45
> >   at file:///home/henio/public_html/Crafty/tests/lib/qunit.js:203
> >   at file:///home/henio/public_html/Crafty/tests/lib/qunit.js:361
> >   at process (file:///home/henio/public_html/Crafty/tests/lib/qunit.js:1453)
> >   at file:///home/henio/public_html/Crafty/tests/lib/qunit.js:479
> > 
> > Controls - Multitouch simulation
> > Message: One finger currently touching
> > Actual: 0
> > Expected: 1
> > at file:///home/henio/public_html/Crafty/tests/lib/qunit.js:556
> >   at file:///home/henio/public_html/Crafty/tests/controls.js:46
> >   at file:///home/henio/public_html/Crafty/tests/lib/qunit.js:203
> >   at file:///home/henio/public_html/Crafty/tests/lib/qunit.js:361
> >   at process (file:///home/henio/public_html/Crafty/tests/lib/qunit.js:1453)
> >   at file:///home/henio/public_html/Crafty/tests/lib/qunit.js:479
> > 
> > Controls - Multitouch simulation
> > Message: Two entities received TouchEnd
> > Actual: 0
> > Expected: 2
> > at file:///home/henio/public_html/Crafty/tests/lib/qunit.js:556
> >   at file:///home/henio/public_html/Crafty/tests/controls.js:51
> >   at file:///home/henio/public_html/Crafty/tests/lib/qunit.js:203
> >   at file:///home/henio/public_html/Crafty/tests/lib/qunit.js:361
> >   at process (file:///home/henio/public_html/Crafty/tests/lib/qunit.js:1453)
> >   at file:///home/henio/public_html/Crafty/tests/lib/qunit.js:479

heniotierra:
@starwed I was using the extension in a wrong manner, now I got to use it correctly and it works, however touchDispatch still won't execute. If I do `Crafty.stage.elem.addEventListener("touchstart", function(){ console.log("this is executed!") }))` this is executed, but not touchDispatch. What do you think that could mean? The code in the last link was updated: https://gist.github.com/heniotierra/fcfd6cf011a66f11561f

I'm not sure the coordinate values are correct, but the events should be caught anyway(?).

starwed:
Thinking about it, you should just mock up an event object and call `touchDispatch` directly.  That way, you can test that the logic is correct, but don't need to muck about with the actual DOM events at all!  However, remember that the stage offset (as well as scroll position and some other stuff) is subtracted from the event coordinates you pass to the dispatch functions.

(I actually did succeed in mocking up some tests for mouse clicks, using `createEvent`, `initMouseEvent`, and `dispatchEvent` -- see [this stackoverflow answer](http://stackoverflow.com/a/16509592/460138) for more.  You do need to keep in mind that the event coordinates get clamped to integers doing this way, which means that it's very tricky to account _exactly_ for the stage offset and target a specific, integer crafty coordinate.)

(_e_: Oh, and the [code](http://pastebin.com/NScQZZX1) I was using to simulate mouse behavior before I decided calling the dispatch functions directly made more sense.)

heniotierra:
@starwed It's working now, but I ignored the coordinates (couldn't find the correct values for them, even considering offset and scroll), that is, the test only checks the fingers. If you don't mind, I used the phantonjs extension for it's able to simulate a TouchList (https://developer.mozilla.org/en-US/docs/Web/API/TouchList).
One more problem now though: after preparing to send all the changes (merging the latest commits made to the upstream repo), and testing again, this was the output: 
$ npm test

> craftyjs@0.6.3 test /home/henio/public_html/Crafty
> grunt check
> 
> Running "browserify:debug" (browserify) task
> Warning: Cannot find module 'brfs' from '/home/henio/public_html/Crafty' Use --force to continue.
> 
> Aborted due to warnings.
> npm ERR! Test failed.  See above for more details.
> npm ERR! not ok code 0

starwed:
> One more problem now though: after preparing to send all the changes (merging the latest commits made to the upstream repo), and testing again, this was the output: 

Sounds like you have a merging conflict -- when I landed the WebGL stuff, it added the brfs requirement.  Your branch was from before that point.  Probably when you added the phantomjs extension, you modified the same file, so you'll have to be careful that your change doesn't just overwrite mine.

You should update your develop branch, and then rebase your branch against that (`rebase develop` from your feature branch).  It'll then notify you about merge conflicts, which you can resolve.

Alternatively, you can just push your changes, and I can resolve the conflict manually on my end.  Your call!  :)

heniotierra:
@starwed  Tis done, sir!

heniotierra:
Another thing that could be implemented in the future is support for touching a same entity with more than one finger... Idk how useful that would be, though... maybe it's not absolutely needed for, say, pinch effect? I mean, someone might get around this shortcome with some creativity and implement a 'pinch touch', with multiple entities attached... not the most elegant solution I guess, but still...

Also, I updated the jsfiddle with the latest fix:
http://jsfiddle.net/tfasjt0m/7/

starwed:
I'll test the updated version later on today, but a quick (but important) question -- is `mockTouchEvents.js` your own code or from elsewhere?  In the latter case you should note the license.

heniotierra:
Not my code, sure, I forgot to inform the license. I'll update it.

EDIT: done. :)

starwed:
I still see the same issue with the updated fiddle: while holding a finger over one entity, touching anywhere else on the screen retriggers that event.

Also, if I touch an entity, then touch it again with a second finger, it triggers twice the second time -- and then lifting up both fingers creates an error: `this.fingers[0] is undefined`.  Probably the same issue.

src/controls.js
```diff
@@ -201,63 +201,308 @@ Crafty.extend({
...
                 e.returnValue = false;
             }
+    },
+    
+    _touchHandler: {
+        fingers: [], // keeps track of touching fingers
+        multitouch: false,
+        
+        handleStart: function (e) {
+            var touches = this.getTouchesFromEvent(e);
```
starwed: Found the bug.  `getTouchesFromEvent` default to returning `e.touches`.  This will include _every_ touch event, not just the ones recently updated.  The bug happens as follows:
- Touch entity A; event triggers properly.
- Touch entity B.  `handleStart()` is called, and retrieves the list of all current touches.  It then triggers a "TouchStart" event for all of them.  Since this includes the original touch on entity A, the bug manifests.

Something similar will happen with the other event handlers.

src/controls.js
```diff
@@ -201,63 +201,308 @@ Crafty.extend({
...
+                
+                if (ent == entityToFind) {
+                    return i;
+                }
+            }
+            return -1;
+        },
+            
+        getTouchesFromEvent: function(e){
+            var touches;
```
starwed: I tried [switching the order here](http://jsfiddle.net/tfasjt0m/8/); defaulting to returning the changedTouches.  This fixes the particular bug I noticed. However, since you manually compare the touch event list each time to determine if a finger is lifted, it then produces false "TouchEnd" events, as `changedTouches` deliberately doesn't include previous events unless they've, well, changed.

I think the correct solution is to always use `changedTouches`, and then change how you handle `touchend` events -- for that event, `changedTouches` will just directly list the touches which are ending!  :)

heniotierra: Thank you for hunting this bug down! I think the info I found about touches and changedTouches led me to believe they were the same thing, but for different browsers. Anyway, I will find some time to look at it soon!

starwed:
Oh, some things about the test you added (sorry to add so many comments here, it's just that it's a big feature!  :) )
- make sure it doesn't fail in non-touch capable browsers.  It passes if I run the phantomJS tests, but fails in desktop Firefox.  If the test isn't applicable for a given browser, it should just be skipped!  :)
- You can get the touches positioned over the entities by adding the offset for Crafty.stage.  So instead of `[100, 80]`, use `[100 + Crafty.stage.x, 80 + Crafty.stage.y]`.  (Or more reasonably, have a function that applies the offset.)  That'll let you make tests for the bugs I found.

heniotierra:
> make sure it doesn't fail in non-touch capable browsers. 

Right, I'll keep that in mind...

> You can get the touches positioned over the entities by adding the offset for Crafty.stage.

I did that while coding the test, but it didn't work... I used jQuery .offset function (http://api.jquery.com/offset/) for `Crafty.stage.elem`. Also, I tried to reset the scroll to zero, to no avail. Maybe there are other things to consider?
EDIT: I will try to add `Crafty.stage.x` and `Crafty.stage.y` directly.

heniotierra:
@starwed Hey there! I got some time today to deal with this. I think everything is fine now. :)

> make sure it doesn't fail in non-touch capable browsers.

Check. This test only runs if browser is PhantomJS, since mock-phantom-touch-events is a PhantomJS plugin.

> You can get the touches positioned over the entities by adding the offset for Crafty.stage.

Check.

I had to make a little change to mock-phantom-touch-events so that we can specify the `touch.identifier`s.

Also, the handling functions are only using `changedTouches` now, and I fixed `.handleEnd`, and finally I added a test for the bug you mentioned (see `touchStart2` variable).

starwed:
Ok, when I play around with an updated fiddle, looks like everything is working properly!  And the tests seem to work, too.

I'll go ahead and merge this.  Nice work!

starwed:
I've discovered something kind of weird, thought @heniotierra might have some input.

In writing some tests for Crafty.viewport.reset, I discovered that there were some failures caused by a previous test calling `Crafty.stop()`.  But when I removed that function, I start getting failures in the multitouch tests.

So something about the tests that were part of this PR _depend_ on Crafty.stop() being previously called in the test suite when running in PHantomJS.  You can reproduce this by commenting out the test in core.js:L336; when running `grunt check`, the multitouch test fail.  Weirdly, they still work when run in the browser....

Any ideas why this might be so?

heniotierra:
@starwed  None from the top of my head. Maybe some kind of inconsistency in nodejs? I will have a look at it later on today or tomorrow, then I'll let you know!

starwed:
Ah, figured it out. You call `addEventListener` in the test.  It should already be set by Crafty, of course, but `Crafty.stop()` was removing the event listeners.  When I removed the earlier test that called `stop()`, it meant "touchstart" and "touchstop" were triggering twice.

Since `Crafty.stop()` _also_ removes the function bound to "Load", a second call to `Crafty.init()` won't work as planned.   This is basically the same issue I was having with the viewport tests; we do some stuff on initial setup that isn't replicated on a second call to `init()`.

heniotierra:
I guess there isn't an easy way to reset the whole engine after each test module? Should `init` do that?

starwed:
> I guess there isn't an easy way to reset the whole engine after each test module? Should init do that?

Probably!  For now I'm just going to remove the test calling `stop()`, since it has so many unwanted side effects.

PR #848, opened 2014-12-22
Allow alternate easing functions

starwed:
This allows you to specify alternate easing functions, either by giving the name of a built-in function (such as "smoothStep" or "easeInOutQuad"), or by directly passing a function.  Omitting the function defaults to linear behavior.

Currently there's a very limited selection of built-in functions, but it's trivial to add more!

The base functionality is implemented in Crafty.easing.  The "Tween" component, `viewport.zoom()`, and `viewport.pan()` all now support these variant easings.  

This patch also adds a couple of simple tests.

### DIFF
--- a/src/animation.js
+++ b/src/animation.js
@@ -1,9 +1,41 @@
 var Crafty = require('./core.js'),
 	document = window.document;
 
-Crafty.easing = function(duration) {
+
+/**@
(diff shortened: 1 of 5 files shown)


starwed:
Going to go ahead and land this.

Issue #850, opened 2015-01-09
Draggable does not handle zero length directions.


BillDorn:
I am not sure if this is by design but currently if the direction passed to `dragDirection` is of length zero it results in a division by zero when attempting to normalize the vector. I feel it is more natural to allow zero length directions and simply have the entity not move when dragged.

To implement the proposed behavior a simple check can be added to `dragDirection` and an additional check could be added to `_onDrag` to avoid unnecessary computations.

starwed:
Yup, sounds like a bug to me!  

I'll definitely try to fix that before the next release, but of course anyone should feel free to submit a PR.

PR #851, opened 2015-01-10
Draggable zero length direction checks

BillDorn:
I have added checks for zero length vectors in `dragDirection` and `_onDrag` of the Draggable component.
This fixes bug #850.


starwed:
Thanks!  Looks good to me, so merging.

PR #852, opened 2015-01-12
Remove space character that affected layout

starwed:
There was a non-breaking space character after the jQuery script tag.

This was actually affecting the layout of the page, which I noticed when attempting to trigger mouse events.  Ironically, probably one of the more frustrating bugs to diagnose related Crafty to date.  As you might imagine, "is there an invisible character in the source code" was not high on my list of things to check when layout was being weird.  :)  (And while I have Sublime set to show spaces, it didn't show this particular variant!)


kevinsimper:
White spaces gives some weird problems sometimes :)

PR #853, opened 2015-01-16
Fix a error in doc

XiongLiding:
Second param of Crafty.timer.steptype should be maxTimeStep


starwed:
Thanks for catching that!

PR #854, opened 2015-01-20
Impose hierarchy

starwed:
Things were getting a bit unwieldy, so this organizes things based on folders.  Some things are not a perfect fit for one directory; that can always be cleaned up later.  The folders are
- graphics/ -- anything to do with rendering to the screen.   (This should maybe be subdivided further, by DOM/Canvas etc)
- isometric/ -- For the couple of files related to isometric display
- spatial/ -- The 2D components, collisions, and similar functionality
- controls/ -- Various input functionality
- sound/ -- Just sound.js right now
- debug/ -- Just DebugLayer.js right now
- core/ -- core functionality that isn't specific to the above

It was necessary to change the paths of some of the require statements, of course.

I also split off the "Image" component from `drawing.js` (there was no reason for it to be there) and renamed files such that they were consistently lower case.


starwed:
Looks like if this change goes in I'd have to merge it manually, but I've checked locally that it won't interfere with other PRs (like #846) -- they rebase just fine on top of this.  Git is smart about applying changes to the correct files.

kevinsimper:
I have thought the same, but would it be an idea to split up the components into a `./components` folder?

starwed:
Yeah, I'm kind of ambivalent about that, since it's nice to have components split out, but having them all in one place is a bit weird.  Possibly each top level directory could have a separate `components` folder?

starwed:
This doesn't affect functionality at all, so I'm going to go ahead and merge as-is.  (We can always split up the components into their own files later.)

starwed:
Looks like when I renamed the files lowercase, I didn't update all of them in the `crafty.js` source file.  This worked fine on my mac, which I guess doesn't care about case, but failed on TravisCI.  (This sort of inconsistency is actually why folk recommend just using lower-case everywhere!)  I'll fix that and push the fix.

PR #856, opened 2015-01-26
Preventing default keyboard actions needs to be on original

docmarionum1:
stopPropagation, cancelBubble, etc. need to be used in the original event, not the Crafty copy of it, e.

### DIFF
--- a/src/controls/controls.js
+++ b/src/controls/controls.js
@@ -331,15 +331,15 @@ Crafty.extend({
         //Among others this prevent the arrow keys from scrolling the parent page
         //of an iframe hosting the game
         if (Crafty.selected && !(e.key == 8 || e.key >= 112 && e.key <= 135)) {
-            if (e.stopPropagation) e.stopPropagation();
-            else e.cancelBubble = true;
+            if (original.stopPropagation) original.stopPropagation();
+            else original.cancelBubble = true;
 
             //Don't prevent default actions if target node is input or textarea.
-            if (e.target && e.target.nodeName !== 'INPUT' && e.target.nodeName !== 'TEXTAREA') {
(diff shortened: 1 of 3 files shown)


starwed:
Looks ok to me, but I'll wait a bit before merging in case someone else has a comment. 

If there's a specific easy to replicate issue this fixes, a test would be welcome!  :)

docmarionum1:
I noticed this issue when I embedded my game in an iframe on the Global Game Jam site.  Up/down arrow keys were making the whole page scroll.  Doing the same test manually is simple enough.  But I couldn't get a real automated test for this working because there seem to be some issues around javascript generated keyboard events and initializing the properties correctly.

If anyone else can figure that out, you could recreate the original case by making sure arrowkey input doesn't alter window.pageYOffset.

In lieu of that I made a test with a mock event object.  I didn't see an appropriate file in tests so I created a new one, but let me know if you'd like it moved elsewhere.

starwed:
Thanks!  #846 also introduces `tests/controls.js`, so that seems a good location to me.

starwed:
Hmm, the test you added seems to pass in PhantomJS, but fails when run directly in Firefox and Chrome.

docmarionum1:
I noticed it would fail in browser when only running that one test.  It expected the canvas to exist which it happened to do when run with all the other tests.  Which "target" isn't important so I just set it to document and it seems fine in FF/Chrome for me now.

starwed:
I had to pull this down locally to resolve the conflicts with the other PR that affects `controls.js`.  I went ahead and squashed it down to a single commit when I did so, then merged it in. 

Thanks for the PR!  (Also, nice KoL avatar :) )

PR #858, opened 2015-01-28
Keep the polygon passed to Collision unchanged

Chaosed0:
A common case for this fix is when you make the same polygon and use it with two Collision components. In my case, I discovered it when I created a Polygon rendering component and passed the same Crafty.polygon to both the Collision and Polygon components. This resulted in the polygon component being rendered in the wrong place.

I also added the Crafty.polygon.clone() method to copy the polygon around.

### DIFF
--- a/src/spatial/collision.js
+++ b/src/spatial/collision.js
@@ -99,6 +99,9 @@ Crafty.c("Collision", {
                 //convert args to array to create polygon
                 var args = Array.prototype.slice.call(arguments, 0);
                 poly = new Crafty.polygon(args);
+            } else {
+                //Clone the poly so we don't modify it for anything else that might be using it
+                poly = poly.clone();
             }
             // Check to see if the polygon sits outside the entity, and set _cbr appropriately
             // On resize, the new bounds will be checked if necessary
@@ -118,7 +121,7 @@ Crafty.c("Collision", {
         }
 
         // Finally, assign the hitbox, and attach it to the "Collision" entity
-        this.map = poly.clone();
+        this.map = poly;
         this.attach(this.map);
         this.map.shift(this._x, this._y);
         this.trigger("NewHitbox", poly);
--- a/src/spatial/collision.js
+++ b/src/spatial/collision.js
@@ -118,7 +118,7 @@ Crafty.c("Collision", {
         }
 
         // Finally, assign the hitbox, and attach it to the "Collision" entity
-        this.map = poly;
+        this.map = poly.clone();
         this.attach(this.map);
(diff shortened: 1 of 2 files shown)


starwed:
Hmm, does this actually do what you want?  `.clone()` is just creating a new polygon, and the code for polygon simply assigns the passed array of points to `this.points`.  It doesn't ever create a new array; the cloned polygon would end up sharing it's set of points with the original.

I'd agree that polygon should have a clone method, but I'm not so certain about collision _automatically_ cloning the passed points.  It should be something you opt into; otherwise, it can be a possible performance hit.

Chaosed0:
Oh, yeah, you're right - I was in too much haste while writing up the patch. I'll submit a new patch later today that actually copies the points.

You're right, it would be a performance hit. However, some really unintuitive things can happen with the current behavior of Collision. In my case, with my Polygon rendering component, the polygon was getting rendered way off since it was getting shifted twice. I think the performance hit would be an acceptable tradeoff.

There's two alternatives I can come up with offhand, the lesser of which is to simply change the documentation to note that Collision modifies the polygon passed. The second one is to, instead of shifting the actual polygon, add `this.x` and `this.y` to the polygon points every time they're used. Let me know if either of those are preferable and I'll change the patch accordingly.

starwed:
Ok, after looking through `.collision()` here's my take on this.

The method has 3 signatures:
- If you pass a polygon, it assigns that to `this.points`.
- If you pass in a list of points as separate arguments, it creates a polygon from that.
- If you pass in nothing, it creates a polygon that matches the entity's dimensions.

So long as the user can take the second option, they can avoid creating a temporary object, so everything would be fine.

The way this PR does things would actually clone the polygon even in cases when it was created earlier in the function, so it should probably go further up.  It should also clone the polygon before its rotated!  :)   I'll add a line note.

src/spatial/collision.js
```diff
@@ -118,7 +118,7 @@ Crafty.c("Collision", {
         }
 
         // Finally, assign the hitbox, and attach it to the "Collision" entity
-        this.map = poly;
+        this.map = poly.clone();
```
starwed: This would be the wrong place to clone.  It should probably be in an else branch on line 102, so that the logic is as follows:

``` javascript
if (arguments.length > 1 ) { 
   // create polygon from arguments
} else {
   // clone polygon from single argument
}
```

Chaosed0:
Thanks for the comments, changed accordingly. Apologies for the novice mistakes!

starwed:
Thanks for the PR!

I pulled this down and manually squashed it to one commit.

PR #859, opened 2015-01-30
Organize graphics

starwed:
This is a pretty drastic reorganization of the rendering/viewport/stage code.  The different pieces for handling DOM and Canvas rendering were split up over several files and mixed with each other in a confusing way.

This creates two new objects which have more clearcut responsibilities, and are designed to be clearly parallel with themselves and the webgl rendering code.
- `Crafty.canvasLayer` for init and rendering on the canvas element
- `Crafty.domLayer` for init and rendering on the div element we use to place DOM entities

Several objects are shuffled around or renamed: `Crafty.DOM` becomes `Crafty.domHelper`, `Crafty.rectManager` absorbs `Crafty._rectPool` and gets its own file, and `Crafty.DrawManager` is just gone!

There were very few actual changes to the code logic; it was mostly just renaming things and shifting methods from one object to another.  I did make it so the DOM layer is only created when you actually have DOM entities, and slightly alter how Crafty determines the window size, since that was weirdly tangled up in that init code.

The main impulse for this reorganization wasn't just to clean things up, but to prepare the way for a more abstract approach to creating layers -- we want to be able to create foreground/background/UI layers, for instance, and dictate the render type of each.  This is a good first step towards that.


starwed:
Hmm, while this passes tests, there's some issue coming up when I try the built file in more complicated examples.  I'll try to figure out what's up and push a fix.  (And also a test that catches whatever is going on!)

starwed:
There we go, was just a couple of dumb renaming mistakes.

Unfortunately the way things are setup right now, there's no easy way to check that the rendering logic has been called correctly.  Luckily the test I had for checking WebGL rendering pokes at most corners of that, so it revealed my silly mistakes right away.  :)

(The real solution is probably to have some screenshot comparison tests, but that sounds complicated!)

starwed:
I'll go ahead and merge this.  I will not be shocked if there's some corner case that's been broken, but everything seems to work for me.

kevinsimper:
Very nice! :+1:

PR #861, opened 2015-02-01
Remove test for Crafty.stop()

starwed:
Unfortunately, running `Crafty.stop()` destroys a few types of bound events that aren't set back up on `Crafty.init()`.

Long term, that should be fixed, but for now it shouldn't stop other tests from passing; it was interfering with some tests for `Crafty.viewport` I was trying to write, and there was a workaround in `tests/controls.js` that would break if `stop()` _wasn't_ called.


starwed:
This is pretty small, and only affects the tests, so I'm going to just merge right away.

PR #862, opened 2015-02-01
Reset the pan to 0 on Crafty.viewport.reset();  add tests

### LINKED ISSUE #717: Crafty.viewport.reset() does not pan back to 0. Should it?
The documentation implies that if you pan the viewport (e.g. `Crafty.viewport.x += 100`), then running `Crafty.viewport.reset()` should undo that by panning back. But that is not what happens. It does not pan back.

http://jsfiddle.net/T9Mp9/ (using [Crafty 0.6.2](https://github.com/craftyjs/Crafty/releases/tag/0.6.2) )

Is this unclear documentation, or is it a code error?

starwed:
The reset function should reset the position of the viewport; for some reason it doesn't right now.

This adds tests for that behavior, as well as a couple other aspects of `viewport.reset()`.  I went through and streamlined some of the other viewport tests at the same time.

Fixes #717

### DIFF
--- a/src/graphics/viewport.js
+++ b/src/graphics/viewport.js
@@ -717,6 +717,9 @@ Crafty.extend({
         reset: function () {
             Crafty.viewport.mouselook("stop");
             Crafty.trigger("StopCamera");
+            // Reset viewport position and scale
+            Crafty.viewport.scroll("_x", 0);
+            Crafty.viewport.scroll("_y", 0);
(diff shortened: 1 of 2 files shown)


starwed:
Pretty simple change to Crafty, so merging.

This is technically a breaking change if folk were depending on the previous viewport behavior, but I doubt that was common.

PR #863, opened 2015-02-01
No globals

starwed:
There were a few global variables being created in both tests and the actual Crafty source code.

I found these by running the QUnit tests in browser, where you get an option to check for globals.  The next version of `grunt-contrib-qunit` will have an option that'll let us do that from grunt as well.


starwed:
Low impact change, so merging.

PR #864, opened 2015-02-01
touchcancel handler fix

heniotierra:
TouchCancel event had been overlooked. This fix properly handles `touchcancel`s, triggering "TouchCancel".


starwed:
e: Odd, somehow merged/commented on the wrong PR.  This one is a fine fix too, though, so no problems!

PR #866, opened 2015-02-01
Add `required` field for components

starwed:
This makes it easier for one component to require additional components.  You'd use it like:

``` javascript
Crafty.c("BlueSquare", {
   required: "2D, Canvas, Color",
   init: function() {
       this.color("Blue");
   }
}
```

This is a pretty simple change that should improve code readability.

I'd have made it requires instead of required, but that would interfere with the `requires()` method.  We could perhaps special case these sort of special component properties to not actually be extended.


kevinsimper:
Looks good!

PR #867, opened 2015-02-01
Reduce init garbage

starwed:
I've always noticed a large pause/stutter when Crafty starts up.  After looking through the various bits of init code, I found a couple of places where we probably can do less object/array creation on startup.
- Don't create setter/getter objects from scratch each time on 2D objects
- Don't create an array for iteration when adding components

I haven't yet actually _tested_ this to see if it makes a difference in practice, but I'll try to do that later and post some results.  :)

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -220,20 +220,14 @@ Crafty.fn = Crafty.prototype = {
      * ~~~
      */
     addComponent: function (id) {
-        var comps = [],
+        var comps,
             comp, c = 0;
 
         //add multiple arguments
-        if (arguments.length > 1) {
-            var i = 0;
-            for (; i < arguments.length; i++) {
-                comps.push(arguments[i]);
-            }
-            //split components if contains comma
-        } else if (id.indexOf(',') !== -1) {
+        if (arguments.length === 1 && id.indexOf(',') !== -1) {
             comps = id.split(rlist);
         } else {
-            comps.push(id);
+            comps = arguments;
         }
 
         //extend the components
--- a/src/spatial/2d.js
+++ b/src/spatial/2d.js
@@ -144,9 +144,9 @@ Crafty.c("2D", {
     _changed: false,
 
     
-
-    _define2DProperties: function () {
-        Object.defineProperty(this, 'x', {
+    // Setup   all the properties that we need to define
+    _2D_property_definitions: {
+        x: {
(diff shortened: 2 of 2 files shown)


src/core/core.js
```diff
@@ -220,20 +220,14 @@ Crafty.fn = Crafty.prototype = {
...
-            for (; i < arguments.length; i++) {
-                comps.push(arguments[i]);
-            }
-            //split components if contains comma
-        } else if (id.indexOf(',') !== -1) {
+        if (arguments.length === 1 && id.indexOf(',') !== -1) {
             comps = id.split(rlist);
         } else {
-            comps.push(id);
+            comps = arguments;
```
mucaho, line 230: be careful the `arguments` array [is not a real array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments) - doesn't support `push` for example, but in this case it's fine I guess.

src/spatial/2d.js
```diff
@@ -240,8 +239,15 @@ Crafty.c("2D", {
...
             enumerable: true
-        });
-        Object.defineProperty(this, '_visible', {enumerable:false});
+        },
+        _visible: {enumerable:false}
+
+    },
+
+    _define2DProperties: function () {
+        for (var prop in this._2D_property_definitions){
```
mucaho, line 248: ~~does this only iterate over the properties above, or also over the properties in the property chain, if there are any?~~

mucaho:
Looks good!

src/core/core.js
starwed, line 230: Yeah, it's fine here, but I really should have added a comment about that assumption.

starwed:
I don't actually see much/any improvement on startup, sadly.  The tests I was using are a bit ad hoc (and like two years old), though, and I should see if Chrome or Firefox's dev tools/extensions have nicer ways of detecting GC pauses these days.

mucaho:
It can't hurt though, can it? :)  
Just to be on the safe side, I would recommend adding a second entity to e.g. `tests\2d.js\position` and then testing if changing of one's properties affects the other's.

mucaho:
Can we merge this?

starwed:
Sure, it certainly doesn't _hurt_ anything, and probably makes the code a bit more readable.

PR #868, opened 2015-02-02
Define that the repository should be transformed with brfs

kevinsimper:
With the new WebGL component we need to define that the repository needs to be transformed by brfs to work with browserify.

https://github.com/substack/node-browserify#browserifytransform

How to test:

Install with npm `npm install kevinsimper/Crafty#declare-transform`
Write a simple game with Crafty `game.js`

```
var Crafty = require('craftyjs')

Crafty.init(600, 300);
Crafty.background('rgb(127,127,127)');
//Paddles
Crafty.e("Paddle, 2D, DOM, Color, Multiway")
    .color('rgb(255,0,0)')
    .attr({ x: 20, y: 100, w: 10, h: 100 })
    .multiway(4, { W: -90, S: 90 });
```

Run with beefy https://github.com/chrisdickinson/beefy
`$ beefy game.js`

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
   "jsfiddle": "disable",
   "homepage": "https://github.com/craftyjs/Crafty",
   "dependencies": {
+    "brfs": "1.0.1"
   },
   "devDependencies": {
     "grunt-contrib-connect": "^0.8.0",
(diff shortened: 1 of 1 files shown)


starwed:
Looks good, but I think there are some unrelated changes to package.json?  (The "title" and "jsfiddle" fields have been deleted.)

kevinsimper:
Ahh, i just deleted them because they are obsolete :) I will add them again and remove them later

Jsfiddle is from the deprecated craftycomponent.com webpage and so is title.

starwed:
Oh, it would probably have been fine to do that in a single commit, just wanted to make sure they weren't removed by accident!

PR #869, opened 2015-02-05
Abstract out common event operations

starwed:
There was a lot of duplicated logic between the various event-related functions.

This patch tries to abstract out the operations they had in common.  The basic idea is that a lot of the logic for bind, unbind, and trigger had to do with iterating through the callback array.   So this defines a set of operations for doing so, and then implements bind/unbind/trigger in terms of those operations.

Other features of this patch
- Give the Crafty object an "id"; `Crafty[0] == "global"` now.  Previously this had to be special cased in the event logic.  (It still is in one place, unfortunately.)
- Create a function `addCallbackMethods` which extends an object with the new low level callback methods.  The object itself still needs to provide bind/unbind/trigger/etc
- Iterate through collections of objects, rather than calling `.each()` with a new anonymous function.

In addition to cleaning up some of the existing code, this will make it much easier to create new objects which can listen to events.  I have a partial idea for how to make that even eaiser, but it'll need some work!

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -163,12 +163,14 @@ Crafty.fn = Crafty.prototype = {
 
             //update from the cache
             if (!this.__c) this.__c = {};
+            if (!this._callbacks) addCallbackMethods(this);
 
             //update to the cache if NULL
             if (!entities[selector]) entities[selector] = this;
             return entities[selector]; //return the cached selector
         }
 
+        addCallbackMethods(this);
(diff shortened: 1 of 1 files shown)


starwed:
Since there are no comments, I'll just go ahead and merge this.

(The patch doesn't add any tests, since we already have a pretty good set for events -- when I was working on this, it was nice to have test suite catch mistakes!)

PR #873, opened 2015-02-15
Add setTextAlign() to Text

Chaosed0:
Here's a simple patch that allows users to set textAlign on the text drawn by the Text component. Makes it much easier to center text.

### DIFF
--- a/src/graphics/text.js
+++ b/src/graphics/text.js
@@ -30,6 +30,7 @@ Crafty.c("Text", {
     defaultFamily: "sans-serif",
     defaultVariant: "normal",
     defaultLineHeight: "normal",
+    defaultTextAlign: "left",
     ready: true,
 
     init: function () {
@@ -42,6 +43,7 @@ Crafty.c("Text", {
             "family": this.defaultFamily,
             "variant": this.defaultVariant
         };
+        this._textAlign = this.defaultTextAlign;
 
         this.bind("Draw", function (e) {
             var font = this._fontString();
@@ -52,6 +54,7 @@ Crafty.c("Text", {
 
                 style.color = this._textColor;
                 style.font = font;
+                style.textAlign = this._textAlign;
                 el.innerHTML = this._text;
             } else if (e.type === "canvas") {
                 var context = e.ctx;
@@ -61,6 +64,7 @@ Crafty.c("Text", {
                 context.textBaseline = "top";
                 context.fillStyle = this._textColor || "rgb(0,0,0)";
                 context.font = font;
+                context.textAlign = this._textAlign;
 
                 context.fillText(this._text, e.pos._x, e.pos._y);
 
@@ -143,6 +147,15 @@ Crafty.c("Text", {
 
         var size = (this._textFont.size || this.defaultSize);
         this.h = 1.1 * this._getFontHeight(size);
+
+        /* Offset the MBR for text alignment*/
+        if (this._textAlign === 'left' || this._textAlign === 'start') {
+            this.offsetBoundary(0, 0, 0, 0);
+        } else if (this._textAlign === 'center') {
+            this.offsetBoundary(this.w/2, 0, -this.w/2, 0);
+        } else if (this._textAlign === 'end' || this._textAlign === 'right') {
+            this.offsetBoundary(this.w, 0, -this.w, 0);
+        }
     },
 
     // Returns the font string to use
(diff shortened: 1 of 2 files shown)


starwed:
Thanks!  I think there's a subtle problem with how it interacts with canvas rendering, though.

For performance reasons, we try to only redraw as much of the canvas as is necessary.  Whenever an entity triggers "Invalidate", Crafty will know that it needs redrawing.  But there's an assumption that only the rectangle defined by an entity's position and dimensions will actually need to be redrawn.

Currently when writing text on the canvas, we [do some math](https://github.com/Chaosed0/Crafty/blob/textalign/src/graphics/text.js#L143) to calculate the area of the screen that the text will take up, and set the dimensions accordingly.   But the calculations assume that the text starts at (x, y) and extends out to the right.

So with this PR as is, I think we'd get some funky rendering issues when using canvas text that wasn't left-aligned.  (Note that the partial redraw strategy is only used when less than 60% of the canvas elements need redrawing, so you might not have noticed this come up in testing.)

I suspect it actually would be quite a bit of work to get this to play nice with canvas -- it'd  probably be ok to just leave a note explaining the issue in the documentation.

starwed:
Actually, you might be able to use the [offsetBoundary](http://craftyjs.com/api/2D.html#-offsetBoundary) method to fix that problem -- I forgot there was an API for dealing with rendering outside the standard quad.  So you could extend the rendering boundary to the left when the alignment requires it.

Chaosed0:
Ah, yeah, I forgot about the dirty rectangles. I'll be honest, I've no idea if this fixes the problem or not - it seems a bit hard to test systematically - but it seems like it should.

As an aside, is there any interest/work on changing the way Crafty's positioning system works? It seems like origin should not only be taken into account for rotation, but also for positioning; that is, the origin of the entity would be the point on (this.x, this.y), and not always the top-left. Obviously it would require some amount of refactoring, but it would make centering entities immensely easier.

starwed:
The logic basically looks ok to me, but I think it should be moved into the `_resizeForCanvas` function, which can then be called when you set the alignment.  (Otherwise if you set alignment and then later change the text to a longer/shorter string, the offset will no longer be correct.)

You can go ahead and squash down to one commit once your done, and then I can merge.  

> As an aside, is there any interest/work on changing the way Crafty's positioning system works? It seems like origin should not only be taken into account for rotation, but also for positioning; that is, the origin of the entity would be the point on (this.x, this.y), and not always the top-left. Obviously it would require some amount of refactoring, but it would make centering entities immensely easier.

My gut reaction is that this would end up making us do a lot more calculations behind the scenes.  I think it might be better to provide a nice interface for centering.   (Although, because we use getters/setters for `x` and `y` and track `_x` and `_y` behind the scenes, it's possible we could use some tricks there to include the origin in what hapens?)

Chaosed0:
Wow, github handled that forced push gracefully :)

I've made the changes and corrected one thing - turns out I did need to reset the MBR if the alignment is changed back to left align.

danielmoniz:
I think there is almost never a good reason to do a force push on Github,
especially when so many people use the library. You can always just git
revert to create a new commit that undoes a previous commit. Please do that
instead in the future; otherwise, many people might have merging issues
when they go to update their Crafty library one day.

starwed:
@danielmoniz You misunderstand the situation -- this is exactly what we [recommend](https://github.com/craftyjs/Crafty/wiki/Rewriting-History), force pushing to your own forked feature-branch to tidy up the pull request.  None of the main repo's history has been altered.

danielmoniz:
Indeed, my mistake! Apologies. That's what I get for using email reply
(instead of checking it out on Github and seeing that it was a pull
request).

starwed:
I think there's one more interaction that needs to be considered -- is it ok to call resizeForCanvas when we're rendering using the DOM?  Seems like you'd end up with the MBR not matching the text in that case, given the differences with how text placement works with the two rendering methods.

(It's kind of ridiculous how complicated the rendering logic is here.  At least we don't try to support WebGL for text!  :) )

Chaosed0:
Wanted to quickly update and say that I haven't abandoned this, just trying to work through the last few days of a project before I get back to this.

Chaosed0:
Many moons later, I've updated the patch so that resizeForCanvas isn't run when Text is DOM. I've barely worked with the DOM component, but I assume that it creates a div and text would be aligned within that div - so, you're right, we wouldn't want to offset the MBR. More to the point, calling resizeForCanvas actually isn't possible when DOM rendering is used, since _drawContext is undefined.

It kind of sucks that the text will render differently depending on whether we're using DOM or Canvas when the text alignment is changed. I'm starting to think this patch was a bad idea to begin with.

starwed:
> It kind of sucks that the text will render differently depending on whether we're using DOM or Canvas when the text alignment is changed. I'm starting to think this patch was a bad idea to begin with.

I think having control over canvas text alignment is pretty useful, so I went ahead and merged this.

PR #875, opened 2015-02-25
Add additional Keyboard docs

### LINKED ISSUE #782: Better documentation for "KeyDown" and related
Hi,
I was wondering if there is an alternative to isDown that only triggers at the moment you press the key.

Something like this:
this.isPressed
and this.isReleased

mucaho:
Fixes #782 .


starwed:
Looks good, merging.

PR #877, opened 2015-03-02
Improve headless support; misc fixes

### LINKED ISSUE #440: CraftyJS & NodeJS

mucaho:
#440 has seen some attention lately: this PR continues to implement parts of #603 in order to allow running crafty on node
- every file declared a `var document = window.document` statement, without actually needing it => removed it where it was not needed
- many tests added the components `DOM`, `Canvas` & `Color` => removed them where not needed
- removed `events.html` which is superseded by `events.js`
- renamed `src/dom-helper` to `src/dom-helper.js`

@starwed
Since we are cleaning up, do we need the now merged [event-loop-fix](https://github.com/craftyjs/Crafty/tree/event-loop-fix) branch, or can it safely be deleted?


starwed:
Some good cleanup fixes.  I'll leave this up for a bit longer in case someone else spots an issue, but it looks like a straightforward merge.

> Since we are cleaning up, do we need the now merged event-loop-fix branch, or can it safely be deleted?

No, that can be safely deleted.  (I could have sworn I already did that!  Oh well...)

mucaho:
Ok, deleted the branch now.

I added another commit which changes header and footer of every source file. This way we can construct multiple instances of Crafty, which is needed to support running on node later on. The size of the minified crafty version did not change.

kevinsimper:
I really like the idea of getting Crafty to run on Node.js, i am also publishing it right now under `craftyjs` on npm.

I can see that this is a minimal to get Crafty working as the it is now, but I think that the changes leaves a code smell on what is a browser javascript which has been retrofitted to node.js

Instead of having one global `Crafty` object and distributing it out to all the files, the correct way would be to return objects from the different file and then gather that into one `Crafty` object.

Global objects is general not something you want as it makes it difficult to see where things come from and what changes what.

I would suggest that it looks something like this where every file returns either a function or a object that is then gathered and put together to a Crafty object.

```
var 2dcomponent = {

};

module.exports = 2dcomponent;
```

main.js

```
var 2dcomponent = require('spatial/2d.js');

var Crafty = {
  components: {
    '2d': 2dcomponent
  }
};
```

mucaho:
Yeah, I discovered the node module by accident and was pleasantly surprised! :+1:
There's even the possibility to use the [server version of crafty on node](https://docs.npmjs.com/files/package.json#main) and the [client version of crafty in browser](https://github.com/substack/node-browserify#browser-field) with the same `require('craftyjs')` statement:
**package.json**

``` json
{
    ...,
    "main": "src/crafty_server.js",
    "browser": "src/crafty_client.js",
    ...
}

```

> Instead of having one global Crafty object and distributing it out to all the files, the correct way would be to return objects from the different file and then gather that into one Crafty object.

Sounds great, but most of the code looks like `Crafty.c("MyComponent", ...)` and `Crafty,extend({...})` :( . How to return this easily?

mucaho:
so @starwed , @kevinsimper : is it ok to merge this as it is, and try to compose Crafty properly from the different modules later?

starwed:
I don't really like that it adds extra, pointless code to the release version.  Browserify doesn't strip out all the module wrapper functions, but they can't really be _used_ as modules by including `crafty.js`.

Is there a way around this?

mucaho:
> Is there a way around this?

There is one other option without cluttering source code (that I know about).
Since `require` caches the objects exported, constructors are needed to create a fresh Crafty instance from a required source file.
However there is a dirty hack which makes `require` re-evaluate the exported object - [require-no-cache](https://www.npmjs.com/package/require-nocache).
Using this hack we can avoid writing constructors and only change the `crafty.js` file. However, this relies on the current implementation of `require` and might not work in future.

mucaho:
ok, I removed the last commit we were discussing about, and will push it into another PR
merging the other changes

PR #879, opened 2015-03-04
Make timer.stop work for implementations of setInterval returning object

mbolis:
The `timer.stop` method is not working in node.js (when testing) and test tasks keep hanging after completion.
This is due to node's `setInterval` implementation returning an object token instead of a number.
Changing the check in `timer.stop` solves this issue.
I believe there should be no side effects to it.

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -1271,7 +1271,7 @@ Crafty.extend({
             stop: function () {
                 Crafty.trigger("CraftyStopTimer");
 
-                if (typeof tick === "number") clearInterval(tick);
+                if (typeof tick !== "function") clearInterval(tick);
 
                 var onFrame = window.cancelAnimationFrame ||
                     window.cancelRequestAnimationFrame ||


starwed:
Seems like a strange decision on node's part, but I guess there must have been some reason?
This is a pretty low risk change, so I'll just merge right away.  

Thanks for the PR!

mbolis:
I was a bit surprised too... not to mention how much I banged my head against it before I understood what was going on!

starwed:
As an aside, I'd guess we never noticed that in our unit tests because we're running qunit through PhantomJS, which sets up up a browser environment which would include the "normal" version of `setInterval`.

mbolis:
Yeh, that pretty looks like the Right Way to test it.
I just started out testing my regular JS objects and functions, then I had a bunch of Crafty components I wanted to test too.
Do you also manage to check code coverage with PhantomJS tests?

PR #881, opened 2015-03-08
Fix particles when viewport scaled

mucaho:
In the current version particles do not consider the current viewport zoom when drawing themselve to the canvas. This PR fixes that and also adds an option to specify an offset from the base entity it is attached to.

original contribution in pr #743 by @johnHacksworth .
additional documentation & misc added by me.

### DIFF
--- a/src/graphics/particles.js
+++ b/src/graphics/particles.js
@@ -25,34 +25,36 @@ Crafty.c("Particles", {
      * @example
      * ~~~
      * var options = {
-     *	maxParticles: 150,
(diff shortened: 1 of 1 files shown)


mucaho:
The issue can be seen in the [requirebin demo here](http://requirebin.com/?gist=2d266a1505772d0ef2d9). Also uploaded to `craftyjs/demos/devDemos` for future reference.

mucaho:
minor fix, so going ahead and merging since no objections

PR #883, opened 2015-03-16
Allow creation of multiple Crafty instances

mucaho:
Continuation of #877 .

By wrapping each source file into a module.exports constructor,
mulitple instances can be constructed.
The initial crafty instance is constructed using the empty
constructor of `core.js`, which is then handed to the other source
modules.
Remove unneeded dependency on animation.js in sprite-animation.js.
Revert HashMap's intersect check to use hardcoded intersect
functionality in order to avoid dependency on Crafty.

Alternatively, to avoid polluting the generated Crafty source file with `module.exports` statements, [another approach](https://github.com/craftyjs/Crafty/pull/877#issuecomment-77738720) could be considered

### DIFF
--- a/src/controls/controls.js
+++ b/src/controls/controls.js
@@ -1,5 +1,6 @@
-var Crafty = require('../core/core.js'),
-    document = window.document;
+module.exports = function(Crafty) {
+var document = window.document;
+
 
 Crafty.extend({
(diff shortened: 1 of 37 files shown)


mucaho:
Closing, superseded by #902, #907, #919

PR #884, opened 2015-03-17
Various improvements to Motion system

mucaho:
Continuation of #876 
- Change units in motion system to use user intuitive pixels per frame and degrees per frame  
  Now when user sets `vx = 1`, the entity will move `1px` each frame, depending on the current target fps `Crafty.timer.FPS()`. Note that this should also work in the cases when user selects different `timer.stepMode` (e.g. `variable`), as the calculation is still done using `dt` and a `dtFactor` derived from the desired target FPS.
- Fixes bug that when setting `gravityConst` before activating `gravity()` leads to inconsistent state
- Fixes bug that leads to incosistent state when entity intersects with multiple ground entities
- Small improvements `Supportable`, `Twoway` and `Multiway`
- Add additional tests
### Future additions
- Allow Supportable to check for SAT collision (see [discussion](https://groups.google.com/forum/#!topic/craftyjs/2_PCJK8B6Ek)). However, what to do about entity snapping? `this.y = hit._y - this._h; // snap entity to ground object`. 
  As you can see, this won't snap correctly with the polygon hitbox being larger/smaller than the actual entity. Should the user be allowed to specify his own snap function?
- I think we should limit the downwards velocity when gravity is active (introduce settable `terminal velocity` that defaults to the velocity caused by gravity after 12 frames) and document for user that he should make his platforms at least `terminal velocity + 1` high - `max(entity._h, platform._h) > terminal velocity`

### DIFF
--- a/src/controls/controls.js
+++ b/src/controls/controls.js
@@ -1017,8 +1017,6 @@ Crafty.c("Keyboard", {
  * @see Motion
  */
 Crafty.c("Multiway", {
-    _speed: 3,
-
     init: function () {
         this.requires("Motion");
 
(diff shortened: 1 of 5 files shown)


mucaho:
The contents of this patch [have been tested in practice](https://groups.google.com/forum/#!topic/craftyjs/L_1KzbynGOM) and are part of implementing the Motion system, so going ahead and merging this.

Issue #882, opened 2015-03-15
Allow to override pixelart setting per-entity


letmaik:
For a game I'd like to use a combination of pixelart and non-pixelart and obviously I'd need to set the resampling method individually for the entities. Currently, `Crafty.pixelart()` is global, so I could imagine that in addition you could override this default value for a given entity with something like:

``` js
Crafty.e('2D, Canvas, mysprite').attr(...).pixelart(true);
```

Combining both resampling methods in a single canvas is easily possible: http://jsfiddle.net/nLh3razm/ And for DOM anyway.

starwed:
That's a pretty reasonable idea.  
- I do think there might some issue in conjunction with zoom, though.  That is, in addition to scaling an individual entity, the entire canvas can also be scaled, and the way that's handled can't be tweaked on a per-entity basis.
- The way it's handled on the WebGL backend is per texture; it might be a little complicated to implement properly, but it's definitely possible

If you want to take an attempt at implementing this I'd be happy to give feedback.

letmaik:
Some thoughts:
- I think setting this per-texture makes way more sense than per-entity, I didn't think that through. So the WebGL way seems right. This means it would apply to sprites and images.
- About zooming, well, even with the current pixelart enabled I never got non-blurry results when zooming in. When I zoom the jsfiddle the pixelart one is crispier but it's still not perfect. So this seems to be a browser-implementation issue or just how canvas works currently. Do you think there is any special additional behavior when this feature request would be implemented?

I can't say currently whether I have the time for it, I may have a look though in case it's too easy.

mucaho:
> When I zoom the jsfiddle the pixelart one is crispier but it's still not perfect. So this seems to be a browser-implementation issue or just how canvas works currently

I think this might be related to the actual implementation of pixelart in Crafty. Crafty currently does `image-rendering: crisp-edges` in the best case scenarion, maybe we should change this to `image-rendering: pixelated` in the best case scenario. See [mdn image smoothing docs](https://developer.mozilla.org/en-US/docs/Web/CSS/image-rendering).

EDIT
The above paragraph is about image smoothing when rendering with `DOM`. It does not explain the weird behavior when rendering with `Canvas` as noted in the quote.

PR #886, opened 2015-03-18
Per-texture pixelart setting

letmaik:
This partly implements #882.

Canvas fully works. WebGL works, but it doesn't listen for pixelart change events from an entity. DOM is not implemented yet, but should be easy.

Main use case is probably via the Crafty.sprite() function, where I added another parameter. Looking at the code is probably the best to understand it. Due to the different ways things are rendered in Canvas, WebGL, DOM I had to make it quite flexible, but I think the design is not bad.

This pull request should not be merged yet but serve for discussion. It probably needs changes, and the DOM implementation.


letmaik:
Closing as outdated.

Issue #885, opened 2015-03-18
Produce dev builds regularly (daily)


letmaik:
Currently public builds only exist for the "release" branch. I think it would be nice to have continuous builds with a fixed dev URL so that interested people can easily test the latest unit-tests-passing dev version without actually being forced to download npm etc. and build it themselves.

The main problem is hosting -- building and uploading is very easy with Travis CI. It shouldn't be any burden to anyone, so I suggest some kind of free hosting. The easiest thing that comes to my mind is to create a separate github repo like crafty-dev-builds which exactly has two files: crafty.js and crafty-min.js. These can always be referenced via rawgit.com. Inside this repo there would be a travis config which just builds the latest crafty commit and if tests pass then it would push this to the crafty-dev-builds repo (by encrypting ssh credentials etc., have done that before as well). This could be invoked daily with a cronjob (I used http://tron-ci.herokuapp.com before). So totally self-autonomous.

Let me know if you have another idea or if you need help with that. I think it's a nice addition.

starwed:
That's a pretty good idea.  It would make sense to use the Crafty-Distro repo for that -- we could have a nightly directory.

letmaik:
Probably better to use a new orphan branch "nightlies" in the Crafty-Distro repo. Also, it probably makes sense to only store crafty.js there, and not crafty-min.js. Otherwise this may encourage people too much to use this for non-testing purposes. Also, checking in the minified one creates probably massive git diffs every time, which is not the case for the crafty.js one.

starwed:
I've had a look at this -- it's not clear to me how to have Travis commit the files when it's triggered by the cron job.

letmaik:
Have a look here: https://github.com/neothemachine/pixelpitch
Every week a cronjob triggers a travis rebuild and the [travis config](https://github.com/neothemachine/pixelpitch/blob/master/.travis.yml) of the master branch gets run, and in that travis pushes to the gh-pages branch and thereby updating a webpage. Exactly the same concept, except that we have a single branch and no website.

kevinsimper:
@neothemachine If you want live build, you can just do `npm install craftyjs/crafty` which installs the latest from github repository.

letmaik:
@kevinsimper Please read my issue text again. This is about not needing npm etc. and just having a simple link available which points to the current dev version. This makes it very easy to let people test new features before release. Personally, I don't use nodejs for anything, so the barrier is quite high just to get the latest version.

[...]

PR #889, opened 2015-04-04
Have travis upload "nightly" builds

starwed:
When a travis build is triggered that is not part of a pull request, it will run a script to upload the build to Crafty-Distro
- It will upload using the OAuth token of the account @craftyjs-machine-user, which has write permissions only to Crafty-Distro and craftyjs.github.com
- The upload script is in the file `travis-nightly.sh`
- I've also tried to configure tron-ci to automatically run the travis build script once a night, but I'm not actually sure that's necessary -- I think it should just run the script after we merge a PR

Thanks to @neothemachine for providing the basis of the script (see #885)!  ~~This particular version hasn't been tested yet.~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,14 @@
 language: node_js
 node_js:
-  - "0.10"
+  - '0.10'
 before_script:
   - npm install -g grunt-cli
+
+after_success: ./travis-nightly.sh
+
 cache:
   directories:
     - node_modules
+env:
+  global:
+    secure: p6SeAGd+Gohr6XeDkb5JPSQBu88UxAjljSMK0BNLJQgr8GkoA3mSJWy4uX0YjNq93VX4UMFxr0D4MlkcWZpX+cdN8K9PpxOgTq8Ml1Ljs355zbX9U/CdY99okiwzHktY4qKXBPsHLiEuCBeJ6pxs/v7m9z88mkjwDSunJG7F1oo=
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,8 +1,14 @@
 language: node_js
 node_js:
-  - "0.10"
+- '0.10'
 before_script:
-  - npm install -g grunt-cli
+- npm install -g grunt-cli
+
+after_success: ./upload-nightly.sh
```
letmaik: travis-nightly.sh?

letmaik:
Tron-ci is not necessary here as you do it after each successful build, which is better anyway. 
Currently this would upload for each branch, but I guess you want "develop" only, so you have to add an additional check in the travis-nightly.sh script.

As the Crafty-Distro repo is empty, it doesn't need to be setup as the new nightlies branch will be a root branch. However, when you create another branch, then make sure that you don't branch that off of nightlies but rather create it as a new root "orphan" branch. By keeping the nightlies branch completely independent you could even regulary remove/recreate the branch to prevent the repo from growing large.

letmaik:
Sorry, I was wrong about the repo setup, just tried it... The branch has to exist already to use `-b nightlies`, otherwise git complains when checking out. So just manually create this branch.

starwed:
Thanks, updated it with those fixes (runs the correct script, checks against the branch being built).  I'll try to setup a real test in my repo to make sure it actually functions as intended now.

starwed:
Ok, I tested it in my own repo and found a few bugs (now fixed).
- syntax error (missing space before `]`) in the conditional
- had to modify the permissions of the shell script (`chmod ugo+x travis-nightly.sh`)
- needed to double quote the commit message (otherwise the variable was interpreted as a string)

starwed:
Well, let's just merge this and see if it works!

PR #890, opened 2015-04-11
Streamline intersect checks

### LINKED ISSUE #874: Migrate rectangle comparisons from `2D` component to `Crafty.rectManager`
`Crafty.rectManager` was created for working on rectangles (in motivation of #578 ), therefore:
- #708 delegates `2D.intersect(otherRectangle)` to `Crafty.rectManager.overlap(thisRectangle, otherRectangle)`
- `2D.within(otherRectangle)` could also delegate to a `Crafty.rectManager.within(thisRectangle, otherRectangle)`
- `2D.contains(otherRectangle)` could also delegate to `Crafty.rectManager.contains(thisRectangle, otherRectangle)`

The above `2D` methods are currently not used by the library (except one case), so performance should not be an issue, as they will be mostly used by user. The documentation of these methods could be updated in order to inform the user that he could use the `rectManager` methods directly instead. The one exception case is located in `Supportable`, so instead of `obj.intersect(this)` `rectManager.overlap(obj, this)` should be used.

Also, add appropriately formatted documentation to `rectManager.merge` or make this method private via underscore(s).

mucaho:
Remove unneeded dependency on animation.js in sprite-animation.js.
Revert HashMap's intersect check to use hardcoded intersection in order to avoid dependency on Crafty.
Revert 2d.intersect to hardcoded intersection (to be streamlined with rest of 2d rect methods, fixes #874).

### DIFF
--- a/src/graphics/sprite-animation.js
+++ b/src/graphics/sprite-animation.js
@@ -1,5 +1,4 @@
-var Crafty = require('../core/core.js'),
-	Animation = require('../core/animation.js');
+var Crafty = require('../core/core.js');
 
 
 /**@
--- a/src/spatial/2d.js
+++ b/src/spatial/2d.js
@@ -488,7 +488,8 @@ Crafty.c("2D", {
             };
         }
 
-        return Crafty.rectManager.overlap(mbr, rect);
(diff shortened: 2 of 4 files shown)


mucaho:
small change, so merging

PR #892, opened 2015-04-18
Split input generating components from input consuming components

mucaho:
This PR continues the work on #578 , #440 and #603 :
Main goal of PR is to split the input generating components (-> inputs.js) from the components that accept this input and modify the entity (-> controls.js).
To facilitate this split, the `Draggable` component had to be reworked. It was split into `MouseDrag` (generates events) and `Draggable` (consumes events and moves entity). Unfortunately found no meaningful way of writing tests for it.
Any input on this greatly appreciated, since a lot of code gets moved around in the last commit.


mucaho:
Did some more tests, _seems_ fine on FF and chrome.
@starwed can we merge this or also test if everything works fine in IE and mobile devices? (how would I go about that? :) )

starwed:
Assuming that the "reorganize code" commit _only_ reorganizes the code, I think this looks reasonable to me.

I did spot a bug where "Twoway" requires "Fourway" instead of "Multiway".  (But that was in the existing code, not introduced in this PR.)

By the way, there are a couple of convenience methods for creating entities I'm not sure you're aware of (they were added after the last release)
- you can indicate required components with a `required` field
- you can autobind events using an `events` field, which should be preferred for any event which will be present for the entire existence of the component
  As an example:

``` javascript
Crafty.c("Hello", {
    required: "Keyboard",
    events: {
        "KeyDown": function() { window.alert("hello"); }
    }
}
```

mucaho:
Thank you for taking a look at it!

I just double checked with an external program, code was only moved in last commit except the category names in the documentation and the module names in the tests were changed appropriately.

These convenience methods sound nice. Do they also serve another purpose (e.g. to show dependencies between components/events)? I will gradually adapt the components to use them, once I do some restructuring of code.

Unfortunately, this PR conflicts with #898 . Waiting for it to be resolved.

mucaho:
Development on linked PR has paused for the moment, so going on ahead and merging this. Will update conflicted PR to new codebase.

PR #893, opened 2015-04-18
Update Isometric Documentation

airza:
Working from the example on the site, it was not obvious to me that diamondIso existed.  This caused me to reimplement its functionality before realizing that it already existed.  The purpose of this documentation is to make it more clear to a casual reader that if they want to create an isometric grid that is aligned at 45 degrees, diamondIso is available.


starwed:
Thanks!  I went ahead and squashed it down to one commit before merging directly.

(Docs on the site won't be updated until the next time they're generated, though.)

PR #896, opened 2015-04-25
Fix supportable ground detection bug

mucaho:
Only one entity is chosen among all intersected ground entities;
however, if this entity does not pass "CheckLanding" requirement, the
player will not land on any ground entity, even if the other entities
satisfy the requirement.

Demonstrated in [this demo](https://github.com/craftyjs/demos/blob/master/devDemos/example_realisticJump.html) when trying to jump in the middle of the first two platforms.


mucaho:
No objections, so merging.

PR #897, opened 2015-04-29
Diamond improvements

airza:
Putting some basic framework in for an isometric grid.  This allows objects to be placed in an iso grid with an X, Y, and Z axis, and it correctly assigns z-height values in the DOM so that the objects render properly.  It also adds rudimentary notions of tiles being more than 1 tile high - a reference to the same object is added and removed to different Z levels.

It also gutted a lot of the old API- I couldn't find what it was supposed to do in a lot of cases and it didn't have documentation.  If we know anyone who uses this or the person who wrote it I can make it compatible.

### DIFF
--- a/src/isometric/diamond-iso.js
+++ b/src/isometric/diamond-iso.js
@@ -5,35 +5,40 @@ Crafty.extend({
     /**@
      * #Crafty.diamondIso
      * @category 2D
-     * Place entities in a 45deg diamond isometric fashion. It is similar to Crafty.isometric but has another grid locations
+     * Place entities in a 45deg diamond isometric fashion. It is similar to isometric but has another grid locations
      * In this mode, the x axis and y axis are aligned to the edges of tiles with x increasing being down and to the
      * right and y being down and to the left.
      */
     diamondIso: {
         _tile: {
             width: 0,
-            height: 0,
-            r: 0
+            height: 0
+        },
+        getTileDimensions: function(){
+            return {w:this._tile.width,h:this._tile.height};
         },
         _map: {
             width: 0,
-            height: 0,
-            x: 0,
-            y: 0
+            height: 0
         },
-
         _origin: {
             x: 0,
             y: 0
+        },
+        _tiles: [],
+        getTile: function(x,y,z){
+            return this._tiles[x][y][z];
         },
         /**@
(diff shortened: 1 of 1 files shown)


starwed:
Thanks!  I've never used the isometric components, so I'm uncertain how/whether anyone is actively using those.  Maybe someone else will have some comments, but if you tried to use the old API and found it broken/undocumented, I'm totally fine with just scrapping it if no one has objections.

One problem, though: This passes the qunit tests, but Travis fails it because of linting errors.  You can run `grunt jshint` locally to quickly get those lints results, which will tell you how to fix them.

It would be good to go ahead and squash the fixes down using `git rebase -i` once you address that.  (Detailed instructions [here](https://github.com/craftyjs/Crafty/wiki/Rewriting-History) if you need them.)

airza:
Makes sense.  My ultimate goal is to roll iso and diamondiso into one component with different axes conventions (though i'm not sure how one might use the regular iso's axes system)  so I might copy the tests over to diamond before I check in.

mucaho:
It would also be nice if we can make `iso` and `diamondIso` independent from `Crafty.viewport` (so we can use it in an environment that doesn't have graphics - see #902 ).

airza:
Sorry, I hopped on a plane right after i made the first pull request! I'll
take a look at this when I get back on monday.

airza:
I think this should work now! I don't know if it notifies someone, but it should.

starwed:
> I think this should work now! I don't know if it notifies someone, but it should.

Thanks!  For reference, commenting notifies me, but just pushing to the branch does not.

Anyone who uses isometric have any comments before I merge this?

airza:
Once it's all integrated into dev I can pull it down and integrate the next set of improvements (I've got entities walking on the grid able to do pathfinding correctly);  sorry it's written against master instead of dev :(

starwed:
> Once it's all integrated into dev I can pull it down and integrate the next set of improvements (I've got entities walking on the grid able to do pathfinding correctly); sorry it's written against master instead of dev :(

Sounds good! but we'll let that be a separate PR.  And this one looks like it's correctly based off the development branch.  (We long ago nuked `master` and set `develop` to be the default for pull requests.  The other branches are `testing` and `release`.)

PR #898, opened 2015-05-01
Add jump keys parameter to TwoWay component

petarov:
Users may now configure a single (Number) or multiple (Array)
additional jump keys in the awesome Twoway component. :metal: 

Not sure, if I edited the docs header correctly.

### DIFF
--- a/src/controls/controls.js
+++ b/src/controls/controls.js
@@ -1342,7 +1342,7 @@ Crafty.c("Twoway", {
      * @sign public this .twoway([Number speed[, Number jump, {Number|Array} jumpKeys]])
      * @param speed - Amount of pixels to move left or right
      * @param jump - Vertical jump speed
-     * @param jumpKeys - Defines one or more additional entity jump keys
+     * @param jumpKeys - Defines one or more entity jump keys
      *
      * Constructor to initialize the speed and power of jump. Component will
      * listen for key events and move the entity appropriately. This includes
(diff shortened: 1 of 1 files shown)


starwed:
I think it would be better if the optional argument _replaced_ the existing jumpkeys, since that can also be a problem.

(As an aside that won't matter if you just replace the existing keys, you should generally avoid `concat` when performance matters, because it creates a new array.  You can [apply the `push` function](http://stackoverflow.com/questions/1374126/how-to-extend-an-existing-javascript-array-with-another-array/1374131#1374131) instead.)

mucaho:
Another alternative would be :
- rename the existing `Twoway` component to `Jumpway`
- add non-optional jump and move keys to `Jumpway` constructor (similar to `multiway(...)`)
- make a new component `Twoway` which calls the `Jumpway` constructor with the default jump and left/right keys (similar to `Fourway`)
- new `Twoway` constructor only accepts left/right speed and optionally jumpSpeed

This doesn't change any code from user perspective but it streamlines it with `Multiway` and `Fourway`. What do you guys think?

petarov:
@mucaho Sounds good to me, although I'm not really familiar with the underlying components. My goal here was to just find a way around an issue that I had where `Z` triggers both a jump and shoot on a tiny game I work on.

mucaho:
resolved with #900

PR #900, opened 2015-05-11
Add jump keys parameter to TwoWay functionality

mucaho:
Alternative to #898 . What do you think @petarov ?
In order to specify custom keys you can do something like this:

``` js
Crafty.e("..., Multiway, Jumpway")
     ...
    .multiway(3, { RIGHT_ARROW: 0, LEFT_ARROW: 180 })
    .jumpway(6, [ 'UP_ARROW' ]);
```

### DIFF
--- a/src/controls/controls.js
+++ b/src/controls/controls.js
@@ -3,7 +3,9 @@ var Crafty = require('../core/core.js');
 /**@
  * #Draggable
  * @category Controls
- * Enable drag and drop of the entity. Listens to events from `MouseDrag` and moves entitiy accordingly.
+ * Enable drag and drop of the entity. Listens to events from `MouseDrag` and moves entity accordingly.
+ *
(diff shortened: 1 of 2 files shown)


petarov:
Looks awesome! :beer: 

I'm not a big fan of omitting params (`jumpSpeed`), but that's shouldn't be an issue at all.

mucaho:
> I'm not a big fan of omitting params (`jumpSpeed`)

Yeah optional params are used throughout the library (they allow multiple functionalities within one method so that API doesn't explode). In this case, users can call the constructor multiple times to change the jump keys without the need for specifying/changing the jump speed.

mucaho:
No objections thus far, it's not a big change, follows the idea of 208c3789f171192a836408bb66625c164abc1561 of splitting components, merging.
More tests for control components to be added in future.

PR #901, opened 2015-05-12
Update Crafty_diamondIso_polygon implementation

mucaho:
Crafty_diamondIso_polygon still uses old Crafty_polygon signature.
This commit fixes that and brings it in line with #835  .


starwed:
Looks good, merging.  (Though, we probably should add tests to cover isometric!  It's a bad sign that this wasn't caught by our tests.)

PR #902, opened 2015-05-12
Build & test crafty for nodejs

mucaho:
Behold! The time of running crafty on node is nigh! :) #440, #603
### Implemented thus far:
- `grunt-node-qunit` can be used to run the existing QUnit tests on node. caveat is that there is a name conflict between node's module and QUnit's module (hence c1c3301)
- encapsulated some test files in anonymous function (like rest of the tests) - 15c33f6
- some moving of tests is necessary, so that we can run whole test files on node (016f27e, 6a97e55)
- mock `Keyboard.isDown` in Multiway tests (703f039)
- build crafty for server (8e5d33e) - this is easily achieved with a bit of trickery using `require-nocache`. Users can go ahead and add `craftyjs` to their dependencies in their `package.json` of their node project. After that they can construct a new instance of crafty by calling `var Crafty = require('craftyjs')()`.
- perform tests on crafty for server instance (cc893ad). essentially the tests are run for the functionalities that are included in the headless build
### Outstanding issues:
- ~~`grunt-node-qunit` doesn't like async tests. haven't been able to solve this problem yet (may need to convert all async tests into sequential tests in worst case)~~ **forgot to `Crafty.init()` - all ok regarding async tests**
- still need to adapt the ~~`Crafty.support`, `Crafty.storage`~~(**adapted**) `isometric` functionalities so that they can run on node
- still need to find cleaner way of providing dummy/mocked methods ([crafty_headless.js](https://github.com/craftyjs/Crafty/compare/develop...mucaho:build_headless_new?expand=1#diff-10816ad19b11adbf7c4cd4ce895202e8); without those crafty doesn't work
### Notes

Would appreciate any pointers!
Preliminary plan is to include the commits gradually - first 6 commits are pretty straightforward and "finished".

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ build/api-gen.js
 .project
 .idea
 node_modules/
+localStorage/
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -100,9 +100,7 @@ module.exports = function (grunt) {
         },
 
         qunit: {
-            all: [
-                'tests/index.html'
-            ]
+            all: ['tests/index.html'] // TODO add headless tests here
         },
 
         jsvalidate: {
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -100,7 +100,47 @@ module.exports = function (grunt) {
         },
 
         qunit: {
-            all: ['tests/index.html'] // TODO add headless tests here
+            all: ['tests/index.html']
+        },
+
+        'node-qunit': {
+            all: {
+                deps: 'tests/lib/helperFunctions.js',
+                code: 'tests/index_headless.js',
+                tests: [
+                    'tests/common.js',
+                    'tests/core.js',
+                    'tests/2d.js',
+                    'tests/logging.js',
+                    'tests/controls.js',
+                    'tests/events.js',
+                    //TODO add these once isometric adapted:
+                    //'tests/isometric.js',
+                    'tests/math.js',
+                    'tests/model.js',
+                    'tests/storage.js',
+                    'tests/systems.js',
+                    'tests/time.js',
+                    'tests/tween.js',
+                    'tests/issue746/mbr.js',
+                    'tests/issue746/pos.js',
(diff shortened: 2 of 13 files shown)


mucaho:
@starwed, if you got time/motivation: What do you think is the best way to proceed with providing different implementations for `Crafty.storage` & `Crafty.support`? Try to squeeze in alternative implementation into the existing files (by checking if `window !== undefined`, etc.) or add code to separate files (e.g. `extensions_headless.js` & `storage_headless.js`). However, what to do about documentation in the latter case - duplicate it?

starwed:
I've just finished teaching for the semester, so I should have some more time to devote to Crafty for a bit!

Perhaps there are just some node modules that duplicate browser functionality?  That often seems to be the case.
- We used to have a more flexible, async version of storage (I believe you could configure it to save data via html requests), but it was kind of a mess and the localstorage specs it was using were in heavy flux, so it was constantly broken.  The current version is simpler but less flexible, and I think modern browsers now have much better support for our original implementation.  I'd guess that there might be a 3rd party library we could switch to that would be available on both node and in browsers.
- What are the specific problems with `Crafty.support`?  Shouldn't it just correctly indicate that node doesn't support some functionality?

mucaho:
Great, thanks for your input!
There is a replacement for `localstorage` - [node-localstorage](https://www.npmjs.com/package/node-localstorage). We can use this for now and switch to something more advanced later.
The question is whether to hack this into the existing files (so we mix browser logic with node logic) or add node-specific code to new files (but then we have to be careful to keep both versions of the file consistent)?
EDIT: aha, problem in `Crafty.support` is it accesses `window`, `document`, `navigator`, so a few additional conditions have to be added, so I guess you are saying it's fine if we squeeze those conditionals in there?

mucaho:
Went ahead and manually merged the first few commits, since they were pretty straight-forward and streamlined tests a bit.

mucaho:
Rebased against new codebase, going ahead and merging this in order to move forward #914 .

PR #906, opened 2015-05-17
New scripts for generating json doc data from source

starwed:
This changes our documentation generation to produce a single json file, rather than a complete set of api files.  There's now a build script in the site repo that'll produce them from the json file.

There are two scripts:
-  `parseSourceDocs.coffee` takes the source files and parses each line, returning a single array of nodes
-  `parseNodes.coffee` takes the array of nodes and returns a structured json document


starwed:
Low risk (since it doesn't touch Crafty source code) so merging.

kevinsimper:
It uses tabs instead of the old spaces? :)

Nice that it generates Json, i tried the same long time ago, but the tags did not quite match up with the source at that time. Has this changed?

starwed:
I think the previous problem was using an existing jsdoc tool -- this was custom built for our style.  (It's also hopefully easy to modify if we want to add new tags like warnings or stability tags.)

PR #907, opened 2015-05-20
Exports objects out of files and assemble in crafty.js

kevinsimper:
This helps to avoid globals and makes it more modular as the components can be loaded individual.

It is a step towards making it work with node.js.

I have not done it with the rest of the modules as I want to get feedback first :)

### DIFF
--- a/src/core/animation.js
+++ b/src/core/animation.js
@@ -1,6 +1,3 @@
-var Crafty = require('../core/core.js');
-
-
 /**@
  * #Crafty.easing
  * @category Animation
@@ -25,7 +22,7 @@ var Crafty = require('../core/core.js');
  * ~~~
  * @see Tween, SpriteAnimation
  */
-Crafty.easing = function(duration, easingFn) {
+var easing = function(duration, easingFn) {
 	this.timePerFrame = 1000 / Crafty.timer.FPS();
 	this.duration = duration;   //default duration given in ms
 	if (typeof easingFn === "function"){
@@ -39,7 +36,7 @@ Crafty.easing = function(duration, easingFn) {
 };
 
 
-Crafty.easing.prototype = {
+easing.prototype = {
 	duration: 0,
 	clock:0,
 	steps: null,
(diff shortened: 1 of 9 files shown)


mucaho:
That's exactly what you [suggested before](https://github.com/craftyjs/Crafty/pull/877#issuecomment-76948080). I like the idea very much, it's clean approach, however it's a lot of work.
Crafty already works on nodejs with minimal changes (see #902 ), even the tests run on nodejs successfully.
Lets also hear @starwed 's opinion on this.

kevinsimper:
It took me 20 minutes, so the rest is maybe 30 minutes :) So that is okay.

But you see that there is quite some different approaches to extending the `Crafty` object. Some extend the whole object, other attach to a property and then we have `Components`.

kevinsimper:
Nobody seems to comment against it, so merging and will make a new pull request with the rest of the work done.

mucaho:
Right now you plan to export one component / extension-object per file, right? See [starwed's comment](https://github.com/craftyjs/Crafty/issues/578#issuecomment-27119884) from before 

> > I'd like to see only one component per file.
> 
> Hmm, I'm not so certain about this as a hard rule. Some components are very short, and depend heavily on other components. For instance, should "FourWay" be in a separate file from "MultiWay"? All the former component does is call multiway. Having to constantly hop between files hurts readability just as much as having to navigate a huge monolithic file.
> 
> There are scores of components that should be in their own files, though, so I agree in general. And I think everything in its own file, while not ideal, is certainly an improvement over the current situation.

I tend to agree that all `*way` components are easier to understand if they are in one file. So if we go down this route (to bundle multiple components into a file) then exporting objects gets messy, as you have to export an array of components and/or extension objects - it would be arguably worse than the way it is now.

kevinsimper:
@mucaho I understand what you're saying, and I think it can be solved by making a function in Crafty that takes an array of components and loops over them. That would also make it easier to import 3rd party components.

The problem that I am solving is not putting each component in each file, but removing the need for Crafty to be global variable (globals are bad).

But I agree that some components should stay together because they are so close together in functionality.

PR #908, opened 2015-05-20
Remove unnecessary properties from package.json

kevinsimper:
There is a couple of properties in package.json that is not necessary and are leftovers.

The licenses is also wrong according to the docs, so I changed it.
https://docs.npmjs.com/files/package.json#license


kevinsimper:
Merging as it is harmless

PR #911, opened 2015-05-21
 Warn when using old syntax in Crafty.load

### LINKED ISSUE #910: Crafty.load() hangs when supplied with an array of strings specifying assets
When one supplies an array of strings specifying resources to Crafty.load(), a game simply hangs without outputing any error or in other way indicating asset loading failure. 

This is a misleading behavior, suggesting that assets are loading, while the opposite is true.

starwed:
Since we changed the method signature of Crafty.load, we should warn when people try to use the old syntax.

This PR also adds a very simple wrapper (`Crafty.log`), so that the game doesn't simply crash in IE9 when logging is disabled.

This fixes #910.

### DIFF
--- a/src/core/loader.js
+++ b/src/core/loader.js
@@ -226,6 +226,10 @@ Crafty.extend({
      */
     load: function (data, oncomplete, onprogress, onerror) {
       
+        if (Array.isArray(data)) {
+            Crafty.log("Calling Crafty.load with an array of assets no longer works; see the docs for more details.");
+        }
+
         data = (typeof data === "string" ? JSON.parse(data) : data);
       
         var j = 0,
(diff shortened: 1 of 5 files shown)


starwed:
(Updated PR: I fixed a problem with `Crafty.log` and added tests and documentation for it.)

kevinsimper:
I think we should use something like `debug` because it is already widely used and it is very easy to use in node.js as well.
https://github.com/visionmedia/debug

And you can multiple debuggers and enable only a subset of them by doing `.debug('components:*')`

:)

starwed:
I'm open to that, but it should probably be its own PR.  I think that even if we merged this super-simple implementation, it would be pretty straightforward to switch over to something else?

kevinsimper:
Yes, lets do that! :+1:

PR #912, opened 2015-05-21
Add a server for viewing the documentation

starwed:
This adds a new grunt task `grunt api-server` that will serve the api documentation directly from `api.json`.

Use `grunt view-api` to build the docs and then view them in your default browser.

This copies over some scripts from the craftyjs.com repo, but perhaps there's a better way to share them.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,5 +1,6 @@
 require("coffee-script");
 var fs = require('fs');
+var open = require("open");
 
 module.exports = function (grunt) {
     var pkg = grunt.file.readJSON('package.json');
@@ -24,6 +25,16 @@ module.exports = function (grunt) {
         grunt.file.write(outputFile, apiJSON);
         grunt.log.writeln("Wrote api data to " + outputFile);
     };
+
+    var apiServer = require("./build/api-gen/dynamic-server.js");
+    function runApiServer() {
+      var done = this.async();
+      apiServer(grunt, "./build/api.json");
+      setTimeout(function(){
+        open("http://localhost:8080");
+      }, 100);
+    }
+
 
     // Project configuration.
     grunt.initConfig({
@@ -104,6 +115,12 @@ module.exports = function (grunt) {
                     keepalive: true
                 }
             }
+        },
+
+        open: {
+            api : {
+              path: 'http://localhost:8080/',
+            },
         }
 
     });
@@ -144,5 +161,8 @@ module.exports = function (grunt) {
 
     // Run only tests
     grunt.registerTask('validate', ['qunit']);
+
+    grunt.registerTask('api-server', "View dynamically generated docs", runApiServer);
+    grunt.registerTask('view-api', ['api', 'api-server'] );
 
 };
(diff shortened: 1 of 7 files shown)


Gruntfile.js
```diff
@@ -25,6 +26,16 @@ module.exports = function (grunt) {
         grunt.log.writeln("Wrote api data to " + outputFile);
     };
 
+    var apiServer = require("./build/api-gen/dynamic-server.js");
+    function runApiServer() {
+      var done = this.async();
+      apiServer(grunt, "./build/api.json");
+      setTimeout(function(){
+        open("http://localhost:8080");
```
kevinsimper, line 34: Why do you have the same `open` here, when you also have a grunt task that does the same?

kevinsimper:
Really nice with React components :+1: 

Do we need to commit the `api.json`?
Could we split up the react components into their individual files? It is easier to get an overview this way and later make improvements to the individual components :)

Gruntfile.js
starwed, line 34: Ah, I forgot to remove the grunt task.  It didn't quite work the way I wanted, and when I looked at the source I realized it was just a wrapper for open!

starwed:
> Could we split up the react components into their individual files? It is easier to get an overview this way and later make improvements to the individual components :)

I thought about that, but it would be a lot of files!  In any case, this is just a direct copy of what's in the site repo, so the change would need to be made there.

> Do we need to commit the api.json?

Nope, I accidentally copied over the version from the site repo.  (The one generated by the build script is already in .gitignore.)

starwed:
(I fixed those issues and pushed the amended commit.)

mucaho:
> This copies over some scripts from the craftyjs.com repo, but perhaps there's a better way to share them.

There are [git submodules](http://git-scm.com/book/en/v2/Git-Tools-Submodules) and [alternatives](http://blogs.atlassian.com/2013/05/alternatives-to-git-submodule-git-subtree/). Github supports `submodules` (they are shown as _regular_ folder in github repo), don't know about alternatives. Haven't investigated in detail, but might be what you are looking for.

build/api-gen/server-side.jsx
```diff
@@ -0,0 +1,46 @@
...
+var ToC = docComp.ToC;
+var DocPage = docComp.DocPage;
+var EventPage = docComp.EventPage;
+
+
+// We create a StaticPage component for generating the page
+// Unlike in thre dynamic case, we assume we can just pass the necessary state directly
+var StaticPage = React.createClass({
+	render: function() {
+		var pages = this.props.index.pages;
```
kevinsimper, line 13: Using tabs here? :)

Should we point grunt at validating files in `build` as well? Maybe editorconfig? http://editorconfig.org/

kevinsimper:
I think that those React components will at some point be put into different files, because else will it be unmanageable later to require and reuse those components in different places. It also makes it easy to see what depencies what components use like `marked`, that maybe needs to updated later.

starwed:
git submodules are actually a bit of a pain; you have to manually pull in and commit changes, and we'd have to juggle a _third_ repo.  I actually think just using `cp` will be easier for sharing code between two places, since hopefully the code won't change very often.

starwed:
Since this certainly doesn't break anything, I'm going to go ahead and merge.

PR #913, opened 2015-05-25
This code uses the first result for zero filling an integer on Google…

airza:
… and then casts it immediately back to an integer.  Let's just use multiplication.  Also, using Crafty.tween fails for incomprehensible reasons so let's round float Z values up in the meantime while we figure out why.  (Tween correctly avoids calculating by stepping through values and instead interpolates based on the start and end values.  I can't think of any other legitimate usage for fractional z values up.)

A final problem was:  The code was triggering the reorder event before the new z value was assigned; this is bad because reordering doesn't happen until afterwards.  So, any reordering code still sees the old Z value and essentially does nothing.

In the face of all sanity it looks like setting the _globalZvalue here is either getting set again somewhere else, or there's some reason for the really weird reason that it's setup like it is.

### DIFF
--- a/src/spatial/2d.js
+++ b/src/spatial/2d.js
@@ -16,15 +16,6 @@ var M = Math,
     PI = M.PI,
     DEG_TO_RAD = PI / 180;
 
-Crafty.extend({
-    zeroFill: function (number, width) {
-        width -= number.toString().length;
-        if (width > 0)
-            return new Array(width + (/\./.test(number) ? 2 : 1)).join('0') + number;
-        return number.toString();
-    }
-});
-
 /**@
  * #2D
  * @category 2D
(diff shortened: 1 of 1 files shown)


starwed:
Thanks for the PR!

And yeah,  I'm not really sure what was going on there; looks like it's been that way for several years at least. It wasn't causing bugs (other than with tweening), but it doesn't seem to make much sense.
- The `globalZ` value is currently set at the bottom of the `_attr` function, so it was only the event trigger that was wonky.
-  The "reorder" trigger doesn't seem to be used anywhere in Crafty's own code, which is why it wasn't causing any bugs.  I suppose it's there for people to hook into, but it isn't actually documented anywhere?  (It should also be renamed to "Reorder" for consistency.)

This patch looks fine to me, but looking at the `_attr` function I'm struck once more that it's fairly pointless -- very little of the code of the different branches is really shared, and since it's something that gets called constantly, the overhead of a function call and the conditional statements might actually matter.

PR #915, opened 2015-05-28
Fix some formatting issues in our doc blocks

starwed:
We had inconsistent conventions regarding `@param` and `@returns` tags
- In some places, the description of the param/return value was continued across multiple lines
- In many other places, the description of the method as a whole was place directly after such tags

While the latter outnumbered the former, I think its useful to be able to have longer param/return descriptions split across multiple lines.  So to allow this while not breaking the docs, this fixes all incidences of the second style.  (This is probably also more readable when looking directly at the source.)

### DIFF
--- a/src/controls/controls.js
+++ b/src/controls/controls.js
@@ -68,6 +68,7 @@ Crafty.c("Draggable", {
      *
      * @sign public this .dragDirection(degree)
      * @param degree - A number, the degree (clockwise) of the move direction with respect to the x axis.
+     *
      * Specify the dragging direction.
      *
      * @example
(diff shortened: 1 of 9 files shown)


kevinsimper:
Is this something we can have a checker for?

starwed:
There were 1 or 2 instances I missed involving the `@trigger` tag, but they're fixed in my big doc PR.

>  Is this something we can have a checker for?

Not really, because in some places it's done on purpose.  It's nice to break long text up over multiple lines.

I did use a script to find all existing places where it was happening, so it wasn't too hard to audit them. (I just modded the build script to warn when those tags were consuming additional lines of text.)

PR #916, opened 2015-05-28
Now tweening doesn't screw up if the components are changed.

airza:
BUGFIX: If the tweening component is added/removed multiple times from a component, it gets two copies of this, which makes tweens go too quickly.

### DIFF
--- a/src/core/tween.js
+++ b/src/core/tween.js
@@ -11,7 +11,7 @@ module.exports = {
     this.tweenGroup = {};
     this.tweenStart = {};
     this.tweens = [];
-    this.bind("EnterFrame", this._tweenTick);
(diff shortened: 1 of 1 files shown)


airza:
I experimented with dynamically binding and unbinding the frame handler when there were no tweens to do, but surprisingly there wasn't a performance increase (it seemed like a function call per tweened object every frame was a waste)  so I left it out and just added this.

starwed:
Looks good, and a simple fix.

Another option would have been to unbind when the component is removed.  We quasi-recently added an `events` field which automatically takes care of that stuff.  (At some point I should go through and update our built-in components to  use new features!)

PR #918, opened 2015-06-02
Fix numerous small documentation issues

starwed:
This PR does two main things
- Update the doc generator and viewer to handle new tags (`@warning`, `@note`) and cross-page anchor references
- Fixes tons of small issues I found while auditing our docs as part of the 0.7.0 release.  This includes typos, malformed syntax, missing parameters, incorrect return types, convoluted phrasing, and pretty much anything else that jumped out at me!

I deliberately skipped auditing 
- the new Motion & co., since I think those are still in a bit of flux.
- The math.matrix methods, since they seem to be using a jsdoc style we haven't ever supported.  That'll require a separate PR to straighten up, one way or the other

There's a parallel PR that adds these new capabilities to the craftyjs repo.

### DIFF
--- a/build/api-gen/doc-components.jsx
+++ b/build/api-gen/doc-components.jsx
@@ -75,6 +75,21 @@ var DocLink = React.createClass({
     } else if (target[0] === ".") {
       linkText = target;
       href = "#" + cleanName(target);
+    } else if (target.indexOf("#") > 0){
+      var i = target.indexOf("#");
+      var page = target.substring(0, i);
+      var anchor = target.substring(i+1);
+
+      // Handle link text based on context
+      if (anchor.indexOf(page) >= 0) {
+        linkText = anchor;
+      } else if (anchor[0] === ".") {
+        linkText = page + anchor;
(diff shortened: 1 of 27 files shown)


kevinsimper:
Looks good!

We should add documentation generation to travis! Wouldn't that be possible to test as well?

starwed:
> We should add documentation generation to travis! Wouldn't that be possible to test as well?

There are some simple things I could add to the build scripts that would be good to check.  Not sure if there are any good tests to apply to the final json file in isolation.

kevinsimper:
Would those things be including if there is any parsing error or that the api can not be generated correct?

starwed:
> Would those things be including if there is any parsing error or that the api can not be generated correct?

Kind of... a couple of obvious things to check for are 
- unknown tags, like if someone added a `@stability` tag to the docs without updating the generator
- Pages with no description at the top, or blocks which don't specify any category or page they belong to

We could also have a small sample input, and check that the emitted json has the correct properties.

PR #919, opened 2015-06-03
Commonjs

kevinsimper:
Follow up on my previous pull request #907

I have made this in several commits, so that it is easier to get an overview.

There is still a lot of places where the global variable is directly. There is still more work to be done.

All those where the extend is used, and the file has private variables, it would be two different properties if you try to require the files and use the Crafty namespace.

The solution to that would be to make a main object (like a namespace) and then only export that. That would mean you would have to access the component through that namespace.

### DIFF
--- a/src/controls/controls.js
+++ b/src/controls/controls.js
@@ -7,7 +7,7 @@ var Crafty = require('../core/core.js');
  *
  * @see MouseDrag
  */
-Crafty.c("Draggable", {
+exports.dragableComponent = {
     _origX: null,
     _origY: null,
     _oldX: null,
@@ -123,7 +123,7 @@ Crafty.c("Draggable", {
             this.y = this._oldY + (e.realY - this._origY);
         }
     }
-});
+};
 
 /**@
  * #Multiway
@@ -135,7 +135,7 @@ Crafty.c("Draggable", {
  *
  * @see Motion, Keyboard
  */
-Crafty.c("Multiway", {
+exports.multiWayComponent = {
     _speed: null,
     
     init: function () {
@@ -345,7 +345,7 @@ Crafty.c("Multiway", {
 
         return this;
     }
-});
+};
 
 
 /**@
@@ -359,7 +359,7 @@ Crafty.c("Multiway", {
  *
  * @see Supportable, Motion, Keyboard, Gravity
  */
-Crafty.c("Jumpway", {
+exports.jumpwayComponent = {
     _jumpSpeed: 6,
 
     /**@
@@ -492,7 +492,7 @@ Crafty.c("Jumpway", {
         this._jumpSpeed = jumpSpeed;
         return this;
     }
-});
+};
 
 /**@
  * #Fourway
@@ -505,7 +505,7 @@ Crafty.c("Jumpway", {
  *
  * @see Multiway
  */
-Crafty.c("Fourway", {
+exports.fourwayComponent = {
 
     init: function () {
         this.requires("Multiway");
@@ -540,7 +540,7 @@ Crafty.c("Fourway", {
 
         return this;
     }
-});
+};
 
 /**@
  * #Twoway
@@ -553,7 +553,7 @@ Crafty.c("Fourway", {
  *
  * @see Multiway, Jumpway
  */
-Crafty.c("Twoway", {
+exports.twowayComponent = {
 
     init: function () {
         this.requires("Multiway, Jumpway");
@@ -594,4 +594,4 @@ Crafty.c("Twoway", {
 
         return this;
     }
-});
+};
--- a/src/controls/device.js
+++ b/src/controls/device.js
@@ -1,7 +1,6 @@
 var Crafty = require('../core/core.js');
 
-
-Crafty.extend({
+module.exports = {
     /**@
      * #Crafty.device
      * @category Misc
@@ -156,4 +155,4 @@ Crafty.extend({
             }
         }
     }
-});
+};
--- a/src/controls/inputs.js
+++ b/src/controls/inputs.js
@@ -1,7 +1,7 @@
 var Crafty = require('../core/core.js'),
     document = window.document;
 
-Crafty.extend({
+exports.inputsObject = {
     over: null, //object mouseover, waiting for out
     mouseObjs: 0,
     mousePos: {},
@@ -560,7 +560,7 @@ Crafty.extend({
             return false;
         }
     }
-});
+};
 
 //initialize the input events onload
 Crafty.bind("Load", function () {
@@ -652,7 +652,7 @@ Crafty.bind("CraftyStop", function () {
  * @see Crafty.touchDispatch
  * @see Crafty.mouseButtons
  */
-Crafty.c("Mouse", {
+exports.mouseComponent = {
     init: function () {
         Crafty.mouseObjs++;
         this.requires("AreaMap")
@@ -660,7 +660,7 @@ Crafty.c("Mouse", {
                 Crafty.mouseObjs--;
             });
     }
-});
+};
 
 /**@
  * #Touch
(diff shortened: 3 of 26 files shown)


starwed:
I still don't quite get the motivation here -- could you explain the specific benefits?

kevinsimper:
@starwed Yes, I did not quite explain that.

One of the benefits is that we are moving away of using globals. Going to borrow some from this article:
http://singlepageappbook.com/maintainability1.html

_"Dependencies are easy to locate, without being modifiable or accessible in the global scope. Ever been confused about where a particular function comes from, or what the dependencies of a particular piece of code are? Not anymore: dependencies have to be explicitly declared, and locating a piece of code just means looking at the file path in the require statement. There are no implied global variables."_

Other benefit is that this is a step towards using Crafty on Node.js, but there globals is not a feasible way, because that means you would only allow the server to run one game, which would be pretty limited. 

I did also just now reorganise crafty.js, so it looks a little prettier than before.

starwed:
There's just something that bugs me about how this treats components.  The component object is defined in a module, then named in the central file.  But there are plenty of hard-coded references to components in the codebase; if you pulled in "Supportable" but didn't call it that, components like "Gravity" which require it would break.

Perhaps we should have an alternate syntax for `Crafty.c` that would accept objects, allowing this:

``` javascript
// 2D.js
module.exports.2dComponents = {
     "2D":{ ... component object .. },
     "Motion": { ... etc ... }
});

// crafty.js
Crafty.c(  require("spatial/2D.js").2dComponents );
```

That would make the component definitions easier to read, and a 3rd party could still reuse and rename a specific one.

## But... I think there's a problem

I don't think your current approach would actually allow you to maintain multiple instances of Crafty.  When you require a node module multiple times, it will return _the same object_ in some circumstances.  That means that different game instances would be sharing the same fundamental objects, which would almost definitely mess things up.  (I haven't specifically checked this, but everything I've read says [node works this way](http://stackoverflow.com/questions/8887318/understanding-node-js-modules-multiple-requires-return-the-same-object).  Let me know if I'm totally off-base here!  _e:_ Also, a [blog post](http://justjs.com/posts/singletons-in-node-js-modules-cannot-be-trusted-or-why-you-can-t-just-do-var-foo-require-baz-init) on the topic that explains it in a bit more detail.)

The first solution that occurs to me is to return object generating functions, rather than objects:

``` javascript
// 2D.js
module.exports.2dComponents = function(){
    return {
     "2D":{ ... component object .. },
     "Motion": { ... etc ... }
    });
}

// crafty.js
Crafty.c(  require("spatial/2D.js").2dComponents() );
```

But if we want to do that, there's also another approach we could use, which is closer to our current structure:

``` javascript
// 2D.js
module.exports.define2DComponents = function(Crafty){
    Crafty.c("2D", { ...component obj...});
    Crafty.c("Motion", {...etc...});
};

// crafty.js
require("spatial/2D.js").define2DComponents();
```

This last approach would have the benefit of allowing components/objects/etc to directly refer to the Crafty objects without it being a global.  It has the downside that it's less flexible -- you either define all or none of the components in a module, and you can't rename them if that's what you want.

kevinsimper:
That was a long response, I will try to explain it my best

> But there are plenty of hard-coded references to components in the codebase; if you pulled in "Supportable" but didn't call it that, components like "Gravity" which require it would break.

Yes, and this is not a complete solution, but a step towards it, there is still more work to be done! Rome was not built in one day :)

> Perhaps we should have an alternate syntax for Crafty.c that would accept objects, allowing this:
> 
> // 2D.js
> module.exports.2dComponents = {
>      "2D":{ ... component object .. },
>      "Motion": { ... etc ... }
> });
> 
> // crafty.js
> Crafty.c(  require("spatial/2D.js").2dComponents );
> That would make the component definitions easier to read, and a 3rd party could still reuse and rename a specific one.

Yes, I talked about something similar here #907: 

> I understand what you're saying, and I think it can be solved by making a function in Crafty that takes an array of components and loops over them. That would also make it easier to import 3rd party components.

But i don't see how overwriting `module.exports` and using `exports` is different?

> I don't think your current approach would actually allow you to maintain multiple instances of Crafty. When you require a node module multiple times, it will return the same object in some circumstances. That means that different game instances would be sharing the same fundamental objects, which would almost definitely mess things up. (I haven't specifically checked this, but everything I've read says node works this way. Let me know if I'm totally off-base here! e: Also, a blog post on the topic that explains it in a bit more detail.)

Yes, and this pull request did not solve all the problems, this was only a step towards getting rid of Globals which would prevent it. To handle multiple instances of Crafty, you would have to use the `new` keyword, which is how `express.js` and many others do.

> module.exports.define2DComponents = function(Crafty){
>     Crafty.c("2D", { ...component obj...});
>     Crafty.c("Motion", {...etc...});
> };
> 
> // crafty.js
> require("spatial/2D.js").define2DComponents();
> This last approach would have the benefit of allowing components/objects/etc to directly refer to the Crafty objects without it being a global. It has the downside that it's less flexible -- you either define all or none of the components in a module, and you can't rename them if that's what you want.

How does `Crafty` get here `function(Crafty){`, when you don't pass a global variable here: `define2DComponents();`?

I think it is a anti pattern and something that resemble codesmell when to pull a global reference around in different modules, you never know where the global object is modified, which makes it really hard to debug. That is one of the main cases with the popular React.js, where you can only pass data down to children, but never upwards.

You can see how `express.js` does, it assembles all in this file and then make it possible to create multiple instances of express.js, https://github.com/strongloop/express/blob/master/lib/express.js.

starwed:
> Rome was not built in one day :)

Certainly true!  :)

> But i don't see how overwriting module.exports and using exports is different?

Sorry, I think I was unclear there -- I was suggesting that the field names of each object would be interpreted automatically as the name for the component.

> To handle multiple instances of Crafty, you would have to use the new keyword, which is how express.js and many others do.

Right, that's roughly equivalent to what I'm talking about.  It means you can't just return objects from the modules, you have to return functions that generate and return the objects.  They use `new` because they're using the functions as constructors, but we wouldn't need to do that with components.  But I think it complicates the idea of exporting multiple components from one module.

> How does Crafty get here function(Crafty){, when you don't pass a global variable here: define2DComponents();?

That was a dumb typo on my part, it should be passing an object in.  I believe this is how jQuery works when you run multiple instances in the same context?  (Louis based Crafty's original structure pretty heavily on jQuery -- relevant here is that both use the `.extend` idiom.)

starwed:
Oh, and my main point about passing in the Crafty object was that it would allow us to move from Crafty being a global object more quickly.

I think in the long term we can actually remove most hardcoded references to it in the code base through the use of the new system idiom.  Once we release the next version I want to go through and convert most of our subsystems (such as the viewport or the rendering layers) to use it.

mucaho:
> I don't think your current approach would actually allow you to maintain multiple instances of Crafty. When you require a node module multiple times, it will return the same object in some circumstances. That means that different game instances would be sharing the same fundamental objects, which would almost definitely mess things up. (I haven't specifically checked this, but everything I've read says node works this way. Let me know if I'm totally off-base here! e: Also, a blog post on the topic that explains it in a bit more detail.)

I don't know if I explained in detail but there are node modules that delete the `require` cache and thus return new objects every time you require something - [see it here in action](https://github.com/craftyjs/Crafty/pull/902/files#diff-10816ad19b11adbf7c4cd4ce895202e8R1)

matthijsgroen:
It would be awesome to write code like this:

```js
import { entity } from "crafty/core";
import Keyboard from "crafty/components/keyboard";
import Motion from "crafty/components/motion";

const ent = entity(Keyboard, Motion);
```

This way we can add a lot of components to Crafty, and tools like rollup or webpack will only include the actual needed code 👍 

For now I use this construct for my components:

```js
import Crafty from "src/crafty";
const ControlScheme = "ControlScheme";

Crafty.c(ControlScheme, {
  init: function() {}
});

default export ControlScheme;
```

This way I force myself to import the component before using, and letting the linter warn me if I imported Components I'm not using anymore.

and in the `src/crafty` I prevent it from being a global:

```js
import "craftyjs/dist/crafty";

const Crafty = window.Crafty;
delete window["Crafty"];

export default Crafty;
```

PR #921, opened 2015-06-05
A link where you can easily see the example online and edit yourself

kevinsimper:
I think examples where you can easily see and play around with the code itself is very cool!


starwed:
Good idea, but maybe it should be a `[link]()` instead of a raw url?

kevinsimper:
Your right! :+1: I changed the line to be a link instead!

starwed:
Thanks!

PR #922, opened 2015-06-10
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,5 @@
 {
   "name": "crafty",
-  "version": "0.6.2",
   "main": "dist/crafty.js",
   "license": "MIT",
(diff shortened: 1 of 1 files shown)


kevinsimper:
Actuatlly we should remove bower.json as we have another repo for distributing to bower https://github.com/craftyjs/Crafty-Distro

starwed:
We definitely need to sort out our bower story.  Might as well merge this for now, though; no harm if its removed later.

kkirsche:
Also for what it's worth Bower may be under NPM anyway as that's a consideration bower/bower#1520

PR #923, opened 2015-06-15
Motion tweaking

starwed:
Several small issues with Motion/Control components
- Fix bugs with NewDirection (and angular version)
  - Fixes a rounding issue when calculating the sign of negative fractions
  - Triggers NewDirection in the frame where the direction changes, instead of the next frame
  - Renames NewRotation to NewAngularDirection and fixes the same issues
  - Adds/fixes related tests
- Sync motion with clock rather than number of frames
  - Instead of adjusting the speed based on the frame rate, try to saty in sync with the game clock by using frame.dt.  Note that speeds are in pixels per second.
  - This is a **breaking change**, because it changes how we interpret speeds.  (Pixels per second instead of pixels per frame.)
  - Since it's a breaking change, I needed to fix several tests.  I found the existing tests very hard to follow, so I added a lot of clarification, and removed some that seemed redundant.
- Change Jumpway -> Jumper
-  Reword motion docs

### DIFF
--- a/src/controls/controls.js
+++ b/src/controls/controls.js
@@ -144,7 +144,7 @@ Crafty.c("Multiway", {
         this._keyDirection = {}; // keyCode -> direction
         this._activeDirections = {}; // direction -> # of keys pressed for that direction
         this._directionSpeed = {}; // direction -> {x: x_speed, y: y_speed}
-        this._speed = { x: 3, y: 3 };
+        this._speed = { x: 150, y: 150 };
 
         this.bind("KeyDown", this._keydown)
             .bind("KeyUp", this._keyup);
@@ -185,7 +185,7 @@ Crafty.c("Multiway", {
      * #.multiway
      * @comp Multiway
      * @sign public this .multiway([Number speed,] Object keyBindings)
-     * @param speed - Amount of pixels to move the entity whilst a key is down
+     * @param speed - A speed in pixels per second
      * @param keyBindings - What keys should make the entity go in which direction. Direction is specified in degrees
      *
      * Constructor to initialize the speed and keyBindings. Component will listen to key events and move the entity appropriately.
@@ -193,8 +193,8 @@ Crafty.c("Multiway", {
      *
      * @example
      * ~~~
-     * this.multiway(3, {UP_ARROW: -90, DOWN_ARROW: 90, RIGHT_ARROW: 0, LEFT_ARROW: 180});
-     * this.multiway({x:3,y:1.5}, {UP_ARROW: -90, DOWN_ARROW: 90, RIGHT_ARROW: 0, LEFT_ARROW: 180});
+     * this.multiway(150, {UP_ARROW: -90, DOWN_ARROW: 90, RIGHT_ARROW: 0, LEFT_ARROW: 180});
+     * this.multiway({x:150,y:75}, {UP_ARROW: -90, DOWN_ARROW: 90, RIGHT_ARROW: 0, LEFT_ARROW: 180});
      * this.multiway({W: -90, S: 90, D: 0, A: 180});
      * ~~~
      *
@@ -234,12 +234,13 @@ Crafty.c("Multiway", {
      * @sign public this .speed(Object speed)
      * @param speed - New speed the entity has, for x and y axis.
      *
-     * Change the speed that the entity moves with. 
+     * Change the speed that the entity moves with, in units of pixels per second.
+     *
      * Can be called while a key is pressed to change speed on the fly.
      *
      * @example
      * ~~~
-     * this.speed({ x: 3, y: 1 });
+     * this.speed({ x: 150, y: 50 });
      * ~~~
      */
     speed: function (speed) {
@@ -360,7 +361,7 @@ Crafty.c("Multiway", {
  * @see Supportable, Motion, Keyboard, Gravity
  */
 Crafty.c("Jumpway", {
-    _jumpSpeed: 6,
+    _jumpSpeed: 300,
 
     /**@
      * #.canJump
@@ -441,7 +442,7 @@ Crafty.c("Jumpway", {
      * #.jumpway
      * @comp Jumpway
      * @sign public this .jumpway([Number jumpSpeed,] Array jumpKeys)
-     * @param jumpSpeed - Vertical jump speed
+     * @param jumpSpeed - Vertical jump speed in pixels per second
      * @param jumpKeys - Keys to listen for and make entity jump in response
      *
      * Constructor to initialize the power of jump and keys to listen to. Component will
@@ -450,8 +451,8 @@ Crafty.c("Jumpway", {
      *
      * @example
      * ~~~
-     * this.jumpway(6, ['UP_ARROW', 'W']);
-     * this.jumpway(['UP_ARROW', 'W']);
+     * this.jumper(300, ['UP_ARROW', 'W']);
+     * this.jumper(['UP_ARROW', 'W']);
      * ~~~
      *
      * @see Supportable, Motion, Keyboard, Gravity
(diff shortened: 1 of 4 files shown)


starwed:
As mentioned, this does have a breaking change: it changes speeds from pixels per frame to pixels per second.

I think this is a more natural definition in most cases -- I often had to choose very small speeds or accelerations with the old units.  Any arguments to  keep it as is, though?

starwed:
A quick note about tests: lots of tests are certainly way better than none!  But I wanted to give a few recommendations.  (I'm not trying to pick on the tests for Motion, they just happened to be what I happened to interface with on this patch!  And the thorough testing definitely let me feel good about making the changes!  :)  )
- If tests don't have a description string, it becomes a bit difficult to discover which specific test is failing.
- Ideally, each `test()` block should test one set of things in reasonable isolation, unless it's _supposed_ to be an integration test.  (Those are good too!)
- There's no need to test every aspect of a behavior.  (For instance, a lot of the tests for multiway checked that the velocity was set correctly _and_ that the entities moved the correct distance -- the latter is actually testing "Motion", not "Multiway".)

tests/2d.js
```diff
@@ -562,106 +624,105 @@
...
     });
-    e.one("MotionChange", function(evt) { strictEqual(evt.key, "vrotation"); strictEqual(evt.oldValue, -1); });
     e.vrotation = 0;
     Crafty.timer.simulateFrames(1);
 
-    equal(newDirectionEvents, 4);
-    equal(newRevolutionEvents, 3);
-    equal(movedEvents, 6);
-    equal(rotatedEvents, 2);
-    equal(motionEvents, 17);
```
mucaho, line 664: this checks shouldn't be removed, as we can not guarantee that all event callbacks fired as we wanted

mucaho:
I have looked over the PR, good job!
Specifying speed in pixels per second is the most groundbreaking change I have witnessed so far. It's the right thing to do, but this will probably break all existing games! That's why I initially proposed we stick to `dt * 0,05`.
Nice job identifying the problem with the `Math.sign` solution from [SO](http://stackoverflow.com/a/23262976/3041008).

The only thing I would like to nitpick is the order of the `NewDirection` event. The motivation behind it was that `NewDirection` should be fired just before the velocity/acceleration was applied. If user has custom `EnterFrame` callback that changes velocity/acceleration, he can not be sure whether it is applied before/after `Motion` changes the velocity/acceleration. By binding to the `NewDirection` event he can make sure what combined velocity/accel will be applied and still change it to suit his needs (e.g. he can prevent moving left).
Initially I had the `NewDirection` event fire after `Motion` changed it (as it's natural), but maybe there is a corner case later where the paragraph above will be relevant. We can leave it as proposed in this PR for now.

starwed:
> Initially I had the NewDirection event fire after Motion changed it (as it's natural), but maybe there is a corner case later where the paragraph above will be relevant. We can leave it as proposed in this PR for now.

Another possibility would just be to trigger this when the velocity is actually changed.  (So in the setter for the velocity properties.)  That way it wouldn't matter _how_ the velocity was changed.

starwed:
> Specifying speed in pixels per second is the most groundbreaking change I have witnessed so far. It's the right thing to do, but this will probably break all existing games! That's why I initially proposed we stick to dt \* 0,05.

I think there are enough breaking changes since the last version that pretty much any game will require a bit of work to update, so I'm not too worried about this.  I find it makes it easier to reason about how speeds and accelerations, so it should be worth it.

mucaho:
>  I find it makes it easier to reason about how speeds and accelerations, so it should be worth it.

It also plays nicely with tweening.

Ok then, I suggest you readd [some test checks](https://github.com/craftyjs/Crafty/pull/923#discussion_r32488655), as well as update the default values for movement and their respective documentation (`_speed` in `Multiway` and `_jumpSpeed` in `Jumper`).
After that we can merge this :)

mucaho:
> Another possibility would just be to trigger this when the velocity is actually changed. (So in the setter for the velocity properties.) That way it wouldn't matter how the velocity was changed.

Yes, that would be great! Note that it will introduce some overhead probably.

starwed:
Ok, I re-added those tests.  (A better fix would be to break up that long litany of tests into several isolated ones, and then check that each event was called individually, but I think this is fine for now.)

I also adjusted the default speeds/accelerations to match the new time-scale.

src/controls/controls.js
```diff
@@ -450,13 +451,13 @@ Crafty.c("Jumpway", {
      *
      * @example
      * ~~~
-     * this.jumpway(6, ['UP_ARROW', 'W']);
-     * this.jumpway(['UP_ARROW', 'W']);
+     * this.jumper(6, ['UP_ARROW', 'W']);
```
mucaho: update speed here, e.g. `300`

src/controls/controls.js
```diff
@@ -470,14 +471,14 @@ Crafty.c("Jumpway", {
...
         return this;
     },
 
     /**@
      * #.jumpSpeed
-     * @comp Jumpway
+     * @comp Jumper
      * @sign public this .jumpSpeed(Number jumpSpeed)
      * @param jumpSpeed - new vertical jump speed
      *
```
mucaho, line 484: update speed here in example below

src/controls/controls.js
```diff
@@ -234,7 +234,8 @@ Crafty.c("Multiway", {
      * @sign public this .speed(Object speed)
      * @param speed - New speed the entity has, for x and y axis.
      *
-     * Change the speed that the entity moves with. 
+     * Change the speed that the entity moves with, in units of pixels per second.
+     *
      * Can be called while a key is pressed to change speed on the fly.
      *
      * @example
```
mucaho, line 241: update speed here in example below

src/controls/controls.js
```diff
@@ -185,7 +185,7 @@ Crafty.c("Multiway", {
      * #.multiway
      * @comp Multiway
      * @sign public this .multiway([Number speed,] Object keyBindings)
-     * @param speed - Amount of pixels to move the entity whilst a key is down
+     * @param speed - A speed in pixels per second
      * @param keyBindings - What keys should make the entity go in which direction. Direction is specified in degrees
      *
      * Constructor to initialize the speed and keyBindings. Component will listen to key events and move the entity appropriately.
```
mucaho, line 191: update speed here in example below

src/spatial/2d.js
```diff
@@ -1182,9 +1184,9 @@ Crafty.c("Gravity", {
      * #.gravityConst
      * @comp Gravity
      * @sign public this .gravityConst(g)
-     * @param g - gravitational constant
+     * @param g - gravitational constant in pixels per second squared
      *
-     * Set the gravitational constant to g. The default is 0.2 . The greater g, the faster the object falls.
+     * Set the gravitational constant to g for this entity. The default is 50. The greater g, the stronger the downwards acceleration.
```
mucaho: default is `500`

mucaho:
Great, I noticed some minor things that should be updated in documentation.

> Another possibility would just be to trigger this when the velocity is actually changed. (So in the setter for the velocity properties.) That way it wouldn't matter how the velocity was changed.

Should this be in current or seperate PR?

starwed:
Good catch with the examples.

> Should this be in current or seperate PR?

If we do that, it'll be a separate PR.  Since there's an event any time the velocity changes, I'm not sure how important it is to have a separate one that occurs for a particular type of change.

starwed:
Ok, all the examples should be updated now.

letmaik:
I just noticed this since I use the nightlies:
http://neothemachine.github.io/hungrymonkey/

I think to make the transition easier it would good to provide some rough formula to get from pixels per frame to pixels per second, assuming some common framerates like 60fps.

mucaho:
To go from old `[px / frame]` to new `[px / s]` you multiply it by the target frame rate, which defaults to `50`, e.g.:
- old `jumpSpeed = 6 [px / frame]` -> new `jumpSpeed = 300 [px / s]`
- old `gravityConst = 0.2 [px / frame²]` -> new `gravityConst = 0.2 * 50² = 500 [px / s²]`

Good idea, we should probably add this to the changelog

letmaik:
Yep, that worked. Thanks

PR #925, opened 2015-07-11
Diamond iso

airza:
I think the documentation might need fixing, but you can do lots of cool new things.  In particular, pathfinding (there is kind of an inefficient part of it that is O(n) in the size of the grid instead of O(1) though.).  It checks for both empty tiles on the lower part of the grid as well as objects with the component "Obstacle" (There might be some way this can better conform with the rest of the library, but since 2d collision and grid collision don't work the same I wasn't sure.

Also, i made the simplifiying assumption that grids would be 1 ground level and 1 upper level for the purposes of placement and pathfinding - most of the code _probably_ works in 3D for placement but that gets into some deep issues if we have no rotation method.

I've tested it a lot on my game in development, but that game doesn't really try to move the grid or do anything different than the square tiles.


airza:
I don't think I squashed the commits right :s

airza:
Okay, now the documentation at least isn't lying anymore.

airza:
I pulled the latest changes, tested against the game I have in development.  Doesn't break centerAt, has documentation, and adds new functionality!

airza:
Is there anything I need to do in order to get this merged in?  Haven't had the time recently to look at the newsgroup.

Issue #926, opened 2015-07-16
Cross-browser compatibility issues


mucaho:
- Firefox 4 and Opera 12 pass all tests except the [`avoidCss3dTransform` ones](https://github.com/craftyjs/Crafty/blob/0974958727d6f7d256406c81effffecdd0c69ed6/tests/dom.js#L6-L39).
- IE 9 fails 6 tests, will need to install and see which ones exactly.

ping #609, #641, #642, @pawurb 

**EDIT:** we are in luck, IE9 is the only browser available in [browserling](https://www.browserling.com/)'s free plan
![ie9_bugs](https://cloud.githubusercontent.com/assets/3935691/8754050/ff2a6290-2cc0-11e5-947d-192b0b29362b.png)

PR #930, opened 2015-07-17
Replace statements of console log with Crafty log

mucaho:
console log fails on IE9, this PR replaces all `console.log` with `Crafty.log` statements, both in the documentation and source where applicable
fixes part of #926


mucaho:
pretty straightforward change that mostly affects documentation, merging

PR #931, opened 2015-07-19
Enhance Multiway implementation

mucaho:
Benefits:
- Limit diagonal movement to same magnitude as horizontal or vertical movement.
- Allow diagonal movement and arbitrary direction movement with correct speed when pressing down multiple keys (may be of interest for isometric)

Caveat:
- Takes an additional 5% of total frame time on top of current approach in worst-case scenario (smashing combination of 8 keys on a blank screen with 1 entity).

Notes:
- See the [8-way example](https://github.com/craftyjs/demos/blob/master/devDemos/example_8way.html).
- Velocity values handled by Multiway are truncated, so diagonal movement is approximated (can be changed).
- Solves issue described [here](https://github.com/craftyjs/Crafty/pull/794#issuecomment-76159641)
- I tried many alternatives to the current implementation, but my benchmarks showed it's not worth it on newest firefox and chrome
  - checking if we move in only one direction / diagonal so we can avoid `Math.sqrt` calculations, [not worth it](http://jsperf.com/movement-direction)
  - memoization of input parameters, [not worth it](http://jsperf.com/crafty-movement-memoization)
- There is probably a faster approach to handle diagonal movement only (current one is for generic directions)

### DIFF
--- a/src/controls/controls.js
+++ b/src/controls/controls.js
@@ -136,14 +136,21 @@ Crafty.c("Draggable", {
  * @see Motion, Keyboard
  */
 Crafty.c("Multiway", {
-    _speed: null,
-    
+    _speed: null, // { x: speedX, y: speedY }
+    _keyDirection: null, // keyCode -> direction
+
+    _activeDirections: null, // direction -> # of keys pressed for that direction
+    _directionMagnitudes_e6: null, // direction -> { x: magnitudeX, y: magnitudeY }
+
+    _activeSpeedX: 0, // currently active velocity of Multiway on x-axis
+    _activeSpeedY: 0, // currently active velocity of Multiway on y-axis
+    _activeMagnitudeX_e6: 0, // currently active denormalized direction of movement on x-axis
+    _activeMagnitudeY_e6: 0, // currently active denormalized direction of movement on y-axis
+
     init: function () {
         this.requires("Motion, Keyboard");
 
-        this._keyDirection = {}; // keyCode -> direction
-        this._activeDirections = {}; // direction -> # of keys pressed for that direction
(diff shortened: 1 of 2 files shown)


mucaho:
I'm still on the fence whether the additional correctness & complexity is needed for Multiway or not (at least for its current uses).
What do you think @starwed? I'd like to hear your input on this.

src/controls/controls.js
```diff
@@ -155,15 +162,35 @@ Crafty.c("Multiway", {
...
+                /* Following is an optimization of these statements:
+                if (!this.disableControls) this.__unapplyActiveSpeed();
+                this.__addActiveMagnitude(direction);
+                this.__updateActiveSpeed();
+                if (!this.disableControls) this.__applyActiveSpeed();
+                */
+                var oldActiveSpeedX = this._activeSpeedX,
+                    oldActiveSpeedY = this._activeSpeedY;
+
+                var magnitude_e6 = this._directionMagnitudes_e6[direction],
```
starwed, line 181: What's the point of all the `e6` stuff?

mucaho, line 181: Each direction vector is added to the net movement vector when key is pressed.
Each direction vector is substracted from the net movement vector when key is released.
Pressing and then releasing 5 different direction keys should net into 0 movement.
However, due to rounding errors, you don't get 0.
That's why this net movement vector is represented as a number `* 10e6`.

mucaho:
I have put some more thought to it, it's just overly complicated.
Who needs more than 8 directions for keyboard control?
Joystick / Gamepad support (with arbitrary directions) is already provided by [external component](https://github.com/svenjacobs/crafty-gamepad).

starwed:
Closing, since mucaho ended up feeling it was too complicated, and it's also superseded by #1051

Issue #792, opened 2014-06-29
Viewport entity following and canvas zoom/scale conflict bug


georgii-ivanov:
Possible bug is reported in forum post: https://groups.google.com/forum/#!searchin/craftyjs/viewport$20follow$20zoom/craftyjs/lApstQMSg3c/lbsI8t-UFtEJ

Please read this topic and fix it.

JsFiddle here:

http://jsfiddle.net/drG26/3/

mucaho:
This should have been fixed with #934 now, if you encounter any more problems with it feel free to reopen this issue.

Issue #754, opened 2014-03-14
viewport.follow scaling issue


mucaho:
Possible bug reported in [forum post](https://groups.google.com/d/msg/craftyjs/kHu5MAv19BM/rLlpaM6YCiQJ)

> The problem is that .follow doesn't have .scale in account when it calculates how much it have to displace the viewport to center the entity. I've solved the problem with a custom center method, bound to the 'enterFrame' event of the centered entity:
> 
> ``` javascript
>   this.centerViewport = function(entity) {
>     Crafty.viewport.y = (4/3 * Crafty.viewport.height / 2 / Crafty.viewport._scale) - entity.y;
>     Crafty.viewport.x = (Crafty.viewport.width / 2 / Crafty.viewport._scale) - entity.x;
>   }
> ```

I haven't investigated yet. Is this a bug or intended behaviour?

georgii-ivanov:
It is my problem too https://github.com/craftyjs/Crafty/issues/792

mucaho:
This should have been fixed with #934 now, if you encounter any more problems with it feel free to reopen this issue.

mucaho:
`Crafty.viewport.follow` doesn't center screen after `Crafty.viewport.scale()` is called and before entity is moved. Solution: Listen for `ViewportScale` event in addition to `Change` event to trigger center on entity.

mucaho:
> Crafty.viewport.follow doesn't center screen after viewport transformations

fixed with 871ae034d3a46d9eb5dece6ea03e0f74c861346d

PR #934, opened 2015-07-21
Fix viewport methods when viewport scaled

mucaho:
Collection of fixes related to #792, #754 .
This [example](https://github.com/craftyjs/demos/blob/master/devDemos/example_viewport_zoom.html) should correctly display once PR merged.


mucaho:
Long overdue fix, relatively small in scope, merging.

PR #936, opened 2015-07-22
Fix rounding problem with 90 deg rotation

mucaho:
Rounding errors seem to occur randomly, hence hard to test against.


mucaho:
Small fix, is already in library [in some other place](https://github.com/craftyjs/Crafty/blob/ed3c58bc899716a762cbb89f29e3ee919a077522/src/spatial/2d.js#L361-L363), merging.

Issue #904, opened 2015-05-16
Manipulating velocity conflicts with Multiway


mucaho:
As described [in this post](https://github.com/craftyjs/Crafty/issues/903#issuecomment-102533377) resetting velocity `vx = 0` while a key is pressed will lead to undesired behavior after releasing said key.
[Relevant lines](https://github.com/craftyjs/Crafty/blob/e79a84cc5a39a39603b7020192301da1944258fc/src/controls/controls.js#L158-L178):

``` js
_keydown: function (e) { ...
    this.vx += directionSpeed.x;
    this.vy += directionSpeed.y;
... },

_keyup: function (e) { ...
    this.vx -= directionSpeed.x;
    this.vy -= directionSpeed.y;
... },
```

ashes999:
I just ran into this again today. Is there a work-around or fix? Maybe a way to specify an impulse?

starwed:
The workaround would be to not use Multiway, but use it as a model to build your own component.

Multiway should probably be fixed so that letting go of the key will never change direction.

mucaho:
there is no solution to this problem I can think of so far:
- snap velocity hard
  - won't reactivate movement while key is pressed after `v` was set to `0` upon collision event (as in linked code example) - e.g. player temporarily collides with asteroid, then asteroid moves away, but player doesn't move while key is still being pressed
  - won't behave correctly with diagonal movement (where keys affect both `vx` and `vy`)
  - doesn't interact nicely with other features that might affect velocity (e.g. detached entity could retain former parent's velocity in addition to its current velocity)

``` js
_keydown: function (e) { ...
    this.vx = directionSpeed.x;
    this.vy = directionSpeed.y;
... },

_keyup: function (e) { ...
    this.vx = 0;
    this.vy = 0;
... },
```
- check for current velocity in each frame and adjust it if it's different from desired velocity
  - won't behave correctly with diagonal movement (where keys affect both `vx` and `vy`)
  - doesn't interact nicely with other features that might affect velocity (e.g. detached entity could retain former parent's velocity in addition to its current velocity)

``` js
_keyup: function (e) { ...
    if (abs(directionSpeed.x) > 0) this.vx = 0;
    if (abs(directionSpeed.y) > 0) this.vy = 0;
... },

_enterframe: function() {...
   if (this.isDown(key)) {
      if (this.vx < directionSpeed.x && directionSpeed.x > 0) this.vx = directionSpeed.x;
[...]

Issue #899, opened 2015-05-04
Improve checkHits and onHit functionality


mucaho:
Feature request from [forum post](https://groups.google.com/forum/#!topic/craftyjs/GOQvUYrc0jU)

Without changing the ease-of-use of `checkHits` and `onHit` `Collision` methods, they could also inform user that the colliding entities, having the desired component, have changed. 
Possibilities include introducing a new event that is fired each time the colliding entities change (e.g. `HitOn(component)` -> `HitChange(component, oldestHitList)` -> ... -> `HitChange(component, newestHitList)` -> `HitOff(component)` ), or restructuring the event completely (e.g. seperate `HitOn(component, entity)` & `HitOff(component, entity)` events for each colliding entity.)

grozen:
This can definitely be useful, but won't using `resetHitChecks` from the respective `HitOn` handler be sufficient? This will cause additional `HitOn`s to fire even though the collision is still "active". While not as convenient, it will keep the performance penalty an opt-in affair, as what we are essentially suggesting will incur some sort of penalty. Not a very big one, mind you, though I guess that is ultimately up to how many different entities you are colliding with and need to track.

mucaho:
Very fair point regarding the performance - we would need to keep track of each individual entity that is colliding. 
What's the purpose of `resetHitChecks` exactly? Wouldn't it be easier to trigger a separate event (e.g. `HitList(component, colliders)`) with the current colliding entities **every frame**? Is there a use-case I am missing, where resetting the notification about current colliding entities would come in handy?

grozen:
Having collision events fire each frame even while a collision is underway (`HitOn` already returns all the colliders as part of the event data) is equivalent to calling `resetHitChecks` constantly (for example, at the end of a `HitOn` handler).

The only reason I implemented the events this way as part of #147 is that the notion of notifying about a collision only once was already in the code-base. If I remember correctly, I added `resetHitChecks` precisely to circumvent that behavior when desired.

mucaho:
Ok, thank you for clarifying. I thought about it long and hard and think have come to a solution which offers the requested feature (collision events per entity), improves performance (in some cases) and streamlines the current implementation:
[...]

PR #938, opened 2015-07-22
Enhance docs

### LINKED ISSUE #932: Custom collision hitbox ignored
I added the `Collision` component to my `Player` component, and specified the hitbox to match the height of the sprite that I'm using using the `collision()` method. However, when I call the `onHit` method, it gets ran when the hitbox is _not_ intersecting the object in question - it seems to only care about the position of the entire entity.

I have a jsfiddle to demonstrate the issue:  http://jsfiddle.net/nzifnab/LvarggL1/3/

The blue box is the player entity itself, but the brownish box is what I see when I add the "SolidHitBox" component so I can see where the hitbox is.

If you use the arrow keys to move to the right, that entity box is the one that is hitting the block and stopping your movement -- not the hitbox. The player should be able to pass under the block without issue, but cannot.

Any idea what I'm doing wrong?

EDIT: I made the jsfiddle use fourway to make it easier to move around the block and see where the collision is occurring.

Edit2: I attempted to use the nightly build of crafty here: http://jsfiddle.net/nzifnab/LvarggL1/5/

This exhibits the same behavior where the hitbox is not working properly. I had to add a 'console.log' instead of stopping movement, because `this._movement` appears to have been changed/removed.

mucaho:
fixes #932 , it's mentioned 3 times now, hopefully no more confusion in future
gives alternatives to feature request described in #899 
tries to eliminate problems as described in #904

### DIFF
--- a/src/controls/controls.js
+++ b/src/controls/controls.js
@@ -191,6 +191,10 @@ Crafty.c("Multiway", {
      * Constructor to initialize the speed and keyBindings. Component will listen to key events and move the entity appropriately.
      * Can be called while a key is pressed to change direction & speed on the fly.
      *
+     * Multiway acts by adding a velocity on key press and removing the same velocity when the respective key is released.
+     * This works well in most cases, but can cause undesired behavior if you manipulate velocities by yourself while this component is in effect.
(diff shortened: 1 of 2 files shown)


mucaho:
doc only improvement, merging

PR #940, opened 2015-07-23
Fix rotation direction bug and simplify

mucaho:
Small typo has caused bug that is not picked up by tests ([L1382](https://github.com/starwed/Crafty/blob/cafd964727b67b43e14cf0fb92b3258b5476c378/src/spatial/2d.js#L1382) of cafd964727b67b43e14cf0fb92b3258b5476c378)
Simplify direction event for velocity to increase performance.


mucaho:
small fix, merging

PR #941, opened 2015-07-23
Fix loader compatibilty with OpalRB

ashes999:
See: https://github.com/opal/opal/issues/1024. OpalRB adds a bunch of properties to arrays (also see: https://github.com/opal/opal/issues/680). Iterating, here, ends up with an exception, because `current` becomes an instance of `OpalClass`.

Possible solutions:
- Using `for .. of` 
- Calling `continue` if `data[type][asset].hasOwnProperty(current)` is false

I initially verified with the latter solution, but someone suggested that this works better.


mucaho:
~~Calling Crafty.load with an array of assets no longer works; see the docs for more details.
Does calling it with an object fix your issue?~~
Ah, misunderstood, if we change it to `for (var i=0; i<array.size; i++) { foo(array[i]) }` would that work?

ashes999:
@mucaho  I know. I'm passing in an object with arrays. That doesn't fix my issue.

Sample input:

```
{ images: [ '1.png', '2.png'], audio: ['1.ogg'] }
```

Embarrassingly, my fix doesn't work/compile in Chrome. I'll submit an update in a sec.

ashes999:
@mucaho I don't think that will work. Hash/objects don't have a `size` or `length`. Using `of` didn't work, either. I think this is the best solution I have right now. (I don't know why the CI build failed -- maybe I butchered the newlines?)

mucaho:
Aha ok so the data passed to `Crafty.load` can look like this

``` js
data = {
    "audio": {
        "beep": ["beep.wav", "beep.mp3", "beep.ogg"],
        "boop": "boop.wav",
        "slash": "slash.wav"
    },
    "images": ["badguy.bmp", "goodguy.png"],
    "sprites": {
        "animals.png": {
            "tile": 50,
            "tileh": 40,
            "map": { "ladybug": [0,0], "lazycat": [0,1], "ferociousdog": [0,2] }
            "paddingX": 5,
            "paddingY": 5,
            "paddingAroundBorder": 10
        },
        "vehicles.png": {
            "tile": 150,
            "tileh": 75,
            "map": { "car": [0,0], "truck": [0,1] }
        }
    }
}
```

`Crafty.load` iterates over this data like this:

``` js
for (type in data) {
    for(asset in data[type]) {
       var current = data[type][asset];
       foo(current);
    }
}
```

On the 2nd iteration level Crafty does not distinguish between an object or an array being iterated on and that seems to give you trouble in the case the `for ... in ...` statement iterates over an array, right?

Your 2nd proposed solution seems reasonable (ES6 compatibility is lightyears away :) ). On my end it looks like that something has gone wrong with your commit (it shows me 14k line additions and 400 line deletions).

ashes999:
@mucaho I didn't test all those cases. My core change is this addition:

```
if ((typeof current) != 'string') {
  continue;
 }
```

Those lines go [here](https://github.com/craftyjs/Crafty/blob/develop/src/core/loader.js#L308).

I also only tested it in Chrome.

I'm not sure how to "fix" my PR. If you want to reject it, and submit the change yourself, that's cool with me. Or I can close it and try to create a new one.

mucaho:
Your commit seems to fail because `current` is allowed to be an array.
e.g.

``` js
data = {
    ...,
    "images": ["badguy.bmp", "goodguy.png"],
    ...
}
```

> Calling `continue` if `data[type][asset].hasOwnProperty(current)` is `false`

This should do the trick. Note that there may be other places in the library that need a similar adjustment. This is fine in the general case (like here), but in performance critical code that gets executed ~@50Hz every additional, "unnecessary" statement adds up, so we will probably have to profile if that really has an impact there or not.

ashes999:
Your suggestion doesn't seem to work --  it skips over the strings from the images/audio arrays.  

Any other ideas?

mucaho:
Last thing that pops to my mind is to check whether `data[type]` is an array or an object `data[type].constructor === Array` and then iterate accordingly.
I don't know how Opal works, the closed PR used the [solution proposed by the owner himself](https://github.com/opal/opal/issues/680#issuecomment-68798157) - maybe he proposed the solution for object iteration, rather than array iteration as described in the problem statement?

ashes999:
@mucaho I don't understand exactly the change that you're proposing.  I can, however, test it and make sure it works with at least the one case I have.

Right now, your PR doesn't fix my problem.

ashes999:
I stand corrected. Thanks for fixing this so fast.

mucaho:
I have to admit, I'm confused now. What is the actual fix so we can merge it? :)

ashes999:
@mucaho your PR that you submitted fixed the problem. I have some other problem where I ran something which didn't include your fix. My problem is solved, with your commit.

What I'm confused about, is why I don't see your commit in the craftyjs/Crafty commit history. But anyway, I did a manual clone + apply of your change, and I'm good to go now.

Ajedi32:
That PR was closed without being merged, so this actually isn't fixed yet.

mucaho:
Ok, I used some `rebase` magic, maybe that messed up things. Gonna go ahead and merge the relevant PR.

Issue #903, opened 2015-05-12
Gravity + Jetpack?


ashes999:
I'm trying to add "jetpack"-like functionality to my player. When I add the `Gravity` component, I get the following (desirable) behaviours:
- Left/right to move left/right
- Up to jump
- Fall until you hit something solid

What I would like to figure out, is how I can press/hold the up button and fly upward. Is it enough to bind the `EnterFrame` event and add to the player's `y` position? Is there a better way to do this?

NB: if jumping gets disabled in the process, that's cool too.

mucaho:
You could achieve in different ways, as you probably noticed.
[Here is a quick jsfiddle](https://jsfiddle.net/6tf6zbt6/2/) - benefits are you work with vertical velocity so it plays somewhat nicely with gravity.
By adding vertical velocity each frame (up to a limit) you get a nice sluggishness effect.

``` js
.bind("EnterFrame", function(frameData) {
  if (this.isDown("W")) {
    this.vy = Math.max(-5, this._vy - 2.5);
  }
})
```

I would suggest to use the forum for discussing usages of the library and use the issues to post feature request if it's not something trivially doable with the library.

ashes999:
Noted, thanks. I'll close this issue then.

mucaho:
Hold up, wait a minute :) Does it work as you intended?

ashes999:
No, not really. Thanks though.

mucaho:
What's not working? I just tried, it should also work with normal jump behavior: [new jsfiddle](https://jsfiddle.net/6tf6zbt6/4/)

ashes999:
It's not that it doesn't work (although with the new js fiddle, if you hold `w`, you just jump normally) -- it's two things:

1) I find the pre-existing Crafty components really dense. I can't understand them (especially how to extend them), especially without looking at the source.
2) The source, and examples from you, are usually the latest (dev) versions. I have to download and rebuild the local sources, instead of using the latest production version (0.6.2 as of writing).

I'm using CraftyJS for prototyping. The perception is (rightly) "well I can wire up gravity, collision, etc. in a matter of minutes" -- but changing those behaviours (eg. can't jetpack, can't figure out how to make this work) is really a hassle.

As much as you are extremely helpful and provide working examples (something very rare in GitHub culture), I can't expect you to write all my code for me. I don't know where that leaves me, because there's no other tool that gives me this much, this fast.

Like in this case; I had to again resort to using nightlies.

(end soapbox)

[...]

PR #944, opened 2015-07-24
Gravity terminal velocity

mucaho:
Should solve issue as described in 1st paragraph of [this comment](https://github.com/craftyjs/Crafty/issues/903#issuecomment-102638698).
Basically this PR limits downwards velocity of gravity so that it doesn't reach astronomical values.
So this prevents you from falling through floor automatically, if you leave the `terminalVelocity` to the default value (because it's based on entities `h` and maximum allowed frame delta time).
Also some documentation about the effect associated to it - tunneling.

Alternative thoughts:
- Should this be even changeable / documented to user? For player characters with `h = 32` that means maximum movement of `32 px` per frame, which is pretty fast imo. (I haven't noticed any limiting of downwards velocity in an example platformer I tested).
- Have velocity limits be part of Motion? So far we need it just for gravity internally, but could be useful for user created content.

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -1379,32 +1379,45 @@ Crafty.extend({
             /**@
              * #Crafty.timer.steptype
              * @comp Crafty.timer
+             *
+             * @trigger NewSteptype - when the current steptype changes - { mode, maxTimeStep } - New steptype
+             *
+             * Can be called to set the type of timestep the game loop uses.
              * @sign public void Crafty.timer.steptype(mode [, maxTimeStep])
-             * Can be called to set the type of timestep the game loop uses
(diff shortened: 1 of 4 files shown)


mucaho:
@ashes999 , is that what you meant by the first part of your comment? I hope that should fix those cases.

ashes999:
@mucaho based on the comments and code changes, given my very limited understanding of how CraftyJS works, I think this would resolve my tunneling/jittering issues.

mucaho:
Found faster, easier and more reliable way to implement this, which will come in seperate PR. The original implementation described above does not bode well with varying frame times.

Merging some additional tests & minor timer additions for now.

PR #949, opened 2015-07-28
Some code style fixes

Aliance:
wrong jsdoc

### DIFF
--- a/src/spatial/collision.js
+++ b/src/spatial/collision.js
@@ -1,7 +1,6 @@
 var Crafty = require('../core/core.js'),
     DEG_TO_RAD = Math.PI / 180;
 
-
 /**@
  * #Collision
  * @category 2D
@@ -29,7 +28,6 @@ Crafty.c("Collision", {
 
         this.collision();
     },
-
 
     // Run by Crafty when the component is removed
     remove: function() {
@@ -83,8 +81,6 @@ Crafty.c("Collision", {
         this.unbind("Resize", this._resizeMap);
         this.unbind("Resize", this._checkBounds);
 
-
-
         if (!poly) {
             // If no polygon is specified, then a polygon is created that matches the bounds of the entity
             // It will be adjusted on a "Resize" event
@@ -131,7 +127,6 @@ Crafty.c("Collision", {
         return this;
     },
 
-
     // If the hitbox is set by hand, it might extend beyond the entity.
     // In such a case, we need to track this separately.
     // This function finds a (non-minimal) bounding circle around the hitbox.
@@ -139,10 +134,10 @@ Crafty.c("Collision", {
     // It uses a pretty naive algorithm to do so, for more complicated options see [wikipedia](http://en.wikipedia.org/wiki/Bounding_sphere).
     _findBounds: function(points) {
         var minX = Infinity, maxX = -Infinity, minY=Infinity, maxY=-Infinity;
-        var p, l = points.length;
+        var l = points.length;
 
         // Calculate the MBR of the points by finding the min/max x and y
-        for (var i=0; i<l; i+=2){
+        for (var i=0; i<l; i+=2) {
             if (points[i] < minX)
                 minX = points[i];
             if (points[i] > maxX)
@@ -156,16 +151,16 @@ Crafty.c("Collision", {
         // This describes a circle centered on the MBR of the points, with a diameter equal to its diagonal
         // It will be used to find a rough bounding box round the points, even if they've been rotated
         var cbr = {
-                cx: (minX + maxX) / 2,
-                cy: (minY + maxY) / 2,
-                r: Math.sqrt( (maxX - minX)*(maxX - minX) + (maxY - minY)*(maxY - minY))/2,
+            cx: (minX + maxX) / 2,
+            cy: (minY + maxY) / 2,
+            r: Math.sqrt((maxX - minX)*(maxX - minX) + (maxY - minY)*(maxY - minY))/2
         };
 
         // We need to worry about resizing, but only if resizing could possibly change whether the hitbox is in or out of bounds
         // Thus if the upper-left corner is out of bounds, then there's no need to recheck on resize
         if (minX >= 0 && minY >= 0) {
             this._checkBounds = function() {
-                if (this._cbr === null && this._w < maxX || this._h < maxY ){
+                if (this._cbr === null && this._w < maxX || this._h < maxY) {
                    this._cbr = cbr;
                    this._calculateMBR();
                 } else if (this._cbr) {
@@ -186,13 +181,11 @@ Crafty.c("Collision", {
             this._calculateMBR();
             return true;
         }
-
     },
 
     // The default behavior is to match the hitbox to the entity.
     // This function will change the hitbox when a "Resize" event triggers.
     _resizeMap: function (e) {
-
         var dx, dy, rot = this.rotation * DEG_TO_RAD,
             points = this.map.points;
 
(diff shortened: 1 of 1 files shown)


src/spatial/collision.js
```diff
@@ -332,16 +325,16 @@ Crafty.c("Collision", {
      * @see .checkHits
      * @see .hit
      */
-    onHit: function (comp, callback, callbackOff) {
+    onHit: function (component, hit, noHit) {
```
mucaho: I actually liked the previous parameter names more, as the `-Off` part reflected their purpose better - function to be called after collision resolved.
Some suggestions: `callbackOn/-Off`, `hitOn/-Off`, etc.

src/spatial/collision.js
```diff
@@ -353,7 +346,7 @@ Crafty.c("Collision", {
     /**@
```
mucaho: maybe change to `/*@`, so that this internal method not included in public API?

mucaho:
Looks good, have couple of suggestions.

src/spatial/collision.js
Aliance: I agree, but PHPStorm highlights if function argument and phpdoc `@param` are different. May be change it in phpdoc?

src/spatial/collision.js
Aliance: I thought It was included on purpose, but if It was made by mistake I agree that removing this method from public API is good practise. That solve?

src/spatial/collision.js
mucaho: Let's update the documentation then accordingly

src/spatial/collision.js
mucaho: Yes, please change it to `/*@`, it should not be visible

src/spatial/collision.js
```diff
@@ -78,17 +75,15 @@ Crafty.c("Collision", {
      *
      * @see Crafty.polygon
      */
-    collision: function (poly) {
+    collision: function (polygon) {
```
Aliance, line 78: I renamed this argument like it described [here](https://github.com/craftyjs/Crafty/blob/develop/src/spatial/collision.js#L48). You don't mind?

src/spatial/collision.js
```diff
@@ -124,14 +119,13 @@ Crafty.c("Collision", {
         }
 
         // Finally, assign the hitbox, and attach it to the "Collision" entity
-        this.map = poly;
+        this.map = polygon;
         this.attach(this.map);
         this.map.shift(this._x, this._y);
```
Aliance, line 124: Is it okay what shifting is made after attaching and `NewHitbox` event triggering?

src/spatial/collision.js
```diff
@@ -266,7 +258,7 @@ Crafty.c("Collision", {
      *
      * @see 2D
      */
-    hit: function (comp) {
+    hit: function (component) {
```
Aliance, line 261: Renamed as [there](https://github.com/craftyjs/Crafty/blob/develop/src/spatial/collision.js#L235)

src/spatial/collision.js
```diff
@@ -289,11 +281,11 @@ Crafty.c("Collision", {
             id = obj[0];
 
             //check if not added to hash and that actually intersects
-            if (!dupes[id] && this[0] !== id && obj.__c[comp] && overlap(oarea, area))
+            if (!dupes[id] && this[0] !== id && obj.__c[component] && overlap(oarea, area))
                 dupes[id] = obj;
         }
 
-        for (key in dupes) {
+        for (key in dupes) if (dupes.hasOwnProperty(key)) {
```
Aliance: I added `hasOwnProperty` for each for/in, just in case. If you think it's no need for that - just tell me, i'll revert.

src/spatial/collision.js
Aliance: The coments above was assosiated with `_createCollisionHandler` method, I don't know why github renders them here.

src/spatial/collision.js
```diff
@@ -350,38 +342,25 @@ Crafty.c("Collision", {
...
-     * checks for collisions.
-     * @param collisionData - Collision data object used to track collisions with
-     * the specified component.
-     *
-     * This is a helper method for creating collisions handlers set up by
-     * `.checkHits(...)`. Do not call this directly.
-     *
-     * @see .checkHits
-     */
+    // This is a helper method for creating collisions handlers set up by checkHits method.
```
Aliance: I decided to remove all jsdoc after your [comment](https://github.com/craftyjs/Crafty/pull/949/files#r35801524) just to make it clear to other people.

src/spatial/collision.js
```diff
@@ -483,35 +462,35 @@ Crafty.c("Collision", {
      * ~~~
      */
     ignoreHits: function () {
-      var components = arguments;
-      var i = 0;
-      var collisionData;
+        var components = arguments;
```
Aliance: I made a couple of code align changes (from 2 to 4 spaces), if It's not suits you - please tell me, I'll revert it back.

Aliance:
@mucaho I have made the second commit, tell me your opinion on it - if it's not ok, i'll fix if; and if it is okay - i'll rebase 2 commits into one.

src/spatial/collision.js
mucaho, line 78: Yes, this is good.

src/spatial/collision.js
mucaho, line 124: Yes, I'm pretty sure it doesn't matter.

src/spatial/collision.js
mucaho: This is very essential, I'm glad you asked.
This is good practice when developing JavaScript in general, but Crafty tries to avoid any unnecessary statements because of performance. It is even more essential here, because this method gets executed very often - probably every 20ms.
In this case we created the array / object ourselves, so we can be 100% sure there can't be any other properties there (unless some external code overwrites standard prototypes).
Revert please.

src/spatial/collision.js
mucaho: I would leave it as it was and just change the `/**@` to `/*`. This small change will prevent it from being in public API documentation, while still being nicely documented for other developers as before.

src/spatial/collision.js
Aliance: Ok, I agree.

src/spatial/collision.js
```diff
@@ -349,39 +341,31 @@ Crafty.c("Collision", {
...
-
-    /**@
-     * #._createCollisionHandler
-     * @comp Collision
-     * @sign public void .checkHits(String component, Object collisionData)
-     * @param component - The name of the component for which this handler
-     * checks for collisions.
-     * @param collisionData - Collision data object used to track collisions with
-     * the specified component.
+    /**
```
Aliance, line 344: I've change `/**@` to `/**` (removed the `@`) not `/*` because of IDE highlighting, as I can see, you documentation gets only `/**@`?

And I deleted `#._createCollisionHandler` and `@comp Collision` 'cause of its uselessness in _private_ methods, also moved the description to the top of jsdoc.

mucaho:
This PR inspired me to open #951. We will hopefully fix code styles automatically soon, so I suggest you don't bother too much about indentation, as there are ~25 more files that need these fixes too :)

Aliance:
Ok, I see. Do I need to revert indentation changes?

mucaho:
Sorry for the long wait, I have been away from my working PC.
Did a final review, added some documentation code highlighting and merged with a6acb0df73a6fc51c62f415c8c2ed24ebd19d26b.

Issue #954, opened 2015-08-17
_clamp increases bound values when using scale()


gopeter:
When `_clamp` is called and `scale()` is set, the bounds increases by every call of `_clamp` because the values are multiplicated by its reference. That's how `console.log(bound.max.x)` looks like when the player moves:

![bildschirmfoto 2015-08-17 um 10 56 07](https://cloud.githubusercontent.com/assets/1242960/9301509/4cac72da-44cf-11e5-89bd-b69d1342b5c8.png)

Making a deep copy of the bound object solves the problem. Because I don't know how to make a deep copy correctly, I can't provide a pull request. I fixed it very dirty:

```
var b = this.bounds || Crafty.map.boundaries();
var bound = JSON.parse(JSON.stringify(b));

bound.max.x *= this._scale;
bound.min.x *= this._scale;
bound.max.y *= this._scale;
bound.min.y *= this._scale;
```

Just to let you know :)

Another problem I've encountered is the calculation of the the viewport when using `follow()` and `scale()` together. The camera doesn't stop on the bottom and right border when reaching the bound because the `scale` factor isn't included in the calculation, so I fixed it like this:

`if (Crafty.viewport.y < -bound.max.y + Crafty.viewport.height)` becomes `if (Crafty.viewport.y < (-bound.max.y + Crafty.viewport.height) / this._scale)` (and so on).

You can check out a demo here: https://debd.com/clients/oetinger/kk-game-pirates-viewport-pirates/test.html

If you are going to the very bottom you'll see that the camera does't stop correctly and always shows the player in the middle of the screen even though the game border/bound is already reached.

gopeter:
Oh, `Crafty.clone()` does the job! :smiley: 

I'll make a pull request.

mpetrovic:
shouldn't bounds increase with scale, since now the absolute playing area
has changed?

gopeter:
The bounds increases every time the player moves. Bounds should increase only once per scale (I updated my issue, sorry for confusing).

starwed:
Fixed in #955

PR #955, opened 2015-08-17
Fixed _clamp function for scaled viewports

gopeter:
This is a pull request for #954.


gopeter:
Looking at the docs, the code actually should be like this PR:

```
* `x` between `(-Crafty.viewport._x)` and `(-Crafty.viewport._x + (Crafty.viewport._width / Crafty.viewport._scale))`
* `y` between `(-Crafty.viewport._y)` and `(-Crafty.viewport._y + (Crafty.viewport._height / Crafty.viewport._scale))`
```

starwed:
Looks good to me.    I think our current set of tests specifically disable the clamp effect, but we should probably add some additional tests to catch issues like this in the future.

PR #958, opened 2015-09-19
Remove event callback object when empty

### LINKED ISSUE #957: Deleting empty event handlers?
I've been working with a team on an isometric Crafty game for a few months now, and we've got a map of ~36000 tiles. These tiles all have the following binding on initialization:

``` javascript
this.one('EnterFrame', function() {
        this.base.z = this._z;
});
```

. . . because they don't have their z value when they are initialized, but I digress.

With base Crafty, this is incredibly slow. It's slow because when Crafty goes to check on all of the EnterFrame callbacks, [here](https://github.com/craftyjs/Crafty/blob/develop/src/core/core.js#L1635): 

``` javascript
trigger: function (event, data) {
        // (To learn how the handlers object works, see inline comment at Crafty.bind)
        var hdl = handlers[event] || (handlers[event] = {}),
            h, i, l, callbacks, context;
        //loop over every object bound
        for (h in hdl) {

            // Check whether h needs to be processed
            if (!hdl.hasOwnProperty(h)) continue;
            callbacks = hdl[h];
            if (!callbacks || callbacks.length === 0) continue; // <----- HERE
```

It loops through every single tile every single frame, continuing each time. This takes our game's framerate down to ~10 fps, whereas if we modify this line to be this:

``` javascript
if (!callbacks || callbacks.length === 0) {
    delete handlers[event][h];
    continue;
}
```

We get up to the default max of 50 fps no problem. Instead of looping through all of these entities, who have no intention of binding to EnterFrame again, every single frame, it loops through them once and then runs with no issues (so far).

Is there a reason Crafty doesn't remove handlers after they are unbound, or doesn't remove empty handlers? It seems like it could drastically improve performance if you have a lot of entities that run .one bindings. Then again, I don't know if there are any major consequences to doing it.

starwed:
When there are no callbacks left attached to a system or entity, that array should be removed from the global handlers object.

Failing to do so can cause an unnecessary performance hit when a large number of extant entities have previously bound to that event.

Fixes #957.

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -1063,6 +1063,11 @@ Crafty._callbackMethods = {
                     callbacks.splice(i, 1);
                     i--;
                     l--;
+                    // Delete callbacks object if there are no remaining bound events
+                    if (callbacks.length === 0) {
+                        delete this._callbacks[event];
+                        delete handlers[event][this[0]];
+                    }
                 }
             } else {
                 callbacks[i].call(this, data);


starwed:
In this patch I remove the object when deleting the last event from it.

Another strategy would be to remove it when iterating through the global object and finding an empty callback array.

(The logic of removing handlers is in general pretty convoluted, due to the dangers of mutating the callback array while looping through it.)

PR #960, opened 2015-09-27
Handle restart more gracefully

starwed:
This fixes a bundle of bugs related to calling `Craft.stop()` and then `Crafty.init()`.
- Clear global bound events on stop (regression caused by the new callback methods)
- Add a `Crafty._preBind` method for internal use -- anything registered here will be bound when `Crafty.init()` is called (unless we did a soft stop).  This works around a problem where we bound some events only once, when crafty.js was loaded.  (Currently used in inputs.js and viewport.js)
- Make sure that "CraftyStop" is triggered _before_ we remove all event handlers!  (This was messing up cleanup of dom event handlers, which in turn messed up registering new ones.)
- Delete references to previously created graphics layers
- Have graphics layers set their initial values in their init functions

Some of these are really work arounds for how we initialize graphics layers, the viewport, etc.  I think if we switch these to using the new systems idiom we can remove a lot of the changes above.  (Definitely the "prebind" hack.)

### DIFF
--- a/src/core/extensions.js
+++ b/src/core/extensions.js
@@ -177,22 +177,18 @@ module.exports = {
 
         //save anonymous function to be able to remove
         var afn = function (e) {
-            e = e || window.event;
-
-            if (typeof callback === 'function') {
-                callback.call(ctx, e);
-            }
+            callback.call(ctx, e);
         },
             id = ctx[0] || "";
 
-        if (!this._events[id + obj + type + callback]) this._events[id + obj + type + callback] = afn;
-        else return;
-
-        if (obj.attachEvent) { //IE
-            obj.attachEvent('on' + type, afn);
-        } else { //Everyone else
-            obj.addEventListener(type, afn, false);
+        if (!this._events[id + obj + type + callback]) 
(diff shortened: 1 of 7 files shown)


src/core/extensions.js
```diff
@@ -177,22 +177,18 @@ module.exports = {
...
-
-        if (obj.attachEvent) { //IE
-            obj.attachEvent('on' + type, afn);
-        } else { //Everyone else
-            obj.addEventListener(type, afn, false);
+        if (!this._events[id + obj + type + callback]) 
+            this._events[id + obj + type + callback] = afn;
+        else  {
+            return;
         }
```
starwed, line 188: I also simplified the way we attached/removed event listeners -- they were workarounds for IE8 and below, which we no longer support.

starwed:
Not sure of the best way to add tests for this -- we could stop/start Crafty at the beginning of the test suite, but then there's the possibility of missing bugs that only exist on the first call to `init`.    (I did check this locally, and while there were test failures on the develop branch, this PR seems to fix them.)

(I think ideally we'd run all tests both before and after restarting Crafty.)

PR #964, opened 2015-10-05
now tweens can be paused

airza:
I needed to be able to pause tweens so that people in the background would stop moving when dialogue boxes came up.  So, here we are.


airza:
the most powerful of all commits

starwed:
Looks good, merging.  Low risk, so I'll probably pull it into the 0.7 branch.

PR #967, opened 2015-10-16
Fixed docs generation and added more examples.

jimbuck:
There was a whitespace issue regarding documentation generation. Once that was adjusted API documentation generation is working fine.

I've also added a few examples to `Crafty.extend` and `Crafty.clone`. Extend works differently than most libraries, so it is good to have a simple example. And clone just didn't have one, so I made a very basic one for that.


mucaho:
Thanks for the contribution!
Doc additions look good. Although I don't know coffee the change is intuitive. Merging.
EDIT: see also [coffeescript multi-line conditionals](http://thepolymathlab.com/multiple-line-if-statement-in-coffeescript)

Issue #914, opened 2015-05-27
Release 0.7.0


starwed:
Ok, it's been far too long since a release, so lets start preparing for one!  There have been some pretty large changes, so let's call this 0.7.

I just merged the develop branch into the testing branch -- no problems there, at least.

What issues are there blocking the release?
- [x] The introduction of the Motion component had pretty far reaching effects.  There are a couple of related issues, so we should probably address them before an actual release.
- [x] Doing the doc revamp revealed some issues with the api docs.  There are several broken links, some borked formatting, and some naming problems.
- [x] Various new components/methods should have their docs checked; previously it was really annoying to check the generated api docs, but the new grunt task in the develop branch should make that easier.  (`grunt view-api`).
- ~~There have been some recent changes to make crafty more useful in node; that probably needs some testing~~ Punting on this
- ~~Figure out the bower situation~~ Punting on this

Anything else big?

kevinsimper:
Sound good!

mucaho:
WebGL is also big feature.
Will the iso improvements (backward compatible?, airza still working on it) and  #902 be also included? Don't have much time currently, haven't had chance to look at iso improvements and finalize #902 .

starwed:
> Will the iso improvements (backward compatible?, airza still working on it) and #902 be also included? 
- iso: At least some of them -- I think the large changes should probably wait til the next release.  (Sounds like it might take a bit of time to settle on an api for the iso stuff.)
- 902: I guess this is mostly up to you and kevin -- I haven't tried using Crafty that way, so I don't have a good feel for what the current situation is.

starwed:
Current status:
- The doc issues are mostly taken care of in #918, although since the Motion-related components might still be in flux I haven't looked at their docs very closely.
- I'm going to try to build a very simple platformer using the new motion-related components, to get to grips with how they work.
- I don't think there are any WebGL related issues that block the release -- there are some places on the main site that should mention that as a feature, though.

starwed:
We should sort out what we're going to do with Bower.  (I think the general agreement is to use the Crafty-Distro repo for that, but I don't know the specifics of what we need to do to enable that.)

starwed:
@mucaho  @kevinsimper Ok, so the remaining big issues seem to be dealing with bower and node installation... any feedback on what the status is for them?

mucaho:
**Node status:** The majority of stuff is implemented in #902.
[...]

PR #972, opened 2015-11-13
Improve release workflow

mucaho:
Collection of improvements that will reduce human error when releasing
crafty for download, npm and bower.
Grunt task 'release' demands a 'crafty_version' command line parameter
which will update the version in the relevant files.
Additional meta improvements.

Fixes [future bower release inconsistencies](https://github.com/craftyjs/Crafty/issues/914#issuecomment-156398827).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 /crafty.js
-/crafty-min.js
 build/api/*
 build/api.json
 build/parseNodes.js
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,16 +1,13 @@
 require("coffee-script");
-var fs = require('fs');
 var open = require("open");
 
 module.exports = function (grunt) {
-    var pkg = grunt.file.readJSON('package.json');
-    var version = pkg.version;
-    var banner =    '/**\n' +
-                    ' * <%= pkg.name %> <%= pkg.version %>\n' +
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -1,5 +1,4 @@
-/crafty.js
-/crafty-min.js
+dist/
```
starwed: Don't we need to include this for the files to actually be accessible to bower?

The whole root issue is that (contrary to standard practices) bower _requires_ you to include the release artifacts in the repository  If we don't want to do that, the alternative is to have a separate repo for distribution.

mucaho: Hmm good point. If we don't ignore the release artifacts, we will have to check every PR so that users don't accidentally commit their release artifacts (because of security reasons). If we do ignore the release artifacts, I believe you can still add them to the commit (by `git add dist/crafty.js`), but care has to be taken not to forget adding them on release. What do you suggest?

mucaho:
Ok, modified it so that the `dev` build still builds to `crafty.js` (which is still git-ignored), and `release` build builds to `dist/crafty.js` and `dist/crafty-min.js` (which is not git-ignored).

mucaho:
Merging, as default workflow stays same, and on release we will have updated artifacts in `/dist` folder.

PR #973, opened 2015-11-15
Prevent gravity tunneling

mucaho:
Currently a falling object may fall through the floor if the gravity constant is big or if the deltaTime between frames is large.
This commit fixes this by searching for collisions in the whole area a falling entity might have skipped between frames.
This implementation is as (in)precise as the current implementation in regards to choosing the ground entity among multiple ground candidates that are close by.

Follow-up of #944 .
See [example](https://github.com/craftyjs/demos/blob/master/devDemos/example_tunneling.html).
Added [tunneling entry](https://github.com/craftyjs/Crafty/wiki/Crafty-FAQ-%28draft%29#why-are-my-bullets-passing-through-other-entities-without-registering-hits) to FAQ.


starwed:
Would this behave oddly if you were also moving very quickly on the horizontal axis?

The sort of collision detection Crafty uses generally has to just accept that very fast-moving objects can pass through other small objects; disallowing that comes with a performance penalty, because you're searching a larger space.  

It might be that ground detection is a special case where it makes sense to do these sorts of checks?  I'm not too sure...

mucaho:
> Would this behave oddly if you were also moving very quickly on the horizontal axis?

Good thinking. Usually you don't move that quickly on the horizontal axis, so it hasn't been reported as an issue yet (unlike the vertical case). I will update the commit to also consider widening horizontal search space.

> _this change_ comes with a performance penalty, because you're searching a larger space

Yes, you do search a larger space in 1% of the corner cases, where you reach huge velocities; even then the search space shouldn't increase that much. In 99% of cases the search space will remain as it was before this commit (specifically `ccdY` will be zero [here](https://github.com/craftyjs/Crafty/pull/973/files#diff-453060f510aa3958ad510d399bb817c2R1037)). I like to think of it as an fail-safe mechanism with zero amortized cost :)

mucaho:
Edits:
Ok, updated the code so that it also increases the horizontal search space.
Extracted CCD bounding box to `Motion.ccdbr` (in future this bounding box could be persisted to a field if multiple parties will query this new bounding box).
Made the anti-tunneling measure optional via `preventGroundTunneling`.
Added & moved multiple tests.

Merging optional feature.

Issue #950, opened 2015-07-30
Regression tests for visuals / display / graphics


mucaho:
So we added a lot of graphics features and don't have a way to test them currently.
We could remedy this by introducing a new set of testing facilities - visual based regression tests.

**How does it work?**
Basically you have a tool which takes screenshots of web pages (or specific elements of the page), saves those that you approved as a baseline and then each time you run the tests it checks if the new screenshots it takes match the baseline screenshots within a certain error margin.

**How could we integrate it into our existing workflow?**
Preliminary research has shown two potential candidates: [Gemini](https://github.com/gemini-testing/gemini) and [WebdriverIO](http://webdriver.io/).
WebdriverIO appealed more to me from my first impressions, so I will document the proposed process with that tool a bit here:
- there is a [webdriver-grunt](http://webdriver.io/guide/plugins/grunt-webdriver.html) task which we can use to run our tests
- the tests are actually run by a WebdriverIO plugin, namely [WebdriverCSS](http://webdriver.io/guide/plugins/webdrivercss.html); you tell it the `url`, which `HTMLElement` and whether to record baseline screenshots or test against them; tests can pass / fail just like in `QUnit`
- the cool thing is these tests can be run locally in a `phantomjs` browser, but also on different browsers on the Sauce Labs server (see #927). That means we get instant feedback if all[1] of Crafty's features run on all browsers we intend to support!
- there is a [grunt-css-test](https://www.npmjs.com/package/grunt-css-test) node module that could speed up the process; also see [grunt-sauce-driver](https://www.npmjs.com/package/grunt-sauce-driver) to manually run CSS tests

**What has to change if we want to proceed with this?**
Start making tests for `DOM`, `Canvas`, `WebGL`, `Viewport`, etc.. . Look at the webpages generated by those tests and record them as a baseline if they look as expected.

Footnote [1]: All features for which we wrote tests.

PR #974, opened 2015-11-15
Use system objects instead of hardcoded references for graphics layers

starwed:
Currently references to the canvas/dom/webgl graphics layers have to be hardcoded in some places.

This PR takes the current layer objects and defines them as systems using `Crafty.s`.  This means
- Events can be bound directly to the graphics layer, instead of going through global Crafty events
- Cleanup can be done by each layer when the system is destroyed
- The layers can now be recreated anew every time `Crafty.init()` is called

The new design should more easily allow for multiple layers of each type in a future PR, although there's an additional issue that has to be resolved first.  (Refactoring how entities attach themselves to a particular graphics layer so that it's easier to add/remove them.)

_e:_ I should add that I tested that all types of components still rendered with these changes.


mucaho:
Pretty big change, highlighting the significance of visual regression tests (#950) :)
How do users access [`Crafty.canvasLayer`](http://craftyjs.com/api/Crafty-canvasLayer.html) methods (and of other layers) after this change?
Would you kindly explain what the purpose of the last commit is? Aren't systems unique and initialized from scratch once destroyed or are there properties still available after destroying system?

starwed:
>  Is the WebGL div somewhere located in the DOM tree so we can document that?

I believe it's mentioned in the docs for WebGL, but we probably just want to add a section in the non-api documentation that discusses layers.  Since my immediate goal is to completely change how they work, I feel like it'd be best to hold-off on improving the documentation until all the changes have gone through.

> After you destroy systems, do they get reinitialized when requested again?

Currently no, you'd have to explicitly declare them again.  For built-in systems like the ones this PR adds, it becomes Crafty's responsibility to redeclare them when you call `Crafty.init()`.

> Would you kindly explain what the purpose of the last commit is? Aren't systems unique and initialized from scratch once destroyed or are there properties still available after destroying system?

It's the same "shared object" problem mentioned in the [docs for components](http://craftyjs.com/documentation/components.html) -- if you instantiate two systems from the same base object, and that base object has an array or object defined on it, then it'll be shared by both systems.  The immediate issue was actually destroying/recreating a system using stop/init, but it would also matter in the future when you might create multiple layers of the same type.

mucaho:
Ok, feel free to merge when ready.

starwed:
Merging this in.

PR #976, opened 2015-11-21
Entity name accessor

mucaho:
Currently the entity name can only be set, but not queried using public methods.
EDIT: Initially I introduced property-like access (`.name`), however since this may break user code (since it's such a common property name) and it won't be used as often as e.g. `.x`, I decided against it and propose to use `.getName()` instead.


mucaho:
Merging convenience functionality

PR #978, opened 2015-11-23
Unregister webGL entity on destroy.

### LINKED ISSUE #977: WebGL "Number of entities exceeds maximum limit." not clear
There is no clear cause what triggers the `"Number of entities exceeds maximum limit."`.

After looking in the code, https://github.com/craftyjs/Crafty/blob/develop/src/graphics/webgl.js#L68

I learned that this limit is somehow 1024.

But when my game halts with this error, the console returns the following:

`Crafty('*').length => 127`
`Crafty('WebGL').length => 90`

How is this limit reached?

I do create entities which I destroy later on. Does destroying a WebGL entitiy clear it from this limit?

matthijsgroen:
Fixes #977 

Too bad I could not write a test for it though, no webGL in phantomJS.

### DIFF
--- a/src/graphics/sprite.js
+++ b/src/graphics/sprite.js
@@ -198,10 +198,6 @@ Crafty.c("Sprite", {
 
     remove: function(){
         this.unbind("Draw", this._drawSprite);
-        // Webgl components need to be removed from their gl program
-        if (this.program) {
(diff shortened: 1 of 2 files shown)


starwed:
Thanks, this looks good to me -- no idea why this wasn't implemented as part of WebGL.

PR #984, opened 2015-11-27
Updates for most recent versions of node+packages

starwed:
I tried updating to the most recent versions of node and all the various packages.

I ran into some sort of issue with `node-qunit`, where it complained about being unable to find `lib/log.js`.  I don't know if this is a problem with my install, or the package itself.  `grunt-node-qunit` looks to be about 3 years old, so that might be the root of the issue.

I'd assume we should try to resolve that issue one way or the other before merging this.

(_e:_ Forgot to bump the version of .travis.yml.)


mucaho:
I would change node version to latest LTS, that's `4.2.2`.
`node-qunit` is a fantastic node module - I'm currently implementing visual regression tests (#950, current progress see [webdriver branch](https://github.com/mucaho/Crafty/tree/webdriver), see [crafty distro gh-pages branch](https://github.com/craftyjs/Crafty-Distro/tree/gh-pages) for preliminary screenshots) and I'm heavily relying on it.

starwed:
> I would change node version to latest LTS, that's 4.2.2.
> node-qunit is a fantastic node module - I'm currently implementing visual regression tests (#950, current progress see webdriver branch, see crafty distro gh-pages branch for preliminary screenshots) and I'm heavily relying on it.

Downgraded to 4.2.2, and that seemed to solve the issue with `grunt-node-qunit`.  I suspect that the issue is in the grunt wrapper, since it hasn't been upgraded in so long.  As long as we rely on it, it'll block us from upgrading node.  (That said, I bet it'd be pretty easy to fix/fork the grunt wrapper.)

starwed:
I see that you actually just had a PR merged into that `grunt-node-qunit`, so good to know the maintainer is still around.

starwed:
@mucaho  Not sure if it's related to the changes in this PR, but there seems to be some issue where the sauce-labs task has hung.

```
29 Nov 21:25:55 - Starting secure remote tunnel VM...

29 Nov 21:25:57 - Secure remote tunnel VM provisioned.

29 Nov 21:25:57 - Tunnel ID: 1cacd50dd811488b8ccef31541e74aee

29 Nov 21:25:58 - Secure remote tunnel VM is now: new

29 Nov 21:27:17 - Error bringing up tunnel VM.

29 Nov 21:27:17 - Sauce Connect could not establish a connection.
```

mucaho:
> As long as we rely on it, it'll block us from upgrading node ...
> ... good to know the maintainer is still around.

Yeah let's cross that bridge when we come to it :)

> there seems to be some issue where the sauce-labs task has hung

Happened to me too on occasion, I think their servers get overloaded from time to time, I usually fixed it by triggering a rebuild on Travis' website. (It also doesn't help that I have been stressing their servers a bit lately until I can get the configuration fine-tuned).

mucaho:
Ah yes, would you try adding `^` in front of every dependency, so we can depend on "semantically similar" versions.

starwed:
> Ah yes, would you try adding ^ in front of every dependency, so we can depend on "semantically similar" versions.

Sure, I was lazy and just used [npm-check-updates](https://www.npmjs.com/package/npm-check-updates) to update package.json -- I think it kept the prefixing the same, but obviously we should be consistent unless there are specific issues.

starwed:
Added the `^` qualifier to each dependency.  Looks like the open-sauce test didn't run, though?

mucaho:
The open-sauce tests don't run on PRs, they just run after something is merged into a Crafty repo branch (limitation described in PR).

However, I don't fully grasp why the Open-Sauce connect plugin (that one that gave you the error) started and failed in the first place in your previous build (the required open sauce credentials should only be available when running a build for craftyjs/Crafty).

Either way, I think it's best to change it so that the cloud tests are performed when it matters - only on the `testing` branch before release. We have tested it now that Crafty's JS side works on different browser, they just add too much delay and unpredictability for the day-to-day builds. This will be in the upcoming PR I'm finishing.

starwed:
Going to go ahead and merge this.

Issue #753, opened 2014-03-09
2D component initialization should not overwrite pre-existing properties


sbyrnes321:
I think the following code should work correctly;

```
this.x = 50;
this.w = 100;
this.requires('2D');
```

or

```
Crafty.c('koala', {
    w: 100,
    h: 100,
    init: function () { this.requires('2D'); }
});
Crafty.e('koala');
```

Right now neither would work because 2D component initialization overwrites `this._x` and similar properties with 0.

Is it possible to fix this? (Sorry I haven't checked yet myself.)

This issue has come up several times -- see #546 and https://groups.google.com/forum/#!topic/craftyjs/WKyxkURMJVo for example

mucaho:
There are two possibilities I can think of:
- remove [2D's direct object property values](https://github.com/craftyjs/Crafty/blob/develop/src/2D.js#L78-L171) and do `this._x = this._x || 0` for all properties in 2D's `init` method
- `requires` uses `addComponent` under the hood, which uses `extend` under the hood: `extend` copies all object keys and their values from the component to the entity -> restrict copying of keys that already exist in entity (might not work in the general case?)

sbyrnes321:
Well, you would do `this._x = this.x || this._x || 0` probably, because if you set `this.x` beforehand, it's not a setter yet. But sure, something like that would probably work. :-)

kevinsimper:
> remove 2D's direct object property values and do this._x = this._x || 0 for all properties in 2D's init method

This seems like the best solution

> requires uses addComponent under the hood, which uses extend under the hood: extend copies all object keys and their values from the component to the entity -> restrict copying of keys that already exist in entity (might not work in the general case?)

I tried something like it in #713 but it was not that great of a idea, since there were a lof of exeptions and some depend on overwritting properties (mostly functions) in a kind of OOP.

PR #985, opened 2015-11-29
Separate graphics from 2d

starwed:
For several reasons, it would make sense to separate the methods and properties on 2D that are only necessary when the entity is to be displayed.

This PR is a quick test of that idea, creating a "Graphics" component that steals several current 2D methods/properties.  Some open questions:
- Should `.z` (and related properties/methods) be on graphics, or 2D?  It's used primarily for determining how overlapping elements get rendered, but also for screen-input events.  I noticed that even the headless tests reference this, which makes me wonder if it should stay on the 2D component as something fundamental to dealing with any 2D entity, not just those which are being rendered.
- The flip/unflip methods are currently graphics only -- they don't affect anything like the collision hitboxes of an entity.  Should that continue to be the case, or should we promote them to proper 2D methods that handle such cases?
- The existing `_attr` function on 2D is a bit weird, since it pretty much special cases different sets of properties.  We could insert all that logic separately into each setter, although that would definitely result in some duplication of code.
- Maybe we should generalize the code that sets up properties? Maybe a `properties` field on component objects which Crafty would use to setup setters/getters/etc upon installing the component?

### DIFF
--- a/src/graphics/renderable.js
+++ b/src/graphics/renderable.js
@@ -1,0 +1,128 @@
+var Crafty = require('../core/core.js');
+
+
+/**@
+ * #Renderable
+ * @category Graphics
+ * Component for any entity that has a position on the stage.
+ * @trigger Invalidate - when the entity needs to be redrawn
+ */
+Crafty.c("Renderable", {
+
+
+    /**@
+     * #.alpha
+     * @comp Renderable
+     * Transparency of an entity. Must be a decimal value between 0.0 being fully transparent to 1.0 being fully opaque.
+     */
+    _alpha: 1.0,
+
+    /**@
+     * #.visible
+     * @comp Renderable
+     * If the entity is visible or not. Accepts a true or false value.
+     * Can be used for optimization by setting an entities visibility to false when not needed to be drawn.
+     *
+     * The entity will still exist and can be collided with but just won't be drawn.
+     */
+    _visible: true,
+
+
+    
+    _setterRenderable: function(name, value) {
+        if (this[name] === value) {
+            return;
+        }
+
+        //everything will assume the value
+        this[name] = value;
+
+        // flag for redraw
+        this.trigger("Invalidate");
+    },
+
+    // Setup all the properties that we need to define
+    _graphics_property_definitions: {
+        alpha: {
+            set: function (v) {
+                this._setterRenderable('_alpha', v);
+            },
+            get: function () {
+                return this._alpha;
+            },
+            configurable: true,
+            enumerable: true
+        },
+        _alpha: {enumerable:false},
+
+        visible: {
+            set: function (v) {
+                this._setterRenderable('_visible', v);
+            },
+            get: function () {
+                return this._visible;
+            },
+            configurable: true,
+            enumerable: true
+        },
+        _visible: {enumerable:false}
+
+    },
+
+    _defineRenderableProperites: function () {
+        for (var prop in this._graphics_property_definitions){
+            Object.defineProperty(this, prop, this._graphics_property_definitions[prop]);
+        }
+    },
+
+    init: function () {
+        // create setters and getters that associate properties such as alpha/_alpha
+        this._defineRenderableProperites();
+    },
+
+    /**@
+     * #.flip
+     * @comp Renderable
+     * @trigger Invalidate - when the entity has flipped
+     * @sign public this .flip(String dir)
+     * @param dir - Flip direction
+     *
+     * Flip entity on passed direction
+     *
+     * @example
+     * ~~~
+     * this.flip("X")
+     * ~~~
+     */
+    flip: function (dir) {
+        dir = dir || "X";
+        if (!this["_flip" + dir]) {
+            this["_flip" + dir] = true;
+            this.trigger("Invalidate");
+        }
+        return this;
+    },
+
+    /**@
+     * #.unflip
+     * @comp Renderable
+     * @trigger Invalidate - when the entity has unflipped
+     * @sign public this .unflip(String dir)
+     * @param dir - Unflip direction
+     *
+     * Unflip entity on passed direction (if it's flipped)
+     *
+     * @example
+     * ~~~
+     * this.unflip("X")
+     * ~~~
+     */
+    unflip: function (dir) {
+        dir = dir || "X";
+        if (this["_flip" + dir]) {
+            this["_flip" + dir] = false;
+            this.trigger("Invalidate");
+        }
+        return this;
+    }
+});
--- a/src/graphics/webgl.js
+++ b/src/graphics/webgl.js
@@ -30,6 +30,7 @@ Crafty.c("WebGL", {
      * The webgl context this entity will be rendered to.
      */
     init: function () {
+        this.requires("Renderable");
         var webgl = this.webgl = Crafty.s("WebGLLayer");
         var gl = webgl.context;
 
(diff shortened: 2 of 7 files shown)


starwed:
(Just to be clear, this isn't ready to merge -- I opened to see if anyone had some insight into those questions!)

mucaho:
> Should .z (and related properties/methods) be on graphics, or 2D?

Voting for 2D. I have plans for isometric where I will need access to `.z` in order to perform display agnostic calculations (Jumping & Gravity using `z` instead of `y` axis).

> The flip/unflip methods are currently graphics only

Do they flip the rotation origin also? Do they flip position of attached children? I think they should, but for now they just serve for display, Can be moved later on.

> The existing _attr function on 2D is a bit weird, since it pretty much special cases different sets of properties. We could insert all that logic separately into each setter, although that would definitely result in some duplication of code.

Well, setting `.x` now goes through 2 levels of indirection, having only one setter indirection could speed things up a bit.
While you are at it, please take a look at all the `Move`, `Rotated` and similar events. They are very heterogeneous (some trigger with oldValue, some with delta, etc..) and could be streamlined.

Edit: Additionally I would suggest extracting all broadphase related stuff out of 2D (mbr, adding & removing from broad phase map) into a new Broadphase component.

> Maybe we should generalize the code that sets up properties? Maybe a properties field on component objects which Crafty would use to setup setters/getters/etc upon installing the component?

~~Yes, but should the initialization logic be bound to specific components?
What about introducing a new concept of `Property` or `Data` (or w/e name) that gets defined similar to components. Each 'Property' takes care of setting up its setters (+ perhaps input sanitation), getters and change event firing.
Components then glue the data flow from properties to other properties together (either every frame or on events), similar to how Entities model control flow by connecting components.
Additionally, no matter which component requires a property first, it would be set up the same. That could maybe fix the long outstanding issue #753.~~
Edit: Scrap that last idea, it could lead to bad coupling. Having one component that manages / initializes data properties is good. If another component needs that data property, it will have to require the managing component that takes care of updating the property. (This reminds me of one writer - many reader solution to concurrency).

starwed:
Updated:
- Keep `z` in `2D`
- Use "Renderable" instead of "Graphics" for the component name

I noticed that the Touch tests rely on entities being visible, and thus require the Renderable component.  I guess that makes sense.

src/graphics/renderable.js
```diff
@@ -0,0 +1,131 @@
...
+
+/**@
+ * #Renderable
+ * @category Graphics
+ * Component for any entity that has a position on the stage.
+ * @trigger Move - when the entity has moved - { _x:Number, _y:Number, _w:Number, _h:Number } - Old position
+ * @trigger Invalidate - when the entity needs to be redrawn
+ * @trigger Rotate - when the entity is rotated - { cos:Number, sin:Number, deg:Number, rad:Number, o: {x:Number, y:Number}}
+ * @trigger Reorder - when the entity's z index has changed
+ */
```
mucaho, line 9: This component only triggers Invalidate events

src/spatial/2d.js
```diff
@@ -929,6 +834,14 @@ Crafty.c("2D", {
             }
             this.trigger("Move", old);
 
+        } else if (name === '_z') {
+            var intValue = value <<0;
+            value = value==intValue ? intValue : intValue+1;
+            this._globalZ = value*100000+this[0]; //magic number 10^5 is the max num of entities
+            this[name] = value;
+            this.trigger("Reorder");
+            this.trigger("Invalidate");
```
mucaho: duplicate Invalidate trigger, will get triggered down below

src/graphics/webgl.js
```diff
@@ -103,7 +104,7 @@ Crafty.c("WebGL", {
         co.x = coord[0] + (x || 0);
         co.y = coord[1] + (y || 0);
         co.w = w || coord[2];
-        co.h = h || coord[3];
+        co.h = h || coord[3];
```
mucaho: trailing whitespace

mucaho:
Have some suggestions, other than that, good first step in right direction!

src/spatial/2d.js
```diff
@@ -81,6 +82,14 @@ Crafty.c("2D", {
      * @see ._attr
      */
     _z: 0,
+
+    /**@
+     * #._globalZ
+     * @comp 2D
+     * When two entities overlap, the one with the larger `_globalZ` will be on top of the other.
+     */
+    _globalZ: null,
```
mucaho, line 91: `_globalZ` is only needed for display stuff, can this be moved to `Renderable` too?

src/spatial/2d.js
```diff
@@ -929,6 +834,14 @@ Crafty.c("2D", {
...
 
+        } else if (name === '_z') {
+            var intValue = value <<0;
+            value = value==intValue ? intValue : intValue+1;
+            this._globalZ = value*100000+this[0]; //magic number 10^5 is the max num of entities
+            this[name] = value;
+            this.trigger("Reorder");
+            this.trigger("Invalidate");
+            return;
         }
```
starwed, line 843: Nope, because we return early in this branch.

IIRC that's because we want "Reorder" to trigger after the value has been set.

src/spatial/2d.js
starwed, line 91: It's updated by changing `z`, though, so I'm inclined to leave them together.

Also, I think some of the other components _should_ be using `_globalZ`.  Mouse dispatch compares on `z` and uses id as a tie breaker, but if it used `_globalZ` there'd never be any ties.

src/graphics/renderable.js
starwed, line 9: Good catch.

src/spatial/2d.js
mucaho, line 843: Ok, but then you can leave out both `this.trigger("Invalidate");` and the following `return;`, right? The code would then stay the same as in previous implementation, or is there another reason why you introduced those 2 statements?

src/spatial/2d.js
mucaho, line 91: Other alternative would be to listen for `Reorder` events and update `_globalZ` in Renderable accordingly (`z` won't be changed that often in most games). Mouse (similarly to Touch) also implicitly depends on Renderable component (as their common called `findClosestEntityByComponent` function checks for `_visible`).
Don't be fooled by its name, it's not really a spatial property, but rather an imaginary property describing render order (and could be renamed as such).

However, now that I've outlined relevant points, I will be happy either way you decide :)

src/spatial/2d.js
starwed, line 843: Hmm, I'm actually confused now about where this code came from -- maybe I did a bad merge or rebase at some point?

Anyway, removing the trigger and the return.

src/spatial/2d.js
starwed, line 91: That's true, we could move that to be triggered by Reorder.  I'll leave that out of this set of changes, though.

starwed:
Update with fixes to whitespace, event block, and duplicate invalidate logic.

mucaho:
Ok, merging.

Follow-up PR could include:
- ~~I think some of the other components should be using _globalZ. Mouse dispatch compares on z and uses id as a tie breaker, but if it used _globalZ there'd never be any ties.~~ (added via c46afac9322c282aa775298efd3149eda93797b2)
- Move _globalZ to Renderable and be triggered by Reorder (possibly also renamed to _renderOrder)

PR #986, opened 2015-11-30
Add WebDriver tests

### LINKED ISSUE #950: Regression tests for visuals / display / graphics

mucaho:
Explanation moved to [Webdriver test wiki](https://github.com/craftyjs/Crafty/wiki/Webdriver-tests).
## Miscellaneous
- This PR adds example test templates, not all tests covering every functionality, some tests will come in next PR. Additional tests can always be added over time.
- fixes #950
- streamlines Gruntfile

What do you think? This should give us much needed confidence when doing changes to graphics stuff.


mucaho:
Ok, updated the documentation above.
I'm pretty satisfied with the current capabilities for now (after waging battles with the differences in each browser's test capabilities).
@starwed Do you have some comments?

starwed:
Thanks for taking the time to not just implement this, but also thoroughly explain it!

Reading through your comments it sounds good to me.  Just to check my understanding: this particular PR implements the general framework as well as two specific DOM tests?

One theoretical possibility would be to compare DOM, Canvas, and WebGL rendering of the same scene against the same screenshot.  I have a [quasi-complicated scene](http://ludonomicon.com/crafty/triple-render.html) I've been using to check a variety of functionality in the three renderers simultaneously.  (It uses all the transformations, animation, and transparency, as well as Image, Sprite, and Color entities.)

mucaho:
Yes, this PR implements general framework + some example tests (they can be DOM, Canvas, WebGL, doesn't matter).

> One theoretical possibility would be to compare DOM, Canvas, and WebGL rendering of the same scene against the same screenshot.

Yep, you can do that by stuffing all 3 rendering stripes into the same test (as you did in your example scene), and then calling `assertResemble(bounds)` for each of these stripes' bounds.
I have deliberately removed the need to specify URLs & paths inside the tests, but the available commands can be extended to accept test page urls and screenshot file paths, if you want.

> I have a quasi-complicated scene I've been using to check a variety of functionality in the three renderers simultaneously

This is golden! Have to shrink it a bit, but can then be included as a test as explained above. I will also need to split the animation into discrete steps, as we can only compare screenshots and not videos.

mucaho:
Merging then, more tests incoming

Issue #989, opened 2015-12-09
Guard against cross domain security error


mucaho:
Trying to include Crafty v0.7.0 in jsfiddle, jsbin and requirebin fails with `SecurityError: The operation is insecure.` (at least in firefox).
This has probably something to do with browsers preventing scripts (that originate from other domains) from reading browser data. [Same-origin policy](https://en.wikipedia.org/wiki/Same-origin_policy).

v0.7.0 changed localStorage to automatically detect the kind of localStorage (browser / node) to use upon script load - see [this line](https://github.com/craftyjs/Crafty/blob/0.7.0/src/core/storage.js#L2).
One possible solution would be to wrap that statement in `try / catch` block or execute it when storage is accessed for the first time.

mucaho:
fixed in #991

PR #991, opened 2015-12-13
Catch the case when accessing localStorage throws a security exception

starwed:
Hopefully fixes issue #989.

I had to require Crafty here for the logging functionality.

### DIFF
--- a/src/core/storage.js
+++ b/src/core/storage.js
@@ -1,5 +1,11 @@
+var Crafty = require('../core/core.js');
 
-var storage = (typeof window !== "undefined" && window.localStorage) || (new require('node-localstorage').LocalStorage('./localStorage'));
+try {
+  var storage = (typeof window !== "undefined" && window.localStorage) || (new require('node-localstorage').LocalStorage('./localStorage'));
+} catch(e) {
+  var storage = null;
+}
+
 
 /**@
  * #Storage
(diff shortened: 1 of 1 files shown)


mucaho:
Looks good, except for possible complications caused by requiring:
Requiring the `core` module is fine in the browser (as there is only one Crafty instance at any given time), but not that straightforward in node (as there can be multiple Crafty instances running on the server).

Remember, we always delete the cache and require a "fresh" instance of all required modules in node.
Now, in this case it's fine, as we "refreshed" the `core` module before (see [crafty_headless#L4](https://github.com/craftyjs/Crafty/blob/0.7.0/src/crafty_headless.js#L4)). If some day in the future (for w/e reason) the `storage` module gets required before the `core` module, we have a problem.

Is there a "safer" way to get access to the logging feature?

starwed:
> Is there a "safer" way to get access to the logging feature?

We could pull in the logging feature as a separate module, the same way storage is contstructed.

mucaho:
> We could pull in the logging feature as a separate module, the same way storage is contstructed.

Sure, I don't quite follow but sounds good :)

starwed:
> If some day in the future (for w/e reason) the storage module gets required before the core module, we have a problem.

I think it's probably ok for now to simply insist that core is required first, and merge this PR.

I believe we decided in the original discussion that the long term fix would be to make the various modules return functions that accept the `Crafty` object, thus sidestepping the issue of caching completely.

mucaho:
Yes, it's ok. I completely forgot we use this exact code in other src files :)

PR #994, opened 2015-12-19
Add webdriver tests

mucaho:
Continuation of #986:
- Allow tests in subdirectories (better structure of test files)
- Enable reuse of expected screenshot across test files (as discussed in previous PR)
- Allow dynamically created test pages (reduces boilerplate code)
- Enable bi-directional signal and wait between browser and webdriver
- Fix exclude paths in local webdriver tests
- Add additional tests:
  - Color

TODO
- Image
- Sprite
- SpriteAnimation
- Text
- PixelArt
- Manipulated viewport (scale, animations, ...)
- Particles (with 0 randomness)
- starwed's visual integration test


tests/webdriver/color/color-common.js
```diff
@@ -0,0 +1,40 @@
...
+      box.color('rgba(0, 0, 255, 0.5)');
+      signalBarrier('transparent');
+    }
+
+    waitBarrier('initial', initial);
+    waitBarrier('opaque', opaque);
+    waitBarrier('transparent', transparent);
+  };
+
+  return testScript.toString().replace('$renderMethod', renderMethod);
```
starwed, line 26: Is this a standard pattern for these types of tests?  Seems like it should be possible to adjust this dynamically without doing serializing to a script with text replacement, but I might not understand the limitations of the framework.

starwed:
Good work, it'll be nice to have tests that can assert things about our render layers!

tests/webdriver/color/color-common.js
mucaho, line 26: Yeah, the problem is the whole function must not have any external dependencies, as it will be serialized as-is into a test html page.  
Alternatively, you could introduce a `renderMethod` parameter, then you would have to wrap the function into an anonymous function call that also supplies the `renderMethod` parameter. I just thought this was more readable, as you have to manually prepend and append the necessary parentheses for that to work.

matthijsgroen:
What is needed to merge in? The test suite does not have to be complete imho. If this is in, we could all contribute to render tests. Maybe add a task list?

mucaho:
I just wanted to make sure the test framework is as stable as possible, so I wanted to add some more tests to make sure that I haven't missed any major requirements.
Need to also add documentation to the test framework.
But both could be in follow-up PR I guess?

matthijsgroen:
That way others (like me) could also contribute
Op za 27 feb. 2016 om 11:43 schreef mucaho notifications@github.com

mucaho:
Ok, I looked over it again and merged, thanks for the patience.  
Documentation is now on [a wiki page](https://github.com/craftyjs/Crafty/wiki/Webdriver-tests) and will be updated accordingly.

PR #995, opened 2015-12-23
Refactor how components are associated with specific draw layers

starwed:
Abstract out the concept of attaching and detaching from a layer into a set of methods on Renderable.

Also abstract out the concept of a renderable entity being marked as dirty.

The point of all this is to make it possible to have multiple layers of the same type -- for instance, a background and foreground canvas layer.

This probably needs some more polish and thought in terms of how components are added and removed from entities (I think we have some existing bugs/issues there that could be cleaned up as part of this work) but I thought I'd open the PR so folk could give feedback.

### DIFF
--- a/src/graphics/canvas-layer.js
+++ b/src/graphics/canvas-layer.js
@@ -10,6 +10,7 @@ var Crafty = require('../core/core.js');
  * Mostly contains private methods to draw entities on a canvas element.
  */
 Crafty.canvasLayerObject = {
+    type: "Canvas",
     _dirtyRects: [],
     _changedObjs: [],
     layerCount: 0,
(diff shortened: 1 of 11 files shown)


starwed:
> This probably needs some more polish and thought in terms of how components are added and removed from entities (I think we have some existing bugs/issues there that could be cleaned up as part of this work) but I thought I'd open the PR so folk could give feedback.

Fixed an issue or two, and now removing one layer and adding another will work pretty smoothly.

starwed:
The logic behind this PR in more detail:
- Each graphics layer object now has three standard methods: `attach(entity)`, `detach(entity)`, and `dirty(entity)`.  A layer will only render objects which are attached to it, and uses the `dirty` method to track which objects have changed and need redrawing.  Some of the logic that used to live in the Canvas/DOM/WebGL components now lives in the layer objects directly.
- The "Renderable" component includes methods that wrap the way an entity will call those methods on the layers, triggering relevant events and so on.
- An entity gets attached to a layer when the relevant component is added.
- The components responsible for specific types of rendering (Color, Sprite, etc) might need to do setup related to the specific layer that will draw them.  (Defining WebGL shaders or creating patterns in a canvas context.)  This code will run when the entity is attached to a layer, and also if it is already attached to the layer when the component is added.  (This is handled by each component separately -- although this results in a bit of duplicate logic, it didn't quite seem worth adding an additional layer of abstraction.)

This arrangement makes it so that you can easily remove and add components like "WebGL" or "DOM" from an entity to control how it is rendered.  It's also an important step in allowing multiple instances of the same type of graphics layer.

kevinsimper:
Looks good, I think this a nice addidtion as well as for performance, since things does not needed to be drawn so much!

I have wondered, is the `.dirty` tactic still viable comparing to the extra code that is needed, comparing to just drawing is all again like?

starwed:
> Looks good, I think this a nice addidtion as well as for performance, since things does not needed to be drawn so much!

Putting background or completely static entities in a separate background layer should definitely have a perf. improvement.

> I have wondered, is the .dirty tactic still viable comparing to the extra code that is needed, comparing to just drawing is all again like?

The plan is that when you create a layer, you'll be able to specify some options: one for canvas would be the threshold for redrawing the whole scene.  (Another super useful one will allow a a layer to respond to viewport changes differently --that would let you have static UI elements, or a background with parralax effects.)

starwed:
> I have wondered, is the .dirty tactic still viable comparing to the extra code that is needed, comparing to just drawing is all again like?

Oh, and there are some nice optimizations you can make with the WebGL  layer: right now we have a huge array representing the data for all WebGL entities, and on rendering we copy it into a background buffer.  By tracking the first/last dirty elements, we could avoid copying static entities at the beginning/end of the array (i.e. static background/foreground entities.)

PR #998, opened 2015-12-28
Pause delay and particles

### LINKED ISSUE #997: Pausing the game
I would like to see some improvements on how to pause the game:

Right now, there is a Crafty.pause() and Crafty.unpause()
They however are not pausing the game, but pausing the entire engine.

For now I bind stuff to a 'GameLoop' event instead of 'EnterFrame' and have a simple inbetween mechanic

`Crafty.bind('EnterFrame', function(fd) { if (Game.paused) return; Crafty.trigger('GameLoop', fd); });`

This way I can pause the gameplay, and keep the engine running (to display a pause menu, etc)

For Tweens, this works nicely because they can be paused and unpaused.
However, some Crafty components can not be stopped this way:
- Delay
- Particles

Would it be nice to have `pause()` and `unpause()` methods for these as well?

( I will try to submit pull requests for these if these are sane improvements )

matthijsgroen:
Next to pausing tweens, this pull request adds the ability to pause delays and particles.

Fixes #997

### DIFF
--- a/src/core/time.js
+++ b/src/core/time.js
@@ -9,7 +9,9 @@
 module.exports = {
     init: function () {
         this._delays = [];
+        this._delaysPaused = false;
         this.bind("EnterFrame", function (frameData) {
+            if (this._delaysPaused) return;
             var index = this._delays.length;
             while (--index >= 0) {
                 var item = this._delays[index];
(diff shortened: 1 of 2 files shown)


matthijsgroen:
@mucaho or @starwed  does this require any approval before merge? Something I missed?

starwed:
In an airport right now, but a couple of quick notes:

There are several copy/paste errors in the docs. (You can use `grunt view-api` to check how they look like live)

I'd prefer the flag booleans to be declared as explicitly false.

matthijsgroen:
@starwed fixed!

matthijsgroen:
Any updates?

kevinsimper:
It should be rebased into 1 commit :)

matthijsgroen:
@kevinsimper done. Dunno why having it in one commit should stall progress of the merge. because we are afraid the project will have too many commits? It's weird policy.

kevinsimper:
It makes it easier to git blame later and it is a pretty normal policy :)

But looks good to me!

starwed:
> Dunno why having it in one commit should stall progress of the merge. because we are afraid the project will have too many commits? It's weird policy.

It makes it easier to trace the history of code changes, both when an individual developer is reading through the history manually, and when using automatic tools like `git bisect`.  (I've used bisect before to track down a hard-to-find visual regression in Crafty.)

A general rule of thumb is that each commit should be complete (adds functionality that works, and passes tests), and that any small fixes or changes to earlier work in a PR should be squashed down to the commit that they affect.

(At my current job we use a different work flow, and it really makes me appreciate the benefits of having a relatively clean git history like we do in Crafty!)

All that said, I often just squash the commits down locally before merging a PR, since I don't necessarily want to increase the barrier of entry to contributing.  :)  In this case I wanted to wait a couple of days before merging, in case another maintainer had comments.

PR #1012, opened 2016-03-03
Various improvements

### LINKED ISSUE #956: Changing scenes with viewport cause DOM appendings
Look at [this small demo](http://jsfiddle.net/Aliance/dxzLxt17/3/). There are:
- black movable character;
- green grass, on which the character moves;
- red buildings, visiting which will cause another scene enter (scene with bigger viewport)

On this another scene there is only one button, clicking on it = returning again to main game scene. Try to do this manipulations a few times. Then inspect DOM, you can find the `container` element (this is my main stage element), and a couple of divs inside it. Is it okay if entering a new scene will cause a new div to be append to main `container`?

May be I do something wrong?

mucaho:
Details in commit log.
Things to keep in mind:
- Checking whether an entity is destroyed (e.g. by checking a stale reference to it) is not directly possible as far as I know. I used `Crafty(ent[0]) === ent` for that, which seems unnecessarily slow. Maybe add a direct functionality (e.g. `ent._destroyed`) for that?
- Currently every entity (even if it does not have the `Keyboard` component) gets informed about key events. Maybe change this in future to be in line with how `Mouse` works? However, selecting entities by component seems slow atm (implementation iterates over all entities and checks for component). We could maybe improve performance by managing a map of each component name to a bag of entityIds it is added to?
  e.g.

``` js
compEntities = {
   "2D": {
       entity1Id: true,
       entityNId: true
   },
   "DOM": { ... }
};
```

### DIFF
--- a/src/controls/controls.js
+++ b/src/controls/controls.js
@@ -111,6 +111,7 @@ Crafty.c("Draggable", {
         this._oldY = this._y;
     },
 
+    //Note: the code is not tested with zoom, etc., that may distort the direction between the viewport and the coordinate on the canvas.
     _drag: function(e) {
         if (this._dir) {
             if (this._dir.x !== 0 || this._dir.y !== 0) {
--- a/src/controls/controls.js
+++ b/src/controls/controls.js
@@ -129,10 +129,15 @@ Crafty.c("Draggable", {
 /**@
  * #Multiway
  * @category Controls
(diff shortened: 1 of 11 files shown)


starwed:
I didn't look at the doc changes in detail, but everything else looks good to me!

starwed:
> We could maybe improve performance by managing a map of each component name to a bag of entityIds it is added to?

I think this would be a good perf improvement.  In a component base system, getting a list of all the entities with a particular component is a pretty fundamental operation, and tracking this should actually be pretty cheap!

mucaho:
Alrighty, thanks for taking a look at it, merging since these changes affect docs mostly.

PR #1017, opened 2016-03-07
Refactor how components are associated with specific draw layers

starwed:
- Abstract out the concept of attaching and detaching from a layer into a set of methods on Renderable.
- Also abstract out the concept of a renderable entity being marked as dirty.
- Minimal support for creating custom layers

This is an enhanced version of #995.  AFAICT the code is now fairly solid, but there are probably some documentation changes that need to be made.

### DIFF
--- a/src/graphics/canvas-layer.js
+++ b/src/graphics/canvas-layer.js
@@ -10,6 +10,7 @@ var Crafty = require('../core/core.js');
  * Mostly contains private methods to draw entities on a canvas element.
  */
 Crafty.canvasLayerObject = {
+    type: "Canvas",
     _dirtyRects: [],
     _changedObjs: [],
     layerCount: 0,
@@ -21,16 +22,46 @@ Crafty.canvasLayerObject = {
     },
 
     /**@
-     * #.add
+     * #.dirty
      * @comp CanvasLayer
-     * @sign public .add(ent)
+     * @sign public .dirty(ent)
      * @param ent - The entity to add
      *
-     * Add an entity to the list of Canvas objects to draw
+     * Add an entity to the list of Canvas objects that need redrawing
      */
-    add: function add(ent) {
+    dirty: function dirty(ent) {
         this._changedObjs.push(ent);
     },
+    
+    /**@
+     * #.attach
+     * @comp CanvasLayer
+     * @sign public .attach(ent)
+     * @param ent - The entity to add
+     *
+     * Sets the entity's draw context to this layer
+     */
+    attach: function attach(ent) {
+        ent._drawContext = this.context;
+        //increment the number of canvas objs
+        this.layerCount++;
+    },
+    
+    /**@
+     * #.detach
+     * @comp CanvasLayer
+     * @sign public .detach(ent)
+     * @param ent - The entity to add
+     *
+     * Removes an entity to the list of Canvas objects to draw
+     */
+    detach: function detach(ent) {
+        this.dirty(ent);
+        ent._drawContext = null;
+        //decrement the number of canvas objs
+        this.layerCount--;
+    },
+    
 
     /**@
      * #.context
@@ -70,8 +101,6 @@ Crafty.canvasLayerObject = {
         c.style.position = 'absolute';
         c.style.left = "0px";
         c.style.top = "0px";
-
-        var canvas = Crafty.s("Canvas");
 
         Crafty.stage.elem.appendChild(c);
         this.context = c.getContext('2d');
@@ -141,7 +170,7 @@ Crafty.canvasLayerObject = {
      */
     _drawDirty: function () {
 
-        var i, j, q, rect,len, obj, ent,
+        var i, j, q, rect,len, obj,
             changed = this._changedObjs,
             l = changed.length,
             dirty = this._dirtyRects,
@@ -193,7 +222,7 @@ Crafty.canvasLayerObject = {
             for (j = 0, len = q.length; j < len; ++j) {
                 obj = q[j];
 
-                if (dupes[obj[0]] || !obj._visible || !obj.__c.Canvas)
+                if (dupes[obj[0]] || !obj._visible || (obj._drawLayer !== this) )
                     continue;
                 dupes[obj[0]] = true;
                 objs.push(obj);
@@ -249,10 +278,12 @@ Crafty.canvasLayerObject = {
 
         //sort the objects by the global Z
         q.sort(this._sort);
+        //console.log("\n--\n" + q.length + " to draw for layer " + this.name);
         for (; i < l; i++) {
             current = q[i];
-            if (current._visible && current.__c.Canvas) {
-                current.draw();
+            if (current._visible && current._drawContext === this.context) {
+                //console.log(i + " drawn");
+                current.draw(this.context);
                 current._changed = false;
             }
         }
--- a/src/graphics/canvas.js
+++ b/src/graphics/canvas.js
@@ -22,33 +22,19 @@ Crafty.c("Canvas", {
 
     init: function () {
         this.requires("Renderable");
-        var canvasLayer = Crafty.s("CanvasLayer");
-
-        this._drawLayer = canvasLayer;
-        this._drawContext = canvasLayer.context;
-
-        //increment the amount of canvas objs
-        canvasLayer.layerCount++;
+        
         //Allocate an object to hold this components current region
         this.currentRect = {};
-        this._changed = true;
-        canvasLayer.add(this);
+        
+        // Add the default canvas layer if we aren't attached to a custom one
+        if (!this._customLayer){
+            this._attachToLayer( Crafty.s("CanvasLayer"));
+        }
+        
+    },
 
-        this.bind("Invalidate", function (e) {
-            //flag if changed
-            if (this._changed === false) {
-                this._changed = true;
-                canvasLayer.add(this);
-            }
-
-        });
-
-
-        this.bind("Remove", function () {
-            this._drawLayer.layerCount--;
-            this._changed = true;
-            this._drawLayer.add(this);
-        });
+    remove: function() {
+        this._detachFromLayer();
     },
 
     /**@
@@ -76,8 +62,6 @@ Crafty.c("Canvas", {
             w: 0,
             h: 0
         }
-
-
     },
 
     draw: function (ctx, x, y, w, h) {
@@ -97,8 +81,8 @@ Crafty.c("Canvas", {
         pos._h = (h || this._h);
 
 
-        context = ctx || this._drawContext;
-        coord = this.__coord || [0, 0, 0, 0];
+        var context = ctx || this._drawContext;
+        var coord = this.__coord || [0, 0, 0, 0];
         var co = this.drawVars.co;
         co.x = coord[0] + (x || 0);
         co.y = coord[1] + (y || 0);
--- a/src/graphics/dom-layer.js
+++ b/src/graphics/dom-layer.js
@@ -9,6 +9,7 @@ var Crafty = require('../core/core.js'),
  * Collection of mostly private methods to represent entities using the DOM.
  */
 Crafty.domLayerObject = {
+    type: "DOM",
     _changedObjs: [],
     _dirtyViewport: false,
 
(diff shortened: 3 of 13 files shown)


mucaho:
Reviewing this (actually your now closed PR) was next thing in my work queue :)
Couple of questions:
- I noticed you moved  `_changed` from `2D` to `Renderable`. However, `Invalidate` is still also triggered when changing any of the 2D properties. Is this intended? (That's the only place besides of the graphics components that it is triggered. `Renderable` could maybe listen to `Change` event (or multiple others) instead of triggering `Invalidate` in `2D`)?
- Why do we need set `ent._drawContext` in addition to `ent._drawLayer`? Can't the entities retrieve it from `ent._drawLayer`? Or better yet, why don't we use the context as the parameter for the `Draw` call? `Color` (and other graphics components) can then use this context to render.
- Would it make sense to create a method `clean` that serves as the opposite of `dirty` and sets changed to `false`?
- There is a lot of code repeated across rendering backend systems, rendering backend components, and specific graphics components. I wish there was a way to inherit that code, to avoid human error when changing something.
- Following are some code comments I made. A lot of them could be false negatives, just sharing some of my thoughts.

src/graphics/canvas-layer.js
```diff
@@ -21,16 +22,46 @@ Crafty.canvasLayerObject = {
...
+     * #.dirty
      * @comp CanvasLayer
-     * @sign public .add(ent)
+     * @sign public .dirty(ent)
      * @param ent - The entity to add
      *
      * Add an entity to the list of Canvas objects to draw
      */
-    add: function add(ent) {
+    dirty: function dirty(ent) {
```
mucaho, line 32: fix docs

src/graphics/canvas-layer.js
```diff
@@ -21,16 +22,46 @@ Crafty.canvasLayerObject = {
...
+    
+    /**@
+     * #.attach
+     * @comp CanvasLayer
+     * @sign public .attach(ent)
+     * @param ent - The entity to add
+     *
+     * Add an entity to the list of Canvas objects to draw
+     */
+    attach: function attach(ent) {
```
mucaho, line 44: fix docs

src/graphics/canvas-layer.js
```diff
@@ -21,16 +22,46 @@ Crafty.canvasLayerObject = {
...
+    
+    /**@
+     * #.detach
+     * @comp CanvasLayer
+     * @sign public .detach(ent)
+     * @param ent - The entity to add
+     *
+     * Add an entity to the list of Canvas objects to draw
+     */
+    detach: function detach(ent) {
```
mucaho, line 58: fix docs

src/graphics/canvas.js
```diff
@@ -22,39 +22,25 @@ Crafty.c("Canvas", {
...
-        });
+    remove: function() {
+        this._detachFromLayer();
     },
 
     /**@
      * #.draw
      * @comp Canvas
-     * @sign public this .draw([[Context ctx, ]Number x, Number y, Number w, Number h])
+     * @sign public this .draw([[Context ctx, ]Number x, Nwumber y, Number w, Number h])
```
mucaho: typo

src/graphics/layers.js
```diff
@@ -0,0 +1,21 @@
+var Crafty = require('../core/core.js');
+
+Crafty.extend({
+    createLayer: function createLayer(name, layerTemplate) {
```
mucaho, line 21: add doc (if you plan on expanding upon this then no need now)

src/graphics/canvas-layer.js
```diff
@@ -193,7 +222,7 @@ Crafty.canvasLayerObject = {
             for (j = 0, len = q.length; j < len; ++j) {
                 obj = q[j];
 
-                if (dupes[obj[0]] || !obj._visible || !obj.__c.Canvas)
+                if (dupes[obj[0]] || !obj._visible || (obj._drawLayer !== this) )
```
mucaho, line 225: Why is the check `obj._drawLayer !== this` needed here? Could a dirty entity be detached from current layer and attached to other layer before being drawn (and thus undirtied)?

src/graphics/canvas-layer.js
```diff
@@ -249,10 +278,12 @@ Crafty.canvasLayerObject = {
 
         //sort the objects by the global Z
         q.sort(this._sort);
+        //console.log("\n--\n" + q.length + " to draw for layer " + this.name);
         for (; i < l; i++) {
             current = q[i];
-            if (current._visible && current.__c.Canvas) {
-                current.draw();
+            if (current._visible && current._drawContext === this.context) {
```
mucaho, line 284: Same question as above for `current._drawContext === this.context`

src/graphics/dom-layer.js
```diff
@@ -113,17 +114,46 @@ Crafty.domLayerObject = {
...
+     * #.detach
+     * @comp DomLayer
+     * @sign public .detach(ent)
+     * @param ent - The entity to remove
+     *
+     * Removes an entity from the layer
+     */
+    detach: function detach(ent) {
+        this._div.removeChild(ent._element);
+    },
```
mucaho, line 155: missing `ent._drawContext = null;` and `this.dirty(ent);` like in CanvasLayer

src/graphics/dom.js
```diff
@@ -115,12 +100,15 @@ Crafty.c("DOM", {
...
+     * Will reattach the entity to the current draw layer
      */
     DOM: function (elem) {
         if (elem && elem.nodeType) {
-            this.undraw();
+            var layer = this._drawLayer;
+            this._detachFromLayer();
             this._element = elem;
-            this._element.style.position = 'absolute';
+            this.attachToLayer(layer);
```
mucaho: typo, should be `this._attachToLayer(layer)`

src/graphics/image.js
```diff
@@ -37,6 +37,7 @@ Crafty.c("Image", {
 
     init: function () {
         this.bind("Draw", this._drawImage);
+        this.bind("LayerAttached", this._setupImage);
```
mucaho, line 40: How is image setup if already attached to layer?

``` js
if (this._drawLayer) {
    this._setupImage(this._drawLayer);
}
```

maybe add this event like in `Color`:

``` js
events: {
   "LayerAttached": "_setupImage"
}
```

src/graphics/image.js
```diff
@@ -37,6 +37,7 @@ Crafty.c("Image", {
...
     },
 
     remove: function() {
```
mucaho, line 43: if not as `events` property, then also unbind `LayerAttached` listener below

src/graphics/image.js
```diff
@@ -88,25 +89,26 @@ Crafty.c("Image", {
             var self = this;
 
             this.img.onload = function () {
-                self._onImageLoad();
+                self._setupImage(self._drawLayer);
             };
         } else {
-            this._onImageLoad();
+            this._setupImage(this._drawLayer);
         }
```
mucaho, line 98: what if the `_drawLayer` isn't existing here yet when `_setupImage` is called?

``` js
Crafty.e("2D, Image, WebGL");
```

src/graphics/renderable.js
```diff
@@ -80,6 +79,40 @@ Crafty.c("Renderable", {
...
+    // Attach the entity to a layer to be rendered
+    _attachToLayer: function(layer) {
+        if (this._drawLayer) {
+            this._detachFromLayer();
+        }
+        this._drawLayer = layer;
+        layer.attach(this);
+        this.bind("Invalidate", this._invalidateRenderable);
+        this.trigger("LayerAttached", layer);
+        this.trigger("Invalidate");
```
mucaho, line 102: Can we change this, so that the entity gets dirtied when added to the layer, instead of triggering `Invalidate`? (to be consistent how the entity gets dirtied when detached from layer)

src/graphics/sprite.js
```diff
@@ -143,11 +143,7 @@ Crafty.extend({
             //set the width and height to the sprite size
             this.w = this.__coord[2];
             this.h = this.__coord[3];
-
-            if (this.has("WebGL")){
-                this._establishShader(this.__image, Crafty.defaultShader("Sprite"));
-                this.program.setTexture( this.webgl.makeTexture(this.__image, this.img, false) );
-            }
+            this._setupSpriteImage(this._drawLayer);
```
mucaho, line 147: what if `this._drawLayer` doesnt exist?
`Crafty.e("2D, Sprite, CustomWebGL");`

src/graphics/sprite.js
```diff
@@ -205,10 +201,20 @@ Crafty.c("Sprite", {
     init: function () {
         this.__trim = [0, 0, 0, 0];
         this.bind("Draw", this._drawSprite);
+        this.bind("LayerAttached", this._setupSpriteImage);
```
mucaho, line 205: shouldn't `_setupSpriteImage` be called if _drawLayer already exists here in addition?

src/graphics/webgl-layer.js
```diff
@@ -379,7 +377,7 @@ Crafty.webglLayerObject = {
         visible_gl.length = 0;
         for (i=0; i < l; i++) {
             current = q[i];
-            if (current._visible && current.__c.WebGL && current.program) {
+            if (current._visible && current.__c.WebGL && current.program && (current._drawLayer === this)) {
```
mucaho: same question as above why we check this new condition

src/graphics/webgl-layer.js
```diff
@@ -414,6 +411,47 @@ Crafty.webglLayerObject = {
...
+     *
+     * Removes an entity from the layer
+     */
+    detach: function detach(ent) {
+        // This could, like attach, be handled by components
+        // We instead handle it in a central place for now
+        if (ent.program) {
+            ent.program.unregisterEntity(this);
+        }
+    },
```
mucaho, line 455: `this.dirty(ent);` & `ent._drawContext = null;` possibly missing here

src/graphics/renderable.js
```diff
@@ -80,6 +79,40 @@ Crafty.c("Renderable", {
         this._defineRenderableProperites();
     },
 
+    // Renderable assumes that a draw layer has 3 important methods: attach, detach, and dirty
+
+    // Dirty the entity when it's invalidated
+    _invalidateRenderable: function() {
+        //flag if changed
+        if (this._changed === false) {
+                this._changed = true;
```
mucaho: fix indentation

src/graphics/canvas-layer.js
starwed, line 225: The same reason why we previously needed the check that the object has the canvas component -- when we search the map, every entity is returned, not just Canvas entities, and we do the filtering here.

At some point I could see adding configurable hashmaps as a layer option, which would prevent collision checking between layers which opt-in, but would probably allow for several performance improvements.

src/graphics/dom-layer.js
starwed, line 155: ~~The `_drawContext` should be set to null, but it doesn't actually need to be flagged as dirty.  (Canvas needs that because it has to manually redraw where the entity was removed, but removing the element from the parent suffices for DOM.)~~

Got a bit confused -- setting the draw context to null/etc happens in the Rendarable method `detachFromLayer.`  This method is for layer-specific cleanup.  (Canvas calls `dirty()` because that's necessary for proper cleanup, due how we handle partial redraws of the canvas.)

src/graphics/dom.js
starwed: Good catch, ~~I was probably seeing all elements in the same DOM layer during testing.~~

Oh, this was in the method that swaps out `_element`, not the initial setup.  I'd better test that it works as expected after the typo fix.

src/graphics/image.js
starwed, line 98: `_setupImage` is called from two places: when a layer is attached, and when the image is assigned.  If either step hasn't been completed, it'll simply return.  (The same logic is used for sprites).

I experimented with a few different designs, but this one seemed the most robust.

src/graphics/sprite.js
```diff
@@ -205,10 +201,20 @@ Crafty.c("Sprite", {
...
     },
 
     remove: function(){
         this.unbind("Draw", this._drawSprite);
+        this.unbind("LayerAttached", this._setupSpriteImage);
+    },
+    
+    _setupSpriteImage: function(layer) {
+        if (!this.__image || !this.img || !layer) return;
```
starwed, line 214: `_setupSpriteImage` and `_setupImage` do the same type of check to make sure the method is only run if both
- the entity is attached to a layer
- the resource to render has loaded

src/graphics/webgl-layer.js
starwed: Same answer, that our search for visible entities returns _all_ entities.  (Although it looks like there's still a now redundant check for the WebGL component here.)

src/graphics/canvas-layer.js
mucaho, line 225: One hashmap per layer and one dedicated hashmap for collision stuff would be the dream (I stumbled upon this "inconvenience" when separating isometric entities into their display part and their game-logic part).  
The only downside I see to this, is sometimes multiple entries (one for display and one for collision - both optional, not every entity needs collision) would have to be managed per entity, so that could have negative performance impact. However, the performance gain (by not considering irrelevant entries inside a layer) could maybe outweigh that.

starwed:
@mucaho Fixed the docs and typo.  I think I addressed all your questions above, but let me know if I missed one!

mucaho:
@starwed 
- Build is failing for some reason?
- Q: `Renderable` could maybe listen to `Change` event (or multiple others) instead of triggering `Invalidate` in `2D`?
- Nitpicks:
  - [Typo](https://github.com/craftyjs/Crafty/pull/1017#discussion_r55255112)
  - [Missing event unbind](https://github.com/craftyjs/Crafty/pull/1017#discussion_r55264648)
  - [Indentation](https://github.com/craftyjs/Crafty/pull/1017#discussion_r55268675)

Other than that, this is a nice next step for having separate drawing layers. If we add a documented functionality to change the z-index per layer, the performance gain should be significant.

starwed:
Fixed the remaining issues you caught.

> Build is failing for some reason?

I think that was because I'd accidentally pushed the branch to the main crafty repo, and then deleted it.  The test was trying to run against that branch but was unable to find it.  The failure disappeared when I repushed.

> `Renderable` could maybe listen to `Change` event (or multiple others) instead of triggering `Invalidate` in `2D`?

That kind of makes sense in principle, but I'd be a little worried about the overhead of adding a nested event on such a basic inner-loop style operation.  For now let's leave it as is, even though it means that 2D triggers events that aren't needed if running inside a non-rendering context.

starwed:
> If we add a documented functionality to change the z-index per layer, the performance gain should be significant.

I've got this (and control over how a layer responds to camera motion) working in a new branch.  :)

PR #1019, opened 2016-03-08
Add Sprite_sprite(tileName)

### LINKED ISSUE #1015: Request: Change sprite by its name in Sprite.sprite
Im not sure, maybe I am missing something but there seems to be no proper way to change an entities sprite / tile of a sprite (not sure, the naming isnt quite clear).  
Lets say we have a sprite(map):  

`Crafty.sprite(50, "imgs/mysprites.png", {dog:[0,0], cat:[0,1,3,1]});`  

So now we have `dog` and `cat` .  
Now I have an entity, animal. I initiate it as a cat.  

`var animal  = Crafty.e("2D, DOM, cat");`  

How would I change the sprite to be a dog?  
The only approach seems to be to remove the `cat` component.. and add a `dog` component? Thats not a proper approach I am sure. How would I do this?

mucaho:
[example usage](https://github.com/craftyjs/demos/blob/master/devDemos/example_sprite_name.html)

### DIFF
--- a/src/graphics/sprite.js
+++ b/src/graphics/sprite.js
@@ -126,6 +126,7 @@ Crafty.extend({
             this.requires("2D, Sprite");
             this.__trim = [0, 0, 0, 0];
             this.__image = url;
+            this.__map = map;
             this.__coord = [this.__coord[0], this.__coord[1], this.__coord[2], this.__coord[3]];
             this.__tile = tile;
             this.__tileh = tileh;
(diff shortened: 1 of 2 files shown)


mucaho:
No objections, thus merging.
As discussed in linked the issue, this introduces the same semantic _"inconsistency"_ as the already available ability to change a sprite component's appearance with new coordinates.

Issue #729, opened 2014-02-08
2D Raycast


mucaho:
Feature request from [forum](https://groups.google.com/forum/#!topic/craftyjs/Lyb2FsGyIU4)

> You'd want to examine how the HashMap code works -- it implements a spatial partition we use for broad phase collisions.  Any implementation of raycasting would probably use that as a starting point.  You could step through the grid sections in order, checking in each one for collisions.  That would definitely be more efficient than using a collision entity, since you could just stop once you found an intersection, and could use a test for ray intersection instead of SAT.

mucaho:
@grozen you are our collision expert :) do you have any additional ideas?

grozen:
@mucaho I'd hardly say that as I actually had very little to do with the actual collision detection algorithm. That being said, this sounds like a great idea if you or someone else are up to it.

Assuming this ends up as a new sort of component, it would be very simple to reuse the collision events that became available as part of #699, though obviously it would be prudent to define this as a new collision type (in addition to MBR and SAT that are currently what `.hit` may return).

mucaho:
Ah ok, I thought you were involved in writing the spatial hashmap.

I took the painful (but also performant) approach by stepping through the spatial grid sections iteratively. A preliminary version can be found on my [raycast branch](https://github.com/mucaho/Crafty/tree/raycast). This is entirely based on [this tutorial](http://www.flipcode.com/archives/Raytracing_Topics_Techniques-Part_4_Spatial_Subdivisions.shtml), which is indirectly linked from Crafty's HashMap code.

mucaho:
Update:

I have implemented the algorithm that finds the nearest bounding box or all bounding boxes that are intersected by the ray in the HashMap. Now I have to figure out how to efficiently get the actual intersections, since bounding boxes do not always correspond to the actual entity. The following questions arise:
- (1) We support only rectangular entities?  
  When rectangular entities are rotated, their bounding box is larger than the actual entity, thus we have to use [line segment intersection](http://stackoverflow.com/questions/563198/how-do-you-detect-where-two-line-segments-intersect) in order to determine if the ray intersects with the rotated entitiy and to retrieve actual intersection point(s).

![](http://i61.tinypic.com/16ly9zo.jpg)
- (2) Even if we support only rectangular entities, how will we determine if the ray hits entities that have transparent texture? The entitiy could be completely transparent, thus ray should not intersect with that entity at all

![](http://i61.tinypic.com/10pdbhv.jpg)

[...]

PR #1021, opened 2016-03-10
Add 2D raycast

mucaho:
**Development finished. Pending review.**
Fixes outstanding issue #729 .

Screenshot taken from playground example:
raycast

There is still lots to do, but I would appreciate a little input on the API design:
- ~~`Crafty.map.boundaries(findKeys)` was extended to find only hashmap cell keys if optional parameter is provided. Should this stay here (introduces some branching in code) or make an explicit method `Crafty.map.keyBoundaries()` (code duplication)?~~ Added a separate, private method `Crafty.map._keyBoundaries()`.
- `Crafty.map.iterateMap(origin, direction, entryCallback)` iterates the hashmap in the direction of the ray. Calls `entryCallback` for each hashmap entry (obj). If `entryCallback` returns `true`, aborts iteration immediately.
- ~~`Collision.approximateDistanceFromRay(origin, direction)` returns the distance to nearest corner of the entity's bounding rectangle (if there is an intersection).~~ Removed from implementation
- `Crafty.polygon.distanceFromRay(origin, direction)` returns the distance to the closest intersection with any of the polygon's segements (if there is at least one intersection).
- `Crafty.raycast(origin, direction, maxDistance, comp, sort)` uses the above described methods to find an intersection. Returns an array of raycast results, each raycast result looks like `{obj: obj, distance: distance, x: intersection_point_x, y: intersection_point_y}`.
  - If `maxDistance` is Infinity, finds all intersections (this is the default case). If its negative, finds only first intersection. If its a positive value, finds intersection up to that distance.
  - If `comp` string parameter is supplied, returns only entities that have the specified component
  - If `sort` boolean parameter is supplied and is set to `false`, then the results are **not** sorted by distance (opt-out, slight performance gain)
  - ~~`checkBroadphase` dictates whether an approximate broadphase intersection check (`approximateDistanceFromRay`) should be used to filter objects, before checking for exact intersection (with `distanceFromRay`)~~ removed from implementation
- `origin` & `direction` describe the ray. `direction` **must** be normalized. Should this be mentioned in the docs, or should input sanitation (and thus this normalization) be done?
- ~~`Crafty.raycast` returns an array of raycast results. Duplicates are weeded out, but the array is not sorted by distance. Should there be an additional parameter `sort` (similar to the `filter` parameter in [`Crafty.map.search`](http://craftyjs.com/api/Crafty-map.html#Crafty-map-search)) that optionally sorts the results (and has performance impact)?~~ Added

Notes:
- ~~I still have to test whether the approximate broadphase intersection check is worth it, and how many segments a polygon must have (increases complexity of the exact intersection check) to be worth it.~~ Removed from implementation
- ~~Collision (and raycast) logic needs the collision bounding rectangle (`this._cbr || this._mbr || this`). Rendering stuff needs the minimum bounding rectangle (`this._mbr || this`).  
  The considered bounding rectangle is hardcoded in the hashmap's functions (currently the collision one is used). This can possibly lead to (and possibly has already lead to; haven't fully tested) problems with drawing dirty regions on Canvas, when there exists a collision hitbox outside of the entity's natural bounds (`_cbr != null`).  
  Having [separate hashmaps](https://github.com/craftyjs/Crafty/pull/1017#discussion_r55499253) for collision stuff and for each layer would help in configuring each hashmap so that the correct property gets accessed for the bounding rectangle. A more immediate fix would be to provide an optional argument to each of the hashmaps methods, which dictates which bounding rectangle property to use.~~ Nope, render code just uses plain rectangles for searching the spatial map, so it doesn't matter.

### DIFF
--- a/playgrounds/2D/raycast/raycast.html
+++ b/playgrounds/2D/raycast/raycast.html
@@ -1,0 +1,28 @@
+<!DOCTYPE HTML>
+<html>
+	<head>
+		<script type="text/javascript" src="../../../crafty.js"></script>
+	</head>
+	<body>
+		<h1>Raycast playground:</h1>
+
+		<p>
+		<h3>Input:</h3>
+			<ul>
+			<li>Drag on the background to initiate a raycast (up to the line length).<br/></li>
+			<li>Drag &amp; drop entities to move them.<br/></li>
+			<li>Hold shift while dragging entities up/down to rotate them.<br/></li>
+		</ul>
+		</p>
+
+		<p>
+		<h3>Output:</h3>
+			<ul>
+			<li>First object hit by ray will be given red outline.<br/></li>
+			<li>For all objects hit (up to the line length) the intersection point will be marked with a purple dot.<br/></li>
+		</ul>
+		</p>
+		<div id="cr-stage" style="border: solid black 8px;"></div>
+		<script type="text/javascript" src="raycast.js"></script>
+	</body>
+</html>
--- a/playgrounds/2D/raycast/raycast.js
+++ b/playgrounds/2D/raycast/raycast.js
@@ -1,0 +1,165 @@
+Crafty.init(640, 640);
+Crafty.background('white');
+
+
+var startPos = {};
+Crafty.e('Background, 2D, DOM, MouseDrag, DebugPolygon')
+      .setName('Background')
+      .attr({w: 640, h: 640})
+      .debugStroke('red')
+      .bind('StartDrag', function(e) {
+        this.debugPolygon(undefined);
+        startPos = {x: e.realX, y: e.realY};
+      })
+      .bind('Dragging', function(e) {
+        this.debugPolygon({points: [startPos.x, startPos.y, e.realX, e.realY]});
+      })
+      .bind('StopDrag', function(e) {
+        raycast(startPos, {x: e.realX, y: e.realY});
+      });
+
+for (var i = 0; i < 10; ++i)
+  for (var j = 0; j < 10; ++j)
+    Crafty.e('Cell, 2D, DOM, Color')
+          .setName('Cell['+i+']['+j+']')
+          .attr({ x: i * 64, y: j * 64, w: 64, h: 64, z: -1})
+          .color((i + j) % 2 ? "rgba(0, 255, 0, 0.1)" : "rgba(0, 0, 255, 0.1)");
+
+Crafty.c("MouseMovable", {
+  required: "Mouse, Draggable",
+
+  events: {
+    "KeyDown": function(e) {
+      if (e.key === Crafty.keys.SHIFT) {
+        this.disableDrag();
+        this.uniqueBind("Dragging", this._dragMouseDrag);
+        this.uniqueBind("StopDrag", this._stopMouseDrag);
+      }
+    },
+    "KeyUp": function(e) {
+      if (e.key === Crafty.keys.SHIFT) {
+        this.enableDrag();
+        this.unbind("Dragging", this._dragMouseDrag);
+        this.unbind("StopDrag", this._stopMouseDrag);
+      }
+    }
+  },
+
+  _stopMouseDrag: function() {
+    delete this._lastMousePos;
+  },
+  _dragMouseDrag: function(e) {
+    if (this._lastMousePos) {
+      var d = new Crafty.math.Vector2D(
+        e.realX - this._lastMousePos.x,
+        e.realY - this._lastMousePos.y
+      );
+
+      this.rotation += d.dotProduct({x: 0, y: 1});
+    }
+    this._lastMousePos = {x: e.realX, y: e.realY};
+  }
+});
+
+var trapezoid = Crafty.e('Trapezoid, 2D, DOM, Color, Collision, SolidHitBox, MouseMovable')
+      .setName('Trapezoid')
+      .attr({w: 200, h: 100})
+      .origin('center')
+      .collision(new Crafty.polygon([50, 0, 0, 100, 200, 100, 150, 0]))
+      .color("rgba(0, 127, 127, 0.25)");
+
+var yellow = Crafty.e('Yellow, 2D, DOM, Collision, SolidHitBox, MouseMovable')
+      .setName('Yellow')
+      .attr({w: 100, h: 100})
+      .origin('center')
+      .collision(new Crafty.polygon([0, 0, 0, 100, 100, 100, 100, 0]));
+
+var parallelogram = Crafty.e('Parallelogram, 2D, DOM, Color, Collision, SolidHitBox, MouseMovable')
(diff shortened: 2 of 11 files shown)


starwed:
I guess finding the closest instance of a particular type of component will be something that's useful, which would require doing the check _inside_ the raycasting?  I wish there was a way to avoid that, though the only other option would to basically have the raycaster be an iterator of some sort, maintaining it's internal state and allowing you to grab the next intersection through a method call.

It might be worth thinking about how this would interact with some sort of compiled _asm.js_ collision engine.  I believe that the first approach would be tough (the compiled code couldn't interact with general crafty code, and so it would be tricky for it to inspect entities for the presence of components), while the latter approach would perhaps be more natural in such a context?

starwed:
> I still have to test whether the approximate broadphase intersection check is worth it, and how many segments a polygon must have (increases complexity of the exact intersection check) to be worth it.

I'd expect that it's not worth it -- if used properly (i.e. don't put 100s entities in the same cell) the spatial map is itself the broadphase check.  This is probably more true of raycasting than regular collisions, since SAT is more expensive than the simpler checks here.

starwed:
> Having separate hashmaps for collision stuff and for each layer would help in configuring each hashmap so that the correct property gets accessed for the bounding rectangle. A more immediate fix would be to provide an optional argument to each of the hashmaps methods, which dictates which bounding rectangle property to use.

I wonder if the best approach for Canvas redrawing would be to have it flag cells of the hashmap as dirty, rather than carefully keeping track of the bounding areas of individual entities.

starwed:
> Crafty.map.boundaries(findKeys) was extended to find only hashmap cell keys if optional parameter is provided. Should this stay here (introduces some branching in code) or make an explicit method Crafty.map.keyBoundaries() (code duplication)?

IIRC calling boundaries is actually pretty expensive, maybe not something we want to do every time raycast is called.  A possible optimization is caching them, and only recalculating the boundaries if the hashmap has been modified?  (i.e. have a dirty flag on the hashmap)

mucaho:
@starwed Thanks for the feedback!

> I guess finding the closest instance of a particular type of component will be something that's usefu

Added

> I'd expect that _the approximate intersection test_ is not worth it

Removed

> A possible optimization is caching _the hashmap boundaries_, and only recalculating the boundaries if the hashmap has been modified?

Added

---

What's still missing:
- [ ] follow-up PR: let Crafty.map.boundaries and Viewport.bounds return a flat object `minX, minY, maxX, maxY`
- [ ] follow-up PR: setting the `_cbr` doesn't update the entry in spatial map, until the entity is moved/rotated/resized
- [x] api documentation
- [x] split the single large test file into smaller ones
- ~~search the spatial map for the `_mbr` instead of the `_cbr` in the render code~~

If you're ok with it, I'd include the latter 2 points into a follow-up PR (with some additional proposals in regards to bounding rectangle handling).

I feel confident about the implementation now, so if you have time and are up for it, take a look at it, I've included extensive implementation-specific documentation.

mucaho:
> It might be worth thinking about how this would interact with some sort of compiled asm.js collision engine. I believe that the first approach would be tough (...), while the latter approach would perhaps be more natural in such a context?

I don't quite understand what the 1st approach and 2nd approach is :)

> I wonder if the best approach for Canvas redrawing would be to have it flag cells of the hashmap as dirty, rather than carefully keeping track of the bounding areas of individual entities. 

In an 'Asteroids' like game, where you have small entities scattered all around the scene, it might perhaps require redrawing all cells? Might work well in the general case though. Maybe some additional metrics / statistics could be considered here, like [heuristical cluster analysis](https://en.wikipedia.org/wiki/K-means_clustering#Complexity)? Might be overkill though :)

starwed:
> I don't quite understand what the 1st approach and 2nd approach is :)

No worries, I was just trying to think through how an underlying optimized algorithm would work, and probably wasn't very clear.  In the end the API the user calls would be unchanged.

> In an 'Asteroids' like game, where you have small entities scattered all around the scene, it might perhaps require redrawing all cells? Might work well in the general case though. 

Well, it would work the same way it does now -- if a large fraction of entities need redrawing, just do them all.

> search the spatial map for the _mbr instead of the _cbr in the render code

I might be forgetting how our existing code works, but what's the change here?

mucaho:
> > search the spatial map for the _mbr instead of the _cbr in the render code
> 
> I might be forgetting how our existing code works, but what's the change here?

Ah, I'm glad you asked, there is no problem at all. Render code calls `Crafty.map.search(rect)`, where `rect` is just a plain rectangle object, which does not have `_cbr` and `_mbr` properties.

Side-note:
Canvas clears some-dirty-entity-encompassing `rect`s and then proceeds to draw all overlapping entities. Note that it does not clear the overlapping entities that are partly inside and partly outside these `rect`s, thus these get drawn again. When drawing the same line multiple times in Canvas, I noticed it getting thicker and thicker. However, these artifacts may not apply here.

---

> > In an 'Asteroids' like game, where you have small entities scattered all around the scene, it might perhaps require redrawing all cells? Might work well in the general case though.
> 
> Well, it would work the same way it does now -- if a large fraction of entities need redrawing, just do them all.

True, I was just wondering about the edge case where percentage of dirty entities < 50% and their dimension << cellsize.

I like this proposal, as it could potentially improve performance for Canvas rendering. Drawing dirty entities requires merging their dirty rectangles, which sounds like an expensive operation. Drawing dirty cells could be done by just clearing each individual cell and drawing its entities.

starwed:
> True, I was just wondering about the edge case where percentage of dirty entities < 50% and their dimension << cellsize.

From what I understand, you always want your cellsize to be some small multiple of the typical entity size.  If you have lots of small entities moving around inside a single cell, collision performance will suffer because the broad-phase pass will leave too many candidates for the more expensive check.

We kind of support setting the cellsize before init, but it should probably be an actual option exposed through the api.

mucaho:
Ok, added the missing stuff (api docs, test file organization and some minor tweaks).

mucaho:
@starwed 
Do you want to take a look at this or should I go ahead and merge it?
We can still fine-tune it before release, if there are any issues.

starwed:
@mucaho  I haven't had time to fully look through it, but so long as it doesn't seem likely to affect performance I don't have any problem with landing it.  Like you say, we can always tune it up before release.  (And if we're concerned, add a warning to the docs that the api is new and possibly unstable.)

Feel free to merge once the conflicts are resolved.

mucaho:
There may be certainly some things that could get optimized.
One notable thing that could be added in future: It would be nice to have method which does multiple raycasts (in different directions, from same origin point) at once. That would remove the need to check if the origin point lies inside the map for each raycast, but rather do it only once.
I'm particularly proud that the implementation does not use a single `Math.sqrt` statement. It uses basic math operations instead. It's a marginal improvement though :)

PR #1024, opened 2016-04-17
Add options to systems & createLayers

starwed:
This adds support for options to `Crafty.s`, and uses that with the `createLayer` method to let you control the behavior of custom graphics layers.

The options added by this PR are:
- `z`, which let's you specify the z order of layers
-  `xResponse`, `yResponse`, and `scaleResponse`, which let you specify how a layer responds to camera motion.  Setting them all to 0 results in a static layer, something that's often been requested.   `xResponse` and `yResponse` by themselves will give parallax effects in conjunction with multiple layers.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,25 @@
+{
+    // See https://go.microsoft.com/fwlink/?LinkId=733558
+    // for the documentation about the tasks.json format
+    "version": "0.1.0",
+    "command": "grunt",
+    "isShellCommand": true,
+    "tasks": [
+        {
+            "taskName": "build:release",
+            "args": [],
+            "isBuildCommand": true,
+            "isWatching": false,
+            "problemMatcher": [
+                "$lessCompile",
+                "$tsc",
+                "$jshint"
+            ]
+        },
+        {
+            "taskName": "test",
+            "args": [],
+            "isTestCommand": true
+        }
+    ]
+}
--- a/playgrounds/parallax.html
+++ b/playgrounds/parallax.html
@@ -1,0 +1,89 @@
+<html>
+<head>
+  <script type="text/javascript" src="../crafty.js"></script>
+  <style>
+    #cr-stage {
+      background-color: gray;
+    }
+  </style>
+</head>
+<body>
+</body>
+<script type="text/javascript">
+  window.onload = function(){
+
+    
+    Crafty.init(200, 200);
+    Crafty.c("Bar", {
+      init: function() {
+        this.requires('2D, Color, Mouse');
+        this.attr({h: 100, w: 30});
+        this.bind("Click", function(){
+          window.alert(this._drawLayer.name);
+        });
+      }
+    });
+    
+    var layerType = "WebGL";
+    
+    // Define 4 different layers with different x parallax 
+    Crafty.createLayer("Background", layerType, {z: 10, xResponse: 0.2, scaleResponse: 1});
+    Crafty.createLayer("Midground", layerType, {z: 15, xResponse: 0.4, scaleResponse: 1});
+    Crafty.createLayer("Foreground", layerType, {z: 20, xResponse: 0.8, scaleResponse: 1});
+    Crafty.createLayer("ActionLayer", layerType, {z: 25, xResponse: 1});
+
+    // Define a UI layer that is completely static and sits above the other layers
+    Crafty.createLayer("UI", "DOM", {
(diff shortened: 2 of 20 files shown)


starwed:
I made the behavior of the `options` object support nested objects, which might be a complication we should just avoid?  The logic is currently local to `Crafty.s`, but I just opened an issue  (#1026) for doing something similar for Crafty as a whole.

mucaho:
Might be worthwhile to consider adding `pixelart` as a layer option too. That should solve feature request #882, as far as I understand.

mucaho:
> I made the behavior of the options object support nested objects, which might be a complication we should just avoid?

Is there a specific use-case that could create an issue here?

---

Would you happen to have an example file so I can test those changes and play a bit around with them?

starwed:
Realized I had completely failed to address issues with mouse events etc -- they need to know how to find which element was below the cursor.  (A similar issue with the debug layer, though that might be trickier, and less important to fix as part of this pr.)

> Would you happen to have an example file so I can test those changes and play a bit around with them?

Sure, I'll upload that once I've fixed the issue above.  (I have more free time than normal this week, so I'm going to try and finish this and a couple of other smaller patches!)

starwed:
Update: Dealing with pointer events was a huge pain, but I think I have them mostly handled correctly now.  There's still some weirdness around the mouseoff event, since the pointer is not actually over the entity that the event will be triggered on, but I think that'll be relatively easy to special-case.

starwed:
> >    I made the behavior of the options object support nested objects, which might be a complication we should just avoid?
> 
> Is there a specific use-case that could create an issue here?

I think this got discussed in a separate issue, but so that this one is self-contained: handling nested objects when merging options introduces a lot of corner cases.  I reverted back to handling only simple top-level properties, since that's a lot easier to understand.  It also allows e.g. triggering an event when an option changes to be more consistent, which is probably something we'd want going forward.

.vscode/tasks.json
```diff
@@ -0,0 +1,25 @@
+{
+    // See https://go.microsoft.com/fwlink/?LinkId=733558
+    // for the documentation about the tasks.json format
+    "version": "0.1.0",
+    "command": "grunt",
```
starwed, line 5: I added this file (`.vscode/tasks.json`) mostly for this option -- it lets you easily run our grunt tasks from within VSCode.  (I've found it better for most things than Sublime.  Plus, despite being an MS product, it's open source!)

Plyossib we should add the `.vscode` directory to `.gitignore` so that everyone can tweak this individually?

If not, I think we can set up some rules to show lints and errors directly in the editor.  Most of the other lines in this file are whatever vscode auto-generates.

starwed:
@mucaho  I added a simple playground file (`parallax.html`) to demonstrate the behavior.

starwed:
I'm going to go ahead and merge this.  AFAIK the only remaining wart is that the debug layer won't always play nicely with the new features, but I think that's ok as an iterative step.

mucaho:
Didn't look in detail yet, but do mouseevent `e.realX` and `e.realY` get properly computed depending on which layer the user clicked?

starwed:
> Didn't look in detail yet, but do  e.realX and e.realY get properly computed depending on which layer the user clicked?

Yes, though it's now assigned inside the `findPointerEventTargetByComponent` method.  (I wasn't super happy about needing to mutate the event inside that method, but it seemed the simplest approach.)

*e:* You can see this in action in the [playground file](https://github.com/starwed/Crafty/blob/6588c594403e47e6b334c378310a7c5a4d0cbf66/playgrounds/parallax.html).

mucaho:
Modified the playground file and the values returned by `realX` weren't what I was expecting. Instead of `realX = Crafty.domHelper.translate(x, y, ent._drawLayer).x` it should suffice to do `realX = Crafty.domHelper.translate(x, y).x`, although the former is still necessary to find correct entity clicked I reckon.   
Edit: It seems I'm misunderstanding things. The `realX` on a layer with `xResponse != 1` isn't actually what it would be on a layer with `xResponse == 1`. When the viewport is moved, entity position's aren't updated, merely their view on them is. Thus having `realX` on its own doesn't make sense - either it's always relative to the "default" (xResonse = 1) layer, or we introduce a `realX` property for each layer. 
However, for now, I think it makes sense to have the default `realX`, not the `realX` from a layer the user clicked.
Edit2: Yeah try adding `Draggable` to those bars, it doesn't work properly. I'm currently working on a viewport functionality and will submit this change as described in the first paragraph into that PR.

PR #1029, opened 2016-04-26
Various improvements

mucaho:
Notes:
- You may want to view diff with whitespace disabled (`url?w=0`) - [[diff link](https://github.com/craftyjs/Crafty/pull/1029/files?w=0)].
- There _should_ be no backwards incompatible changes, except for `.hit`.
  - `.hit` is changed to return `null` instead of `false`, which is also a [falsy value](https://developer.mozilla.org/en-US/docs/Glossary/Falsy), and thus _shouldn't_ change semantics of current code.
  - The reason for this change is to avoid possible deoptimizations. (see [respective wiki entry](https://github.com/craftyjs/Crafty/wiki/Ideas#fix-possible-deoptimizations))
- See [example_spriteAnimation_rowWrap.html](https://github.com/craftyjs/demos/blob/master/devDemos/example-spriteAnimation-rowWrap.html)
- See [example_collision_movement.html](https://github.com/craftyjs/demos/blob/master/devDemos/example_collision_movement.html)
- See [example_minefield.html](https://github.com/craftyjs/demos/blob/master/devDemos/example_minefield.html)
- I don't quite understand what [`tests/camera.html`](https://github.com/craftyjs/Crafty/blob/fbc7d61d95b110a85017a2495252e717f542297d/tests/camera.html) is trying to show. @mpetrovic Would you be so kind and shed some light on it if you happen to remember?


mucaho:
Merging these small, convenience changes.

Issue #1030, opened 2016-05-03
Hex colors with only 3 chars (E.G: #fff) don't work.


ChildishGiant:

matthijsgroen:
That's because it's not CSS. A pull request could be made. Maybe I can do so later this week

matthijsgroen:
There is currently an implementation for it, but is specs are faulty.

The test is that if color `#123` is assigned, the resulting RGB (ranged 255) should be:

```
r = 1
g = 2
b = 3
```

But this should actually be:

```
r = 17
g = 34
b = 51
```

I will update this spec and implementation in a pull request shortly

mucaho:
> There is currently an implementation for it, but is specs are faulty.

Yep, for reference, the [standard css3 color specification](https://www.w3.org/TR/css3-color/#rgb-color) says:

> The three-digit RGB notation (#rgb) is converted into six-digit form (#rrggbb) by replicating digits, not by adding zeros. For example, #fb0 expands to #ffbb00.

PR #1033, opened 2016-05-04
Fix issue #1030

matthijsgroen:
Parse a four character hex string (like #48F) as a 7 character one
(#4488FF)

### DIFF
--- a/src/graphics/color.js
+++ b/src/graphics/color.js
@@ -59,17 +59,24 @@ Crafty.extend({
         }
 
         function parseHexString(hex, c) {
-            var l;
-            if (hex.length === 7){
-                l=2;
-            } else if (hex.length === 4){
-                l=1;
+            var r, g, b,
+                l = hex.length;
+
+            if (l === 7) {
+                r = hex.substr(1, 2);
+                g = hex.substr(3, 2);
+                b = hex.substr(5, 2);
+            } else if (l === 4) {
+                r = hex.substr(1, 1); r += r;
+                g = hex.substr(2, 1); g += g;
+                b = hex.substr(3, 1); b += b;
             } else {
                 return default_value(c);
             }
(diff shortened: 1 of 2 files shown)


matthijsgroen:
@mucaho or @starwed could this be merged? It will change the original specs however (see Diff on test file). This fixes #1030

src/graphics/color.js
```diff
@@ -59,17 +59,18 @@ Crafty.extend({
...
                 return default_value(c);
             }
-            c._red = parseInt(hex.substr(1, l), 16);
-            c._green = parseInt(hex.substr(1+l, l), 16);
-            c._blue = parseInt(hex.substr(1+2*l, l), 16);
+            c._red = parseInt(hex.substr(1, 2), 16);
+            c._green = parseInt(hex.substr(3, 2), 16);
+            c._blue = parseInt(hex.substr(5, 2), 16);
             return c;
         }
```
mucaho, line 81: You could probably optimize this code slightly by removing redundant calls to `substr`:

``` js
function parseHexString(hex, c) {
    var r, g, b,
        l = hex.length;

    if (l === 7) {
        r = hex.substr(1, 2);
        g = hex.substr(3, 2);
        b = hex.substr(5, 2);
    } else if (l === 4) {
        r = hex.substr(1, 1); r += r;
        g = hex.substr(2, 1); g += g;
        b = hex.substr(3, 1); b += b;
    } else {
        return default_value(c);
    }
    c._red = parseInt(r, 16);
    c._green = parseInt(g, 16);
    c._blue = parseInt(b, 16);

    return c;
}
```

matthijsgroen, line 81: Ok I'll update it and I'll squash merge it back into develop
Op wo 4 mei 2016 om 14:18 schreef mucaho notifications@github.com

PR #1034, opened 2016-05-08
Introduce UpdateFrame event

mucaho:
This event replaces the `EnterFrame` event.
The `EnterFrame` event now triggers before the `UpdateFrame` event.

This is a potentially breaking change. While code may still work with binding to `EnterFrame`, it may introduce latency from previous / to next state update.

Could be useful for setting-up movement in response to input events before the state updates.
This is better than doing it after the state updates, which is already possible with `ExitFrame`, because there is added latency before the input is checked and after it is processed (up to `20ms` in the default case).
Additionally, having input processing bound to the `UpdateFrame` does not guarantee it will be executed before the `Motion` updates, which also bind to an arbitrary execution order triggered by `UpdateFrame`.
- [x] Replaces all triggered callbacks of the `EnterFrame` event to be triggered by `UpdateFrame` instead.
- [x] Fixes (hopefully) all api docs to reflect the change.
- [x] Fixes documentation on Crafty's website to reflect the change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Crafty.e("2D, DOM, Color, Collision")
     .attr({ x: 300, y: 150, w: 10, h: 10,
             dX: Crafty.math.randomInt(2, 5),
             dY: Crafty.math.randomInt(2, 5) })
-    .bind('EnterFrame', function () {
+    .bind('UpdateFrame', function () {
         //hit floor or roof
         if (this.y <= 0 || this.y >= 290)
             this.dY *= -1;
--- a/playgrounds/parallax.html
+++ b/playgrounds/parallax.html
@@ -80,7 +80,7 @@
         .textColor("white")
         .textFont({size: '20px', family:'Arial'})
         .attr({x: 10, y: 130, w: 200})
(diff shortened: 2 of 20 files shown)


starwed:
This is a pretty large change.  I don't think it's necessarily bad, but I'm also not sure how much an improvement it is.  I guess trying it is the best way to see!  :)

> Could be useful for setting-up movement in response to input events before the state updates.

IIRC the way input events are handled in Crafty is a little weird -- they don't actually happen at a specific part of Crafty's own game loop, but rather are directly triggered by the browser events.  (Possibly the way forward there would to be to  have the event dispatchers add them to an internal queue which is processed once per cycle.)

mucaho:
Updated this PR.    
One additional change is that the controls system listens to the "EnterFrame" instead of the "EnterFrameInput" event, since "EnterFrame" is the now triggered before the main "UpdateFrame" event.

mucaho:
> This is a pretty large change. I don't think it's necessarily bad, but I'm also not sure how much an improvement it is. I guess trying it is the best way to see! :)

Alright, let's do this!

PR #1040, opened 2016-07-18
Create .doclets.yml (JSDoc CI)

lipp:
Adds documentation to doclets.io service. Future tags and master branch will be published **automatically**.

[Preview generated with my account](https://doclets.io/lipp/Crafty/add-to-doclets)

Doclets also supports "live/editable" examples for the browser, like [this](https://doclets.io/lipp/acme-jsdoc-example/master#dl-greet)! Just replace the @example tag with @playground tag.

@mucaho  @starwed If you like this, you are very welcomed to login to https://doclets.io with GitHub OAuth. If you don't like it, I'd be happy to hear what you don't like about it :)

Steps required (3min):
1) @mucaho  logs in to doclets.io
2) @mucaho  adds Crafty
3) @mucaho  merges this PR

### DIFF
--- a/.doclets.yml
+++ b/.doclets.yml
@@ -1,0 +1,3 @@
+dir: src
+articles:
+ - Readme: README.md


starwed:
@lipp Unfortunately, it looks to me like this probably won't work very well with Crafty -- we use something that looks a lot like jsdoc, but has some custom extensions and differences.  Enough so that we have to maintain a custom parser for our format :( 

If we wanted to switch to a more standard format, we could probably modify the custom parser to transform the docblocks into that format (i.e. rewrite the source files automatically), so it's not completely hopeless.

PS: If anyone is curious about our existing doc build process:
-  [Step 1](https://github.com/craftyjs/Crafty/blob/develop/build/parseSourceDocs.coffee): parse the source docs
-  [Step 2](https://github.com/craftyjs/Crafty/blob/develop/build/parseNodes.coffee): transform the source docs into a json representation
- [Step 3](https://github.com/craftyjs/craftyjs.github.com/tree/master/source/api-gen): use react to build the api part of the site from the `api.json` file.

starwed:
Closing for the reasons discussed above, but feel free to reopen if there's a way around those issues.

PR #1041, opened 2016-07-24
Allow a jump to be triggered directly without faking a keypress

starwed:
A quick PR that would allow you to trigger a jump directly, instead of needing to fake a keypress.  (I've seen this come up a couple time in the google group, and there's no need to make people jump[1] through hoops.)
1.  Wow, didn't even notice the terrible unintentional pun here....


starwed:
This one is a very simple change, so merging.

PR #1042, opened 2016-07-24
Dynamic text updating

starwed:
Allows text entities to automatically update their content each frame if you call `textGenerationMode(true)`.  Supports specifying other update trigger names.

I still need to add tests, and probably change the name of the function that turns this behavior on/off.


starwed:
Renamed the method to dynamicTextGeneration (while fixing a bug with the react docs generator triggered by that name), and added some simple tests.

Issue #1018, opened 2016-03-07
Crafty.webgl:  program._registrySize permanently incrementing


CubixSystem:
I`m trying to implement rendering entities in viewport as @starwed and @Sythe2o0 advise in #963

If i run multiple `registerEntity()`/`unregisterEntity()` then `_registrySize` permanently incrementing until `this._registrySize >= this.max_size` (`this.max_size = 1024;`) and throw error

In `src/graphics/webgl.js`:

``` js
registerEntity: function(e){
        if (this._registryHoles.length === 0) {
            if (this._registrySize >= this.max_size){
                throw("Number of entities exceeds maximum limit.");
            } else if (this._registrySize >= this.array_size) {
                this.growArrays(2*this.array_size);
            }
            e._glBufferIndex = this._registrySize;
            this._registrySize++;
        } else {
            e._glBufferIndex = this._registryHoles.pop();
        }
    },

    // remove an entity; allow its buffer index to be reused
    unregisterEntity: function(e){
        if (typeof e._glBufferIndex === "number")
            this._registryHoles.push(e._glBufferIndex);
        e._glBufferIndex = null;
}
```

`this._registrySize++;` -only incrementing

CubixSystem:
I seems like my midnight coding problem. `unregisterEntity()` wasn't calling in my code.
I'm making some tests to check this.

PR #1045, opened 2016-07-28
Fix #1044 (webgl entities not unregistering) by providing correct argument

### LINKED ISSUE #1044: Counting towards "Number of entities exceeds maximum limit." seems off.
Somehow my game suddenly crashes halfway through (never did before, using same version of Crafty). After some investigation, the game crashes because of an 
"Number of entities exceeds maximum limit."

I understand the limit is 1024 entities, but somehow, when I add logging in this form:

```
setInterval(function() {
  console.log("Entities: " + Crafty('*').length);
}, 3000);
```

the logging reports around 230 entities active (counting all, not only the WebGL ones)

when I add logging around https://github.com/craftyjs/Crafty/blob/develop/src/graphics/webgl-layer.js#L69

The counting (showing the `_registrySize`) somehow jumps from 230 entities to 950 and rises quickly to 1024.

```
console.log("Entities " + (this._registrySize - this._registryHoles.length), "WebGL: " + Crafty('WebGL').length);
```

My question before further investigation and possible fixes: Would counting `Crafty('WebGL').length` be equal to `this._registrySize` ?

matthijsgroen:
See #1044  for details

### DIFF
--- a/src/graphics/webgl-layer.js
+++ b/src/graphics/webgl-layer.js
@@ -296,7 +296,7 @@ Crafty.webglLayerObject = {
         this._canvas = c;
 
         gl.clearColor(0.0, 0.0, 0.0, 0.0);
-        
+
         // These commands allow partial transparency, but require drawing in z-order
         gl.disable(gl.DEPTH_TEST);
         // This particular blend function requires the shader programs to output pre-multiplied alpha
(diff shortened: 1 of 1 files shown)


matthijsgroen:
@mucaho care for a quick review?

starwed:
Looks good to me, nice catch!

starwed:
#1018 was closed by the reporter, but might have been this same bug.

PR #1046, opened 2016-08-16
Change to the fourway component

Wanderer101:
I have tried to limit diagonal movement for the fourway component. Previously, the fourway component was a thin wrapper for multiway, which simpy adds each keypress to the vx and vy. With the changes I propose, the fourway component limits diagonal movement and does not add velocity but sets a new velocity with each key press, and checks for keys still pressed at the keyup event setting the new velocity accordingly. I am using this implementation for an isometric game. If one were to use it for an rpg it would probably have to be limited further. Perhaps an fourway rpg component should be created as well.  

As a final note, I am unsure whether I handled the new binding of keydown and keyup in the simplest way. Let me know.

### DIFF
--- a/src/controls/controls.js
+++ b/src/controls/controls.js
@@ -530,6 +530,59 @@ Crafty.c("Fourway", {
 
     init: function () {
         this.requires("Multiway");
+        
+        this.unbind("KeyDown", this._keydown)
+            .unbind("KeyUp", this._keyup);
+        
+        this.bind("KeyDown", this.__keydown)
+            .bind("KeyUp", this.__keyup);
+    },
+
+    __keydown: function (e) {
+        var direction = this._keyDirection[e.key];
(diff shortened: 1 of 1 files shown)


200sc:
Is it possible, given my understanding that this acts very differently than the current fourway implementation, that this should be a separate component that could be in the [community components list](http://craftyjs.com/components/), instead of replacing the current fourway?

Wanderer101:
It is definitely a possibility to do that. However, the fourway component does seem to need a fix since it is actually just multiway with some keys defined. Therefore, it still provides 8-way movement which I'm thinking is not the intended functionality.

200sc:
Personally, when I use fourway (pretty much exclusively for early builds or quick game jams) I expect diagonal movement.

starwed:
Hopefully #1051 meets the requirements here, definitely reopen this or another PR/issue if there are additional related changes needed.

Issue #1047, opened 2016-09-02
Clamp speed for multiway


matthijsgroen:
I'd like to add a 'clamp' mechanic to the multiway.

It's effect: 

``` javascript
  directions = {UP_ARROW: -90, DOWN_ARROW: 90, RIGHT_ARROW: 0, LEFT_ARROW: 180};

  clamped = Crafty.e('2D, Multiway').multiway({x: 10, y: 10 }, directions, { clamped: true })
  current = Crafty.e('2D, Multiway').multiway({x: 10, y: 10 }, directions)

```

When the up arrow is pressed:

``` javascript
  clamped.vx // 0
  clamped.vy // -10

  current.vy // 0
  current.vy // -10
```

When both the up and the right arrow are pressed:

``` javascript
  clamped.vx // Math.sin(Math.PI / 4) (45deg) * 10 = 7.07
  clamped.vy // Math.cos(Math.PI / 4) (45deg) * -10 = -7.07

  current.vx // 10
  current.vy // -10
```

Should this become a new component like 'MultiwayClamped' or an option of the existing multiway?

matthijsgroen:
Reason for me to want this feature is feedback I got on my game here: http://www.shmup-dev.com/forum/viewtopic.php?f=4&t=515&p=1563#p1562

matthijsgroen:
See #1048 for details

starwed:
Fixed (I think) by #1051, when `multiway` is called with `{normalize: true}`.

PR #1048, opened 2016-09-02
Add "Clamp" option to Multiway

matthijsgroen:
Clamp will simulate the same effect as a analog stick would have.

e.g. The diagonals will not make the entity move faster, but
make it just move in another direction.

Implements #1047

### DIFF
--- a/src/controls/controls.js
+++ b/src/controls/controls.js
@@ -204,28 +204,32 @@ Crafty.c("Multiway", {
     /**@
      * #.multiway
      * @comp Multiway
-     * @sign public this .multiway([Number speed,] Object keyBindings)
+     * @sign public this .multiway([Number speed,] Object keyBindings[, Boolean clamped])
      * @param speed - A speed in pixels per second
      * @param keyBindings - What keys should make the entity go in which direction. Direction is specified in degrees
+     * @param clamped - Clamps the speed for the diagonal movements.
      *
      * Constructor to initialize the speed and keyBindings.
      * Component will listen to key events and move the entity appropriately.
      * Can be called while a key is pressed to change direction & speed on the fly.
+     *
+     * When clamping is turned on diagonal directions will not use the full `speed.x` and `speed.y` but a fraction to
+     * simulate movement of a joystick axis. Pushing right would be a +1 for the right direction, but
+     * diagonal right would only be a +0.7 in that direction, since it is also moving downward or upward.
      *
      * @example
      * ~~~
      * this.multiway(150, {UP_ARROW: -90, DOWN_ARROW: 90, RIGHT_ARROW: 0, LEFT_ARROW: 180});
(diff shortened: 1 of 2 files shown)


matthijsgroen:
I do need to update the documentation still, but the implementation is ready for review

matthijsgroen:
Done!

matthijsgroen:
movement-speed-clamping

To give a better illustration of the effect.
Let's say you want the speed to be 50 units in either direction. At the moment you can exceed that speed by going diagonal. Since you are going UP **and** DOWN with 50. With clamping turned on that would be 35,3 UP and DOWN. (using Sin and Cos of the angle of 45 deg)

src/controls/controls.js
```diff
@@ -294,21 +318,57 @@ Crafty.c("Multiway", {
...
+          var avgAngle = this.__directionsToRad(directions);
+          this.vx -= Math.round(Math.abs(vx) * Math.cos(avgAngle) * 1000) / 1000;
+          this.vy -= Math.round(Math.abs(vy) * Math.sin(avgAngle) * 1000) / 1000;
+        } else {
+          this.vx -= vx;
+          this.vy -= vy;
+        }
+    },
+
+    __directionsToRad: function(directions) {
```
starwed, line 362: IIRC I  looked at this approach before and ran into some problems.  The average of even two angles is ill-defined:  if the inputs are left (280) and right (0), you could get either 90 or 270 degrees as the average.  And neither is what you'd intuitively expect from the averaged input, which would simply be the zero-vector.

Instead of averaging the angle, I think you want to take the existing implementation and normalize the resultant vector.  You'd have an additional pass that would be something like:

``` javascript
var unnormed = sqrt(this.vx * this.vx + this.vy * this.vy);
if (unnormed > 0.00001) {
    this.vx = this.vx * (this._clampSpeed / unnormed);
    this.vy = this.vy * (this._clampSpeed / unnormed);
} else {
    this.vx = 0;
    thix.vy = 0;
}
```

_e:_ Added a branch to catch the zero vector case -- check if the velocity is greater than some threshold, to catch small rounding errors.

starwed:
Added a comment about this particular implementation, but I also have a general plan to try to uncouple some of this logic form Multiway.  I opened #1049 to relate to this.

matthijsgroen:
The problem is also that there is no "clamped speed" the x and y can differ in speed

matthijsgroen:
I could see if I can make the directions method smarter with more tests to rule out contradicting angles

matthijsgroen:
@starwed I now reject all opposing directions before calculating the angle.

starwed:
I think #1051 produces basically the same effect (when you pass in the flag to normalize input.)  If that doesn't produce the necessary effect, definitely feel free to open this/another PR/issue (and yell at me for misunderstanding :P)

PR #1051, opened 2016-09-17
Pull control logic out of Multiway into a separate system

starwed:
This is a first draft at pulling the control logic out of Multiway into a system.  It's not fully tested, ~~and a lot of the names are temporary~~, but since there have been a couple of multiway related suggestions lately I thought I should present this alternative.  @matthijsgroen  @Wanderer101
- The control system lets you define dpads using the same syntax that Multiway uses, and emits events with a directional vector
- A new "Dpad" component allows you to tie callbacks to specific dpad events.  Multiway uses this concept to update the objects speed when necessary.
- The new system has options for normalized/clamped input (#1047) and axis-only movement (#1046, aka only listen to one direction button at a time)
- Behind the scenes, instead of listening for keyup/keydown events, the control system checks which keys are down each frame, and update the direction based on that info.  That makes it easier to maintain a consistent state.
- To remove any ambiguity in when motion/control events happen, I tried introducing a new top level event (called "EnterFrameInput" for now) that fires before "EnterFrame".  The idea here would be to first update any state based on user input, and then actually process the game frame.  I'm not committed to this idea, but it does seem useful.

TODO
- [x] Fix tests

### DIFF
--- a/tests/controls.js
+++ b/tests/controls.js
@@ -1,25 +1,19 @@
 (function() {
   var module = QUnit.module;
-
+  var keysUp = global.keysUp;
+  var keysDown = global.keysDown;
   module("Controls");
-
+  
   test("Multiway and Fourway", function() {
     var e = Crafty.e("2D, Fourway")
                   .attr({ x: 0, y: 0});
 
-    Crafty.trigger('KeyDown', {
-      key: Crafty.keys.W
-    });
-    e.isDown = function(key) {
-        if (key === Crafty.keys.W)
-            return true;
-        return false;
-    };
+    
     e.multiway(50, { W: -90 });
+    keysDown(Crafty.keys.W);
     Crafty.timer.simulateFrames(1);
     equal(e._vy, -50, "Speed is 50 in -y direction");
     equal(e._vx, 0, "Speed is 0 in x direction");
-    equal(e._y, -1, "Moves 1 pixel down in 20 ms");
 
     // Change the key's direction and speed while it's held down
     e.attr({x:0, y:0});
@@ -27,7 +21,6 @@
     Crafty.timer.simulateFrames(1, 20);
     equal(e._vy, 100, "Speed is 100 in +y direction");
     equal(e._vx, 0, "Speed is 0 in x direction");
-    equal(e._y, 2, "Moves 2 pixels up in 20 ms");
 
     // Change the speed with fourway, (W is negative for fourway)
     e.attr({x:0, y:0});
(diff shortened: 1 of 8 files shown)


Wanderer101:
Great to see this component get an overhaul! I'll try to learn something as well by reading up on the changes. Thanks for addressing this issue.

starwed:
I updated this with a few bug fixes, so the new features actually work now.  :)

When you disable multiple directions, it will always stick with the initial direction when you press additional keys.  I think that's what most people would want, but perhaps I'm wrong?

Wanderer101:
Not to be difficult but could it offer both, perhaps set by the options object? I haven't tried the controls when bound to EnterFrame (or EnterFrameInput) yet, but at least my use case needs the new button press to override the previously held one.

starwed:
> Not to be difficult but could it offer both, perhaps set by the options object? I haven't tried the controls when bound to EnterFrame (or EnterFrameInput) yet, but at least my use case needs the new button press to override the previously held one. 

Refactored it to allow this strategy -- the option is now `multipleDirectionBehavior` and takes "all", "first", or "last" as values.  

It now tracks when an "input group" is first active -- so it doesn't care about the individual key you're pressing.   This is probably something you want in general --  many games use charged shots rts or jumps, and now the input can tell you how long it's been depressed.

tests/controls.js
```diff
@@ -95,72 +75,73 @@
...
+    Crafty.timer.simulateFrames(1); 
+    equal(e._vx, 50, "vx = 50 when key is released while control disabled");
 
-    Crafty.trigger('KeyUp', {
-      key: Crafty.keys.D
-    });
-    equal(e._vx, 0, "vx = 0 when key is released while control is disabled");
+    e._vx = 17;
+    Crafty.timer.simulateFrames(1);
+    equal(e._vx, 17, "vx = 17 after being explicitly set while control disabled");
```
starwed, line 100: This is the one point where the behavior has deliberately changed: disabling control no longer resets the velocity of the entity.  Easy enough to set velocity to 0 if that's what you want when calling disable control.

Wanderer101:
Thanks a bunch! I can't wait to try out this new functionality.

matthijsgroen:
I hope to test run this in my game soon, I hope to have time for this tomorrow!

starwed:
I'm going to go ahead and merge this, since it seems to work in my basic tests.  But the internals of the change will remain undocumented for a little bit -- I want to see how easy it is to adapt other types of inputs to use this control pattern, and what changes might be necessary to the underlying system.

matthijsgroen:
@starwed  I'm trying to grasp the logic of the new split-up components, but some parts are not really clear to me.

1. There is a 'ControlSystem' here, you define a virtual gamepad
2. Next link an event of this virtual gamepad to a handler (directionalInput)

As far as I can see, only mouse and keyboard are supported for this virtual gamepad (not a real gamepad)

I like the idea of having controls just emit events, and the entity responds to those events (regardless if it came from keyboard or mouse). But it doesn't seem open for extension to support a real HTML5 gamepad.

So, what is the philosophy? Extend the crafty control system to support real gamepads, or duplicate the virtual gamepad mechanic to emit the same triggers?

Now the gamepad has a special 'GamepadMultiway' component, but I guess this should now work with a normal multiway component when a proper virtual gamepad could be constructed for the real gamepad?

I'm a bit confused what the correct upgrade strategy would be to support https://github.com/svenjacobs/crafty-gamepad/blob/master/src/crafty.gamepad.js
within Crafty 0.80

starwed:
The main goal here was to separate the existing Crafty controls.  The new pattern is to essentially create three layers:

- Input code: listens for HTML events and emits Crafty input events
- Control code: listens for Crafty input events and maps them to control events.
- Game state code: listens for control events and produces state change in the game; controllable entities are the obvious example

Previously Crafty jumbled up the second two.  The code changes in 0.8 mostly focus on separating those out, but *don't* yet provide a lot of support for adding new types of input->control adapters -- it's hardcoded to handle the types that we already handled.   The idea is to iterate on this with real world use cases such as gamepad input.

The next step would probably be to more formally define the contracts that new adapters need to provide.  There's a vague nod in that direction with the two "ToggleInput" objects -- we have both `MouseButtonToggleInput` and `KeyboardToggleInput` classes, which implement a simple `isDown()` method.  Right now everything is hard coded, but in the future we could allow you to register additional input types and make the define functions more generic.  (In addition to gamepads, another obvious use case would be to move an entity in the direction you click/touch.  I have a stub of that work in a branch I'll probably work on once 0.8.0 is released.)

It's possible that we should rename `defineDpad` to something like `defineKeyboardDpad` and reserve `defineDpad` for a future generic api -- thoughts?

>I'm a bit confused what the correct upgrade strategy would be to support (the existing gamepad component) within Crafty 0.8.0

A start would be taking the existing code and getting it to emit control events instead of directly setting the entities state.  Then it might become clearer how to pull that work into the existing controls system (and, at the same time, how to make the existing control system more generalizable.)   That said, it doesn't look like those components rely on Multiway, so perhaps they would continue to work as is for now?

mucaho:
Good job on improving this system! Unfortunately both (!) of my pcs died after the summer and I have just gotten recently to installing a development environment. The control system has been released to public. I have some ideas that should simplify things but would break the API considerably. Is this something that would be ok if I made such a PR in the future, or is it too late now?

starwed:
>I have some ideas that should simplify things but would break the API considerably. Is this something that would be ok if I made such a PR in the future, or is it too late now?

I'm not opposed to breaking the api if the changes are worthwhile!  I'd suggest opening an  issue or stub PR sketching out the approach you're thinking of first.

mucaho:
My main motivation was figuring out the `linkInput` and `unlinkInput` methods, as it didn't seem very Crafty-esque that the controls take callbacks instead of dispatching appropriate events. However, after looking at the code in detail, they are mere convenience wrappers around binding to actual events.
`linkInput` and `unlinkInput` could be removed by triggering and binding to events with the `triggerName` / `dpadName` as the eventName instead.    
I like the rest of the functionality provided by the control system, it opens up many different possibilities without the need to code your own stuff, good job 👍

Issue #781, opened 2014-04-24
License info should be much more obvious; make GPL version clear.


ghost:
I have found nothing about the license on GitHub. There should be a LICENSE file or something in README.
In the footer website, it is written "Crafty is distributed under the MIT or GPL license". But what parts are under MIT or GPL? What version(s) of GPL?

starwed:
> I have found nothing about the license on GitHub. There should be a LICENSE file or something in README.

You're right... the current licensing info is hidden in a pretty weird place; when you compile `crafty.js` with grunt it adds in a banner that [includes the line](https://github.com/craftyjs/Crafty/blob/develop/Gruntfile.js):

> Dual licensed under the MIT or GPL licenses.

This kind of answers your second question -- the intent is that it's [dual licensed](http://stackoverflow.com/questions/3336161/mit-gpl-dual-license-in-commercial-software) under both, so you someone using the code or forking it could do so under the terms of _either_ license.  But looking at the actual text, it doesn't seem to specify which version of the GPL.

None of the original authors are currently working on Crafty, and I'm not quite sure what to do about the ambiguity of the GPL notice.  The MIT license is pretty permissive, though, and I _believe_ you can even redistribute MIT licensed code under GPL 2 with no problems, so this shouldn't be too much of an issue.  I'll do some more research on the matter!

I'll tag @louisstow  in case he has something to add as the original author!  :)

starwed:
(Just wanted to note that anything under GPL 1 or 2 can use any later version, so it's certainly ok to use a GPL3 license if you want!)

ghost:
Ok, thanks for the answer.
GPL2 is incompatible with GPL3 : https://www.gnu.org/licenses/license-list.html#GPLv2
If we believe the FST, MIT license is compatible with GNU GPL, consequently I do not understand why it is dual-license. https://www.gnu.org/licenses/license-list.html#Expat https://www.gnu.org/licenses/license-list.html#X11License

starwed:
> GPL2 is incompatible with GPL3 

Ah, didn't realise that the "or later" clause was optional.

There is at least on difference: If you want to do something like fork the dual-licensed code under the GPL only, then you don't need to keep the original MIT license around.  If it was MIT-only to begin with, you could redistributed under the GPL, but you couldn't remove the MIT notice on the original code.

Jookia:
[...]

PR #1054, opened 2016-10-09
Clarify licensing + MIT only

starwed:
The actual compiled crafty file indicates that it's under the MIT + GPL licenses.  As flagged in issue #781, there isn't an actual license file present in the repo, and there's no mention of what version of the GPL is intended.

This adds the MIT license to the repo.  Possibly we should simply nix the mention of the GPL, since it's too vague to be useful, and the MIT is more permissive in any case.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -8,7 +8,7 @@ module.exports = function (grunt) {
                 ' * <%= pkg.name %> <%= pkg.version %>\n' +
                 ' * <%= pkg.author.url %>\n *\n' +
                 ' * Copyright <%= grunt.template.today("yyyy") %>, <%= pkg.author.name %>\n' +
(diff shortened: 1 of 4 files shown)


starwed:
I believe the license is indicated in some config files  -- I'll look into updating them.

starwed:
I'm going to go ahead and merge this.  To address one concern from the related issue:

> once decision is proposed, it needs to be approved by the author of this project - louisstow

Since he hasn't expressed any particular opinion here, I feel ok going ahead without him -- the point of dual licensing is that anyone can distribute the code under _either_ license.  What we're doing here is simply choosing a single one of the allowed licenses to use going forward.

I created a parallel PR in the docs/site repo: craftyjs/craftyjs.github.com#77

PR #1058, opened 2016-10-30
Upgrade packages and add yarn.lock

starwed:
This updates most packages to their most recent version.  There were a few issues/incompatibilities.
- I left `grunt-webdriver` alone because upgrading it broke the custom `webdriverio` that @mucaho  created.
- There's [an issue](https://github.com/ariya/phantomjs/issues/14050) with the upgraded phantomjs package that prevents passing in the `noglobals` flag when running qunit tests, so I removed that.
- `grunt-node-qunit` made some assumptions about how the node modules were arranged that broke.  (Probably due to upgrading npm?)  It's pointing at a fork I made to fix the issue for now.
- Updating react required a small change to how the api browser generates markup.  The equivalent change was already made in the actual docs repo.
- I used [yarn](https://yarnpkg.com/) to install them, and checked in the `yarn.lock` file.  That should hopefully ensure that a fresh install of Crafty won't get suddenly broken by upgraded dependencies in the future.


starwed:
Looks like package installation fails on old versions of node (and/or the versions of npm associated with them); it treats a conflicting sibling dependency as an error instead of a warning.   I updated the version of node that gets used by Travis.

(The conflicting dependency is caused by the non-updated `grunt-webdriver`, which wants an obsolete version of `grunt`.)

_e:_ I tried running `yarn` on Travis, but it seemed to run into some sort of error.  I imagine they'll add native support for yarn some time soon, so I'll wait til then before making another attempt.

mucaho:
Yeah, sounds like it's time to settle the technical debt.     
Need to finish [qunit test environment for the new version of webdriverio](https://github.com/mucaho/wdio-qunit-framework) so we can use that and finish a simple node qunit runner I was also working on.

mucaho:
I'm currently updating all tests to work on QUnit >2.0.0. Expect heavy changes to test files, I will also organize them a bit.

PR #1059, opened 2016-11-19
Document the  "kind" of thing associated with a docblock

starwed:
I went through and annotated all documented items with a `@kind` tag.  This could be used by the doc generator to show a little more info/context about items.  This PR just has the minimal modifications so that this attribute is parsed correctly.

The main motivation here is the addition of systems to the documentation.

The values I used were: "Component", "System", "Method", "Property", ~~"Constructor"~~ "Class", and "CoreObject".  (The last used to indicate the "important" objects such as `Crafty.viewport`.)

Class could probably be used instead of Constructor.  Crafty doesn't use very many of these in the public interface, but there are a few (such as the polygon class.)

### DIFF
--- a/build/api-gen/doc-components.jsx
+++ b/build/api-gen/doc-components.jsx
@@ -26,8 +26,7 @@ var MarkdownBlock = React.createClass({
   render: function() {
     var raw = this.props.value;
     var rawHtml = this.convert(raw);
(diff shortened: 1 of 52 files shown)


starwed:
Merging this -- next step is to update the doc generator to take advantage of it.

PR #1060, opened 2016-11-27
Yarn/phantomjs test branch

starwed:
The good news is travis ci now supports yarn.  The bad news is that the [phantomjs install is broken](https://github.com/yarnpkg/yarn/issues/1538).  (I think this is environment-specific -- it worked for me on osx but broke on travis.)  This PR is to test the workaround suggested in the linked issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,8 @@ language: node_js
 node_js:
 - '6'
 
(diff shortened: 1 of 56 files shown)


starwed:
Closing, since  #1061  fixed this.

(Instead of juggling env. params, just use a new version of phantomjs that doesn't have the issue!)

PR #1061, opened 2016-12-11
Fix build issue by upgrading from older phantomjs install method

starwed:
Try a different route to fixing the phantomjs+travis+yarn issue.


starwed:
Looks like this worked, so merging.  (It was almost definitely a bad idea to have two versions of phantomjs hanging around in any case.)

Issue #1064, opened 2016-12-24
Hitboxes do not work properly with 'static' layers


Dannesley:
I'm trying to create a dynamic HUD, where clicking on a button will expand a menu. I'm unable to because on a static layer i've created the hitbox of items in that layer moves with the viewport, while the position does not. As such, things like Crafty Mouse events are unreliable. There will be some points where the hitbox and position of the item line up, of course, but not always.

Steps to reproduce:
Have a scene with a movable viewport (e.g. follows a character)
Create a layer where the xResponse, yResponse and scaleResponse are set to 0
Create an entity in that layer, at some static x and y co-ordinate
Try to listen to Mouse events on that entity, such as clicking
Add the WiredHitBox component to see what the hitbox looks like

Outcome:
Mouse events are not fired when clicking the entity in the static layer
Hitbox moves with the viewport, becoming out of sync with the entity position

Expected:
Mouse events fire when clicking on an entity in the static layer
Hitbox mirrors the position of the entity

starwed:
> Hitbox moves with the viewport, becoming out of sync with the entity position

Yes, this (the debug layer stuff) is broken right now -- the debug layer wasn't updated to handle the layer response parameters.  I think I see a pretty quick fix for that, so I'll try to get that merged in soon.  (Just to be clear, this is a problem with the debug layer, not the hitboxes themselves.)

> Mouse events are not fired when clicking the entity in the static layer

Hmm, this *should* work.  Does this occur only once the viewport has panned, or all the time?

starwed:
Landed a fix for the wiredhitbox issue; the [most recent nightly](https://github.com/craftyjs/Crafty-Distro/tree/nightlies) should now display that properly as the viewport pans.

I tested locally, and didn't have an issue with clicking on a static element, so it might be some more specific interaction.  Hopefully easier to figure out with a working debug layer!  Sample code or a jsfiddle might make it clearer what's happening.

Dannesley:
Merry Christmas, and thanks for the quick response!

It happens all the time, but with your fix in, the problem is more apparent.
<img width="1042" alt="screen shot 2016-12-25 at 11 00 07 am" src="https://cloud.githubusercontent.com/assets/6972092/21469181/7154170c-ca91-11e6-9ea5-b39ca0876e25.png">
Seems likely to be a problem on my end, but a bit odd because inspecting the 2D component x and y properties return the expected value. All the static entity hitboxes seem to be offset by the same amount.

[...]

PR #1065, opened 2016-12-24
Debug canvas should  match the view transform of each entities layer

starwed:
Apply the correct context transformation for each debug entity, so that wireframes match the entities visual location.

Related to #1064, but not necessarily the root problem.


starwed:
Since this only affects the debug canvas, going ahead and merging.

PR #1066, opened 2016-12-28
Handle parallax plus scaling more gracefully

### LINKED ISSUE #1064: Hitboxes do not work properly with 'static' layers

starwed:
When the viewport is scaled *and* layers have special x/y response values, it's somewhat complicated to handle correctly.

This replaces the previous broken logic with a "correct" approach. Under the new transform, a zoom in on the center of the viewport correctly keeps layers from moving relative to each other. This uniquely constrains the properties under other camera movements.

I also switched the canvas and dom layers to use the `_viewportRect` helper to establish the necessary transforms.

I believe this fixes #1064

### DIFF
--- a/playgrounds/parallax.html
+++ b/playgrounds/parallax.html
@@ -24,7 +24,7 @@
       }
     });
     
-    var layerType = "WebGL";
+    var layerType = "Canvas";
     
     // Define 4 different layers with different x parallax 
     Crafty.createLayer("Background", layerType, {z: 10, xResponse: 0.2, scaleResponse: 1});
--- a/playgrounds/parallax.html
+++ b/playgrounds/parallax.html
@@ -62,9 +62,9 @@
       // Note that the globalZ value determines which item is on the "top" for mouse clicks
       // not the z value of the individual layer -- do we want to change this?
       for (var i = 0; i < 10; i++){
-        Crafty.e("Background, Bar").color("blue").attr({x: 100*i, y: 10, z: 30});
-        Crafty.e("Midground, Bar").color("purple").attr({x: 100*i, y: 20, z: 30});
-        Crafty.e("Foreground, Bar").color("yellow").attr({x: 100*i, y: 30, z: 10});
+        Crafty.e("Background, Bar, Collision, WiredHitBox").color("blue").attr({x: 100*i, y: 10, z: 30});
+        Crafty.e("Midground, Bar, Collision, WiredHitBox").color("purple").attr({x: 100*i, y: 20, z: 30});
+        Crafty.e("Foreground, Bar, Collision, WiredHitBox").color("yellow").attr({x: 100*i, y: 30, z: 10});
       }
 
       // Some text that shows the viewport's current position, and sits in the UI layer
--- a/src/graphics/canvas-layer.js
+++ b/src/graphics/canvas-layer.js
@@ -154,10 +154,10 @@ Crafty._registerLayerTemplate("Canvas", {
         // Set the camera transforms from the combination of the current viewport parameters and this layers 
         var cameraOptions = this.options;
         if (dirtyViewport && cameraOptions) {
-            var view = Crafty.viewport;
-            var scale = Math.pow(Crafty.viewport._scale, cameraOptions.scaleResponse); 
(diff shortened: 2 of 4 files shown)


starwed:
It was actually a bit complicated to get the result, and too lengthy a digression for an in-code comment.  So here's the logic that leads to the correct transform... feel free to ignore this, I just wanted it to be written down somewhere.  :)

The problem to solve: given the camera and layer settings, find the corresponding x/y/w/h for the layer that needs to be displayed in the viewport.

Since the axes are independent, just solve the problem for the x-axis.  First define the variables
- The scale is `s`
- The x-response factor is `r`
- The viewport coordinates are `x_v` and `w_v`
- The resulting top-left coordinates in the layer are `x_0` and `w_0`

Any given point `x` in the layer can be expressed as `x = x_0 + \lambda w_0`, and will be displayed at viewport position `x_v + \lambda w_v`.
Varying `\lambda` between 0 and 1 restricts us to the visible area.

Define the behaviors in a few simple, special cases:
- If `s` and `r` are both 1, the solution is trivial: `x_0 = x_v` and `w_0 = w_v`.
- If `s = 1` but `r` does not, then `x_0 = r * x_v`, `w_0 = w_v`.
- If `r = 1` but `s` does not, then `x_0 = x_v`, `w_0 = w_v / s`.

When both `s` and `r` are not equal to 1, then we have some freedom about what the transformation will look like.

Two obvious constraints are:
- We should preserve the relation `w_0 = w_v / s` (otherwise the scale factor loses it's meaning[1])
- When either `s` or `r` are set to 1, the previous behavior should emerge

To uniquely determine the transform, we need to pick some additional behavior we wish to preserve.   Consider a layer where `r=1`.  Imagine zooming in on the center of the viewport -- altering the `s` and `vx` such that  the point corresponding to the center of the viewport *remains* at the center of the viewport as we zoom.  It turns out that if we  demand the same behavior for *any* value of `r`, it uniquely determines the transformation.

First, what change in `vx` actually corresponds to such a transformation?  If the middle of the viewport should be fixed, and `r=1`, then that corresponds to saying that (using `1` and `2` for before and after)
  `x_1 = x_v1 + 0.5 w_v = x_2 = x_v2 + 0.5 w_v/s`.  
This then gives us:
  `x_v1 = x_v2 - 0.5 (1/s - 1) w_v`.

Second, what does that tell us about the case when `r` is not 1?

We get a similar constraint for `x_2_0`, by comparing the middle of the viewport before and after the very same transformation, and then subbing in the above expression for `x_v1` in terms of `x_v2`.
```
(x_1 = r * x_v1 + 0.5 w_v) = (x_2 = x_2_0 + 0.5 w_v / s)
  x_2_0 = r * x_v1 + 0.5 w_v - 0.5 w_v / s  
        = r * (x_v2 - 0.5(1/s - 1)w_v) + 0.5 w_v - 0.5 w_v / s 
        = r * x_v2 + 0.5 * (1 - r) ( 1 - 1/s) w_v
```

This is the final result; `x_0 = r * x_v + 0.5 *(1-r)(1-1/d)w_v`.  We can see that when either `r = 1` or `s = 1` the second term vanishes,  leaving results consistent with those special cases.

1. I guess the other option would be to remove this constraint, and try to apply some physical model to the x/y response.  In such a case as you reduced the response factor, the scale factor would have less of an effect.

PR #1071, opened 2017-01-08
Add an alias to map image_whitelist to imageWhitelist

starwed:
There didn't seem to be a natural place to put this, so I added a new aliases module that would define getter/setters that map old property names to the new ones.

@matthijsgroen

### DIFF
--- a/src/aliases.js
+++ b/src/aliases.js
@@ -1,0 +1,16 @@
+
+function createDeprecatedAlias(baseObject, oldName, newName) {
+    Object.defineProperty(baseObject, oldName, {
+        enumerable: false,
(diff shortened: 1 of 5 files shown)


matthijsgroen (approved):
Looks good to me. Should the deprecation also issue a warning on the console?

starwed:
>Should the deprecation also issue a warning on the console?

Was wondering that myself, but I think this is ok for a pre-release fix.  I figured one benefit to making the aliasing quasi-generic is so we can go back and add this in once we have some sort of official deprecation policy.

matthijsgroen:
:+1:

PR #1087, opened 2017-02-26
Allow entities to be "frozen"

starwed:
For various reasons, you might want to essentially remove an entity from the game, but preserve it's current state.   Two example use-cases:

- The entity is temporarily removed from the stage, but will be restored later.
- A large number of similar entities are quickly being created and destroyed; you might want to recycle them instead

This PR introduces basic support for `freeze()` and `unfreeze()` methods on entities.  A frozen entity:
- No longer responds to events
- Will not be rendered by a graphics layer (in the current implementation, it is still considered attached to the layer itself)
- Is removed from the spatial hash map

The current implementation has a couple of caveats that might need to be fixed:
- Entities are still returned by the Crafty() selectors
- Since events are frozen, it might be possible for entities to get into inconsistent states.  I need to make sure that this is handled in a friendly manner for existing entities.

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -1041,6 +1041,93 @@ Crafty.fn = Crafty.prototype = {
             this._unbindAll();
             delete entities[this[0]];
         });
+    },
+
+    /**@
+     * #.freeze
+     * @comp Crafty Core
+     * @kind Method
+     * 
+     * @sign public this .freeze()
+     * 
+     * @triggers Freeze - Directly before the entity is frozen
+     * 
+     * Freezes the entity.  A frozen entity will not receive events or be displayed by graphics systems. 
+     * It is also removed from the spatial map, which means it will not be found by collisions, 
+     * raycasting, or similar functions.
+     * 
+     * This method may be called upon a collection of entities.
+     * 
+     * @note Because the entity no longer listens to events, modifying its properties can result in an inconsistent state.
+     * 
+     * If custom components need to handle frozen entities, they can listen to the "Freeze" event, which will be triggered before the event system is disabled.
+     * 
+     * @example
+     * 
+     * ```
+     * // Freeze all entities with the Dead component
+     * Crafty("Dead").freeze();
+     * ```
+     * 
+     * @see .unfreeze
+     */
+    freeze: function () {
+        if (this.length === 1 && !this.__frozen) {
+            this.trigger("Freeze", this);
+            this._freezeCallbacks();
+            this.__frozen = true;
+        } else {
+            for (var i = 0; i < this.length; i++) {
+                var e = entities[this[i]];
+                if (e && !e.__frozen) {
+                    e.trigger("Freeze", e);
+                    e._freezeCallbacks();
+                    // Set a frozen flag.  (This is distinct from the __callbackFrozen flag)
+                    e.__frozen = true;
+                }
+            }
+        }
+        return this;
+    },
+
+    /**#
+     * #.unfreeze
+     * @comp Crafty Core
+     * @kind Method
+     * 
+     * @sign public this .unfreeze()
+     * 
+     * @triggers Unfreeze - While the entity is being unfrozen
+     * 
+     * Unfreezes the entity, allowing it to receive events, inserting it back into the spatial map, 
+     * and restoring it to its previous visibility.
+     * 
+     * This method may be called upon a collection of entities.
+     * 
+     * If a custom component needs to know when an entity is unfrozen, they can listen to the "Unfreeze"" event.
+     * 
+     * @example
+     * ```
+     * // Bring the dead back to life!
+     * Crafty("Dead").unfreeze().addComponent("Undead");
+     * ```
+     */
+    unfreeze: function () {
+        if (this.length === 1 && this.__frozen) {
+            this.__frozen = false;
+            this._unfreezeCallbacks();
+            this.trigger("Unfreeze", this);
+        } else {
+            for (var i = 0; i < this.length; i++) {
+                var e = entities[this[i]];
+                if (e && e.__frozen) {
+                    e.__frozen = false;
+                    e._unfreezeCallbacks();
+                    e.trigger("Unfreeze", e);
+                }
+            }
+        }
+        return this;
     }
 };
 
@@ -1138,7 +1225,7 @@ Crafty._callbackMethods = {
 
     // Process for running all callbacks for the given event
     _runCallbacks: function(event, data) {
-        if (!this._callbacks[event]) {
+        if (!this._callbacks[event] || this.__callbacksFrozen) {
             return;
         }
         var callbacks = this._callbacks[event];
@@ -1186,6 +1273,7 @@ Crafty._callbackMethods = {
     // Completely all callbacks for every event, such as on object destruction
     _unbindAll: function() {
         if (!this._callbacks) return;
+        this.__callbacksFrozen = false;
         for (var event in this._callbacks) {
             if (this._callbacks[event]) {
                 // Remove the normal way, in case we've got a nested loop
@@ -1194,6 +1282,30 @@ Crafty._callbackMethods = {
                 delete handlers[event][this[0]];
             }
         }
+    },
+
+    _freezeCallbacks: function() {
+        if (!this._callbacks) return;
+        for (var event in this._callbacks) {
+            if (this._callbacks[event]) {
+                // Remove the callbacks from the global list of handlers
+                delete handlers[event][this[0]];
+            }
+        }
+        // Mark this callback list as frozen
+        this.__callbacksFrozen = true;
+    },
+
+     _unfreezeCallbacks: function() {
+        if (!this._callbacks) return;
+        this.__callbacksFrozen = false;
+        for (var event in this._callbacks) {
+            if (this._callbacks[event]) {
+                // Add the callbacks back to the global list of handlers
+                 handlers[event][this[0]] = this._callbacks[event];
+            }
+        }
+        
     }
 };
 
--- a/src/spatial/spatial-grid.js
+++ b/src/spatial/spatial-grid.js
@@ -53,8 +53,9 @@
          * @comp Crafty.map
          * @kind Method
          * 
-         * @sign public Object Crafty.map.insert(Object obj)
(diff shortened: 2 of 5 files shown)


starwed:
>    Entities are still returned by the Crafty() selectors

After thinking about it, I think you actually want this.  If you can freeze entities by calling `Crafty("Octorok").freeze()`, you should be able to unfreeze them the same way.

>Since events are frozen, it might be possible for entities to get into inconsistent states. I need to make sure that this is handled in a friendly manner for existing entities.

The only specific issue I've thought of is that entities can be hooked together with `attach()`.  If the frozen entity is attached to a non-frozen entity, bad things can happen as the frozen entity is moved, but doesn't process the relevant events.  Not sure if we should specifically handle this, or just note in the docs that you should detach an entity before freezing it.

mucaho:
Good idea.   

> Since events are frozen, it might be possible for entities to get into inconsistent states. I need to make sure that this is handled in a friendly manner for existing entities. 

What about input control events? If entity gets frozen while key is pressed it down, could it continue to move forever? However, user controlled events aren't wired to many entities, especially if there are contained in a large entity pool (e.g. bullet entities). 

> Will not be rendered by a graphics layer (in the current implementation, it is still considered attached to the layer itself). Is removed from the spatial hash map

Have you by any chance measured the performance benefit of recycling entities vs creating / destroying new ones?

starwed:
>What about input control events? If entity gets frozen while key is pressed it down, could it continue to move forever? 

That's a good catch -- we should at least make an attempt at handling this gracefully. I think that having pulled the input state out of components and into a system will help here, but the "trigger" style events are still tied directly to the input triggers.

>Have you by any chance measured the performance benefit of recycling entities vs creating / destroying new ones?

Nope -- I was playing around with a kind of "rain drop" simulation, and it *seemed* like this approach improved the responsiveness, but such things can be pretty misleading.  I'll try to get a concrete measurement here.

starwed:
I did some basic benchmarks in Firefox, and it seems like the freeze approach does work best.


**What was tested**

Using a demo of a few hundred "raindrops" falling, I measured the FPS with the following three strategies for creating the raindrops:

- `RaindropFactoryReg`: Destroy and create raindrops as necessary
- `RaindropFactoryFakeFreeze`:  Instead of destroying them, make raindrops invisible,  move them off the screen to avoid collisions, and add them to an object pool for reuse.
- `RaindropFactoryFreeze`: Instead of destroying then, use the freeze method and add them to an object pool for reuse.

For the latter two strategies, there's a maximum size for the object pool, but I don't believe it's reached when letting the demo run at a steady state.

The timer strategy was set to `Crafty.timer.steptype("variable");`

**How it was tested**

For tests I found it was best to close and reopen Firefox to get a consistent state, then use the performance tab to measure the average FPS, letting it run until the average was "steady".  (A lot of times there's a perf hit directly after page load.)  I repeated this process 3 times for each of the strategies, and got the measurements below.

**Results**

- Reg: About 45 fps
- FakeFreeze: About 50 fps
- Freeze: about 56 fps

So it does look like this improves performance.

I then made a small tweak to how frozen objects were reinserted into the spatial graph, to reuse existing objects.  That might have improved the fps a little bit, but it was hard to tell.

**Code**
The test "game" I used is [here](http://ludonomicon.com/Rain/rain.html)  ([repo](https://github.com/starwed/starwed.github.io/tree/master/Rain)), with all three strategy implementations at the bottom.  There's a movable player object, but I avoided actually moving it during the tests.

src/graphics/dom-layer.js
```diff
@@ -64,6 +64,18 @@ Crafty._registerLayerTemplate("DOM", {
...
+    events: {
+        "Freeze":function(){
+            this._lastParentNode = this._div.parentNode;
+            this._div.parentNode.removeChild(this._div);
+        },
+        "Unfreeze":function(){
+            this._lastParentNode.appendChild(this._div);
+            delete this._lastParentNode;
+        }
+    },
```
mucaho: So there is a performance benefit to additionally remove the DOM element from the parent?
Note that the `DOM` component already sets `style.visibility = "hidden";`.

src/spatial/spatial-grid.js
```diff
@@ -66,13 +66,16 @@
...
-                entry = new Entry(keys, obj, this),
                 i = 0,
                 j,
                 hash;
-
+            if(!entry){
+                entry = new Entry(keys, obj, this);
+            } else {
+                HashMap.key(obj, entry.keys);
+            }
```
mucaho: Maybe simplify this to
```js
var keys = HashMap.key(obj, entry && entry.keys),
    entry = entry || new Entry(keys, obj, this)
    i = 0,
    j,
    hash;
```

src/core/core.js
```diff
@@ -1041,6 +1041,72 @@ Crafty.fn = Crafty.prototype = {
...
+     * @triggers Unfreeze - While the entity is being unfrozen
+     * 
+     * Unfreezes the entity, allowing it to receive events, inserting it back into the spatial map, 
+     * and restoring it to its previous visibility.
+     * 
+     * Triggers an "Unfreeze" event.
+     */
+    unfreeze: function () {
+        if (this.length === 1) {
+            this.trigger("Freeze", this);
```
mucaho: Why is there a "Freeze" event triggered here?

src/core/core.js
```diff
@@ -1041,6 +1041,72 @@ Crafty.fn = Crafty.prototype = {
...
+     * It is also removed from the spatial map, which means it will not be found by collisions, 
+     * raycasting, or similar functions.
+     * 
+     * Because the entity no longer listens to events, modifying its properties can result in an inconsistent state.
+     * 
+     * If custom components need to handle frozen entities, they can listen to the "Freeze" event, which will be triggered before the event system is disabled.
+     * 
+     * @see .unfreeze
+     */
+    freeze: function () {
```
mucaho: add the documentation about the event with `@trigger`
describe that freeze can be called on a group of entities with selector
add example for freezing group of entities
analogue for the `unfreeze` method

src/core/core.js
```diff
@@ -1194,6 +1261,32 @@ Crafty._callbackMethods = {
...
+        for (var event in this._callbacks) {
+            if (this._callbacks[event]) {
+                // Remove the callbacks from the global list of handlers
+                delete handlers[event][this[0]];
+            }
+        }
+        // Mark this callback list as frozen
+        // This property shows up in the list of callbacks, so it must be deleted 
+        // (not just set to false) when unfreezing.
+        this._callbacks.__frozen = true;
```
mucaho: Could this be a separate property, like `this.__callbacksAreFrozen` ?

src/spatial/spatial-grid.js
```diff
@@ -66,13 +66,16 @@
          * }
          * ~~~
          */
-        insert: function (obj) {
+        insert: function (obj, entry) {
```
mucaho: document entry parameter serves as entry object for reuse

mucaho:
Thanks for taking the time to benchmark this!   
Added couple of suggestions.

src/graphics/dom-layer.js
starwed: Yeah, looking at it now I don't think it's necessary.  I think I was recalling a past era when drawing was triggered through an event, rather than a direct method call.

src/spatial/spatial-grid.js
starwed: Make sense, and avoids creating a keys object unnecessarily.

src/core/core.js
starwed: Heh, didn't actually do anything, since the entity was frozen.  Deleted.

src/core/core.js
starwed: Seems reasonable.

starwed:
Updated with various PR suggestions.  
- Still need to consider how to handle inputs.   I'm not sure this is a blocking concern, since the main use case of this (object reuse) wouldn't really care too much.
- Still need to mention or handle attached entities.  I think just adding a flag to an entity that it's frozen, and then having `_cascade` skip frozen children would make the most sense.  (If `_cascade` triggered an event you'd get that for free, but there are probably performance wins from just manipulating the entities directly?)

mucaho:
Skimmed over it again, looks good.

> Still need to consider how to handle inputs. I'm not sure this is a blocking concern, since the main use case of this (object reuse) wouldn't really care too much.

For now I think this can be disregarded like you say. I'd inlcude a mention in the apidocs at most.  
And, yeah, that's what I was thinking about when proposing to make changes to the new control system. Controls are a system. They listen to input events or query pressed keys each frame, then link those triggers to methods. If controls were a component there would be two benefits:
1) You can freeze an entity it won't receive any input events. In the PR im working on I'll introduce entity-based keydown states, so once you remove input events, per-entity key states won't change. By eliminating both concerns you eliminate this issue (I hope).
2) Didn't yet try, but having a multiplayer game with both players using same keys could prove difficult.

> Still need to mention or handle attached entities. I think just adding a flag to an entity that it's frozen, and then having _cascade skip frozen children would make the most sense.

I think you [added these safeguards](https://github.com/craftyjs/Crafty/pull/1087/files#diff-453060f510aa3958ad510d399bb817c2R694). So PR is ready to be merged, right?

src/core/core.js
```diff
@@ -1041,6 +1041,93 @@ Crafty.fn = Crafty.prototype = {
...
+            delete this.__frozen;
+            this._unfreezeCallbacks();
+            this.trigger("Unfreeze", this);
+        } else {
+            for (var i = 0; i < this.length; i++) {
+                var e = entities[this[i]];
+                if (e) {
+                    delete e.__frozen;
+                    e._unfreezeCallbacks();
+                    this.trigger("Unfreeze", this);
```
mucaho: small typo, should be `e.trigger("Unfreeze", e)`

mucaho:
I was able to make some small changes to your PR directly! (Pretty neat feature):
* There was a [small typo](https://github.com/craftyjs/Crafty/pull/1087#discussion_r112823694)
* Made `(un)freeze` methods idempotent
* Added tests for "(Un)freeze" events
* e: I set flags to `false` instead of deleting them,  to [prevent possible deoptimizations](https://github.com/craftyjs/Crafty/wiki/Ideas#fix-possible-deoptimizations)

Merging

PR #1089, opened 2017-03-11
Create a 'Centered' component for assigning position by origin

starwed:
It's been a somewhat common request to be able to assign a position based on an entities origin, rather than it's top left corner.

This PR adds a `"Centered"` component that adds `cx` and `cy` properties for this, as well as a couple of helper methods: 
- `centerOrigin`  for moving the origin to the center of an entity
- `shiftAroundOrigin` for moving an entity while keeping its origin fixed to its current stage position.

It's possible that `cx` and `cy` are slightly misleading, since it's not necessarily the center of the entity.  Another option might be decouple the notion of center and origin, and use  `ox` and `oy` for the origin's position, while `cx` and `cy` provide the true center.

### DIFF
--- a/src/spatial/2d.js
+++ b/src/spatial/2d.js
@@ -770,6 +770,8 @@ Crafty.c("2D", {
      *
      * Set the origin point of an entity for it to rotate around.
      *
+     * @triggers OriginSet -- after the new origin is assigned
+     * 
      * @example
      * ~~~
      * this.origin("top left")
@@ -814,7 +816,7 @@ Crafty.c("2D", {
 
         this._origin.x = x;
         this._origin.y = y;
-
+        this.trigger("OriginChanged");
         return this;
     },
 
--- a/src/spatial/centered.js
+++ b/src/spatial/centered.js
@@ -1,0 +1,121 @@
+var Crafty = require('../core/core.js');
+
+/**@
+ * #Centered
+ * @category 2D
+ * @kind Component
+ * 
+ * A component for setting an entities position by its origin, rather than the top left corner.
+ * 
+ * Setting `cx` and `cy` will move the entity such that its origin is at that location. 
+ * This will fire the same events that setting `x` and `y` directly would.
+ * 
+ * The origin of an entity defaults to the top left corner.  To set it explicitly, call the `origin()` method
+ * of the 2D component.  To center the origin, you can also use the `centerOrigin()` method of this component.
+ * 
+ * @trigger Move - when the entity has moved - { _x:Number, _y:Number, _w:Number, _h:Number } - Old position
+ * @trigger Invalidate - when the entity needs to be redrawn
+ * 
+ * @see 2D#.origin
+ */
+Crafty.c("Centered", {
+    properties: {
+        /**@
+         * #.cx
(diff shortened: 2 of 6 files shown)


src/spatial/2d.js
```diff
@@ -770,6 +770,8 @@ Crafty.c("2D", {
      *
      * Set the origin point of an entity for it to rotate around.
      *
+     * @triggers OriginSet -- after the new origin is assigned
```
mucaho, line 773: documented eventName doesn't match event triggered in code

src/spatial/centered.js
```diff
@@ -0,0 +1,121 @@
...
+     * @comp Centered
+     * @kind Method
+     * 
+     * @sign public this .shiftAroundOrigin(dx, dy)
+     * @param dx - Shift the entity by this amount along the x axis
+     * @param dy - Shift the entity by this amount along the y axis
+     *
+     * Moves the entity while keeping the origin fixed on the stage.
+     */
+    shiftAroundOrigin: function(dx, dy) {
```
mucaho, line 111: I would remove the additional indirection from this code and just do something like
```js
this.origin(this._origin.x - dx, this._origin.y - dy)
thix.x += dx;
this.y += dy;
```
directly.

mucaho (commented):
Couple of suggestions:
* I like `ox` and `oy` better.
* Unless I don't fully understand and they refer to different concepts there is no need to introduce the "center" keyword at all, `origin` should suffice.
* I would push this functionality into the `2D` component itself and not in a separate component.
* `centerOrigin` is redundant. We already have `origin('center')`. Rounding feature can be added to that method as an additional argument.
* Would suggest considering `parent` when setting `ox`, `oy`, so that they set the coordinates of the entity relative to its parent and the parent's origin, in addition to considering own origin. Something like (I'm not sure if the math checks out but you get the gist :) ) 
```js
        cx: {
            set: function (v) {
                var x = v - this._origin.x;
                var parent;
                if ((parent = this._parent))
                  x += parent.x + parent._origin.x;
                this._setter2d('_x', x);
            },
            get: function () {
                var x = this._x + this._origin.x;
                var parent;
                if ((parent = this._parent))
                  x -= parent.x + parent._origin.x;
                return x;
        },
```

mucaho:
@starwed Any updates on this? Would be nice to include this.

starwed:
>I would push this functionality into the 2D component itself and not in a separate component.

Makes sense, especially if I just remove the helper methods and stick to `ox` and `oy` setter/getters.  I'll simplify this PR to just add those properties.

>Would suggest considering parent when setting ox, oy

Generally, when you set the position of a child entity, it doesn't pay any attention to it's parent's coordinates.  Is there a reason you think this would be expected behavior here?

starwed:
Enough changed that I'm opening a new PR.

mucaho:
> Generally, when you set the position of a child entity, it doesn't pay any attention to it's parent's coordinates. Is there a reason you think this would be expected behavior here?

Good observation. Easy workaround to get coordinates relative to parent origin space, if needed:
```js
var localX = ent.ox - ent.parent.ox
```

Issue #1076, opened 2017-01-16
Create sprite animations by sprite names


matthijsgroen:
Currently sprite animations can only be done by indices.
It is more flexible to support sprite names, e.g. if you generate spritemaps using spritesheet.js

starwed:
Sounds reasonable to me, we just need to make sure it doesn't interfere with other use cases.

PR #1091, opened 2017-03-12
animation by sprite names #1076

maciekwie:
I've changed the way `Sprite.sprite()` is called in  `._updateSprite()` method so `reel.frames` can be list of sprite names.

I've added `.reelFrame()` method for setting frame with specified sprite.

// It's my first pull request

### DIFF
--- a/playgrounds/animation/sprite-animation.html
+++ b/playgrounds/animation/sprite-animation.html
@@ -1,38 +1,38 @@
 <!DOCTYPE HTML>
 <html>
-	<head>
-		<script src="../../tests/lib/jquery.min.js"></script>
-		<script type="text/javascript" src="../../crafty.js"></script>
-	</head>
-	<body>
-		<h1>Sprite animation playground:</h1>
-		<div id="cr-stage"></div>
-		<script type="text/javascript" src="sprite-animation.js"></script>
-		<div>
-			<button onclick="playgroundPlay();">Play</button>
-			<label for="playReelId">Reel ID:</label>
-			<select id="playReelId">
-				<option value="count" selected>Count</option>
-				<option value="countEven">Count Even</option>
-                                <option value="countEven_names">Count Even (frame names)</option>
-				<option value="short">Short</option>
-				<option value="countSlow">Count Slow</option>
-			</select>
-			<label for="playRepeatCount">Loops:</label>
-			<input type="number" id="playRepeatCount"></input>
-		</div>
-		<div>
-			<button onclick="playgroundPosition();">Jump to</button>
-			<input type="number" id="reelPosition"></input>
-		</div>
-                <div>
-			<button onclick="playgroundFrame();">Jump to by name</button>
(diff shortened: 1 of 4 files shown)


src/graphics/sprite-animation.js
```diff
@@ -406,7 +406,34 @@ Crafty.c("SpriteAnimation", {
...
+     * @sign public this .reelFrame(String frameName)
+     * Sets the position of the current reel by frame name.
+     * @param frameName - Name in the sprite map.
+     * 
+     * Jumps to specifed frame if the reel was created with sprite names.
+     * 
+     */
+    reelFrame: function (frameName) {
+        if (this._currentReel === null)
+            throw("No active reel.");
```
starwed: Could you either put this (and a couple of similar clauses) either in a single line, or explicitly wrap the block in brackets?  The standard approach is 
 
```
if (condition) {
  doSingleThing();
}
```

The idea here is that by omitting brackets, you tempt someone later to do this:
```
if(condition)
   doSomething();
   doSomethingElse();
```
which will not work as expected.  If you have a very short statement, it's also ok to use a single line to make the code a little less noisy, since it doesn't tempt folk the way an indented block does.  :)
```
if (condition) doSomething();
```

src/graphics/sprite-animation.js
```diff
@@ -442,9 +465,13 @@ Crafty.c("SpriteAnimation", {
...
-        var pos = currentReel.frames[currentReel.currentFrame];
-        this.sprite(pos[0], pos[1]); // .sprite will trigger redraw
-
+        var frame = currentReel.frames[currentReel.currentFrame];
+        
+        if(typeof frame === "string") {
+            this.sprite(frame);
+        }
+        else
+            this.sprite(frame[0], frame[1]); // .sprite will trigger redraw
```
starwed: Another block needing brackets.

playgrounds/animation/sprite-animation.html
```diff
@@ -14,6 +14,7 @@
 			<select id="playReelId">
 				<option value="count" selected>Count</option>
 				<option value="countEven">Count Even</option>
+                                <option value="countEven_names">Count Even (frame names)</option>
```
starwed: Tiny nit: the indentation here is off.

starwed (changes requested):
The code looks good, and thanks for adding tests!

I added some notes for a couple of small style issues, but once those are fixed this is good to merge.

maciekwie:
Thank you for feedback!

I've changed conditions according to this rule in the whole file, for consistency.

In `sprite-animation.html` indentation was made with two tabs instead of two spaces, now it's fixed.

starwed (approved):

starwed:
Merged into develop -- thanks for the PR!  (And congrats on your first.)

(I rebased locally to squash down to a single commit, so closing the github PR -- it is merged, though.)

PR #1096, opened 2017-03-20
Allow properties to be defined as part of the component definition

starwed:
Add a very simple shortcut for using `Object.defineProperty` when adding a component.  This replaces the existing setup that "2D" and "Renderable" used.

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -292,6 +292,13 @@ Crafty.fn = Crafty.prototype = {
             if (comp && "required" in comp) {
                 this.requires( comp.required );
             }
+            // Define properties
+            if (comp && "properties" in comp) {
+                var props = comp.properties;
+                for (var propertyName in props) {
+                    Object.defineProperty(this, propertyName, props[propertyName]);
+                }
+            }
             // Call constructor function
             if (comp && "init" in comp) {
                 comp.init.call(this);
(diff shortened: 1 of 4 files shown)


matthijsgroen:
@starwed This is a really great feature, but there has not been a release for a long while... when is an new release planned? I currently use Crafty in a webpack setup (no bower) and now the nightlies won't work for me. Any update on this?

starwed:
>when is an new release planned?

Essentially whenever I have time.  :)  Been a bit busy with work/life stuff lately, but I think what's in `develop` right now should be mostly good to ship as 0.9.0.  There were some breaking changes around the movement related events I want to go back and look at, to make sure I didn't break some use cases.

(And I'll be glad to help out anyone else who has the time to spearhead a release, of course!)

> I currently use Crafty in a webpack setup (no bower) and now the nightlies won't work for me.

I noticed bower itself now suggests that it's deprecated, so perhaps we can get rid of that in favor of supporting webpack.  Is it a bug in Crafty that's causing nightlies to fail?

matthijsgroen:
could be.

I have in my *package.json* :

`"Crafty-Distro": "craftyjs/Crafty-Distro#nightlies"`

I have a `crafty-loader`:

```
import "Crafty-Distro/crafty";
const Crafty = window.Crafty;
delete window["Crafty"];
export default Crafty;
```

And I get this error in the browser:

```
Uncaught TypeError: Cannot convert undefined or null to object
    at init.removeComponent (crafty.js:2699)
    at init.remove (ControlScheme.coffee:8)
    at init.removeComponent (crafty.js:2690)
    at init.reset (Player.coffee:9)
    at init (Player.coffee:5)
    at init.addComponent (crafty.js:2562)
    at Function.e (crafty.js:4066)
    at Object.start (Game.coffee:60)
    at Object.<anonymous> (index.js:15)
    at Object../src/index.js (game.c703b5a08c27f00018cf.js:67336)
```

the line of code that fails within Crafty looks like this: (`init.removeComponent (crafty.js:2699)`)

```
// update map from component to (entityId -> entity)
delete compEntities[id][this[0]];
```

With the 0.8.0 version this works fine.

@starwed 
I could make a new issue for this, but should that be on this repo or the Crafty-Distro one?

PR #1103, opened 2017-04-04
Delete obsolete version file.

Sebring:
Version 0.8.0 since https://github.com/craftyjs/Crafty/commit/6998fc9e5dd81ac253023ad85232404c696484e8

### DIFF
--- a/src/version.js
+++ b/src/version.js
@@ -1,1 +1,1 @@
-module.exports = "0.7.0";
+module.exports = "0.8.0";
--- a/src/version.js
+++ b/src/version.js
@@ -1,1 +1,0 @@
-module.exports = "0.8.0";


mucaho:
This version file shouldn't exist any more. `src/core/version.js` is used instead.    
While we're at it I'd also gitignore `.vscode`.
Another duplication can be found with `supported-browsers-webdriver.json` and `supported-browsers.json`

Sebring:
Ah I see. This file caused me some confusion when checking out latest form release-branch and thinking it was running 0.7.0 still.  I'll just update this PR to delete the file instead.

mucaho:
Thanks! Merged via 1b6614acd1facc8e2eadd80f3d93265d2033b81b

PR #1104, opened 2017-04-09
Organize unit tests

mucaho:
This is a first in a series of PRs that will update our testing infrastructure.
This PR organizes test files into appropriate subdirectories.

Next PR will update qunit version and all unit tests that run in browser in node:
* All test commands are no longer available on global scope, but rather on `QUnit` itself.
* Assertion methods are called on an argument passed into the test functions.


starwed:
Looks good to me -- I confirmed that the tests still ran from the grunt commands and `index.html`, and that the number of tests matched between this branch and develop.

PR #1105, opened 2017-04-09
Remove code duplication

### LINKED ISSUE #1097: crafty-headless uses cached modules
Replace `require` with `requireNew` statements. Multiple headless instances will share state.

mucaho:
Extract common source modules and test files into crafty-common.js and index-common.js respectively.   
No more copy-pasting between those two.
Additionally, fixes #1097.

### DIFF
--- a/src/crafty-common.js
+++ b/src/crafty-common.js
@@ -1,0 +1,34 @@
+// Define common features available in both browser and node
+module.exports = function(requireNew) {
+    if (requireNew) {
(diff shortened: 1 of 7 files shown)


mucaho:
I think that the goal of this PR is straightforward. If there are pointers regarding the implementation, I will gladly adjust it, let me know!   
Merging this in, since it's blocking the test file reorganization.

PR #1107, opened 2017-04-10
Update qunit tests

mucaho:
Follow-up of #1104.
See commits for detailed explanation.

Tests are functionally equivalent to before, except that I converted all `equal` assertions to `strictEqual` ones and all `notStrictEqual` to `notEqual` ones. This should strengthen the assertions in theory, in practice it had no effect whatsoever.

Another thing that could be changed is the identifier for the new `assert` argument that gets passed in each QUnit.test. Global `ok`, `equal`, etc. methods are gone, they have to be called on this `assert` argument.
In order to enhance readability I opted to use `_` instead.
So instead of
```js
test("common color names", function(assert){
  var c = {};
  Crafty.assignColor("red", c);
  assert.strictEqual(c._red, 255, "red is 255");
  assert.strictEqual(c._green, 0, "green is 0");
  assert.strictEqual(c._blue, 0, "blue is 0");
  assert.strictEqual(c._strength, 1, "strength is 1.0");
});
```
we have
```js
test("common color names", function(_){
  var c = {};
  Crafty.assignColor("red", c);
  _.strictEqual(c._red, 255, "red is 255");
  _.strictEqual(c._green, 0, "green is 0");
  _.strictEqual(c._blue, 0, "blue is 0");
  _.strictEqual(c._strength, 1, "strength is 1.0");
});
```
I can change this back if someone has a preference about this.

After this PR lands you have to reinstall `node_modules` via `npm install` or `yarn install`.


mucaho:
Merging this non-functional change in now, since there is one more thing to do to reorganize the tests: Reorganize the test modules. We have e.g. `collision` tests in `2d.js`, although there already exists a `collision.js` file. Next commit will take care of that and that should be the final major change to unit test files.    

If someone has a preference about the identifier mentioned above, this can still be changed, it's as simple as a text replacement in all test files. Let me know.

PR #1036, opened 2016-05-09
Add appveyor integration

mucaho:
Check if Crafty builds and passes tests on Win platforms.
This removes worries of failing native development dependencies (e.g. `phantomjs`) across platforms.
[Tested on my own fork](https://ci.appveyor.com/project/mucaho/crafty) that everything works correctly.
- [x] Add [appveyor.yml](https://www.appveyor.com/docs/appveyor-yml)
- [x] [Register](https://www.appveyor.com/docs#step-2-add-your-project) the craftyjs/Crafty repository with appveyor
- [x] Update documentation in [wiki:Automatic-Build-Testing](https://github.com/craftyjs/Crafty/wiki/Automatic-Build-Testing)
- [x] Add [status badge](https://www.appveyor.com/docs/status-badges#badges-for-projects-with-public-repositories-on-github-and-bitbucket)
- [ ] Optionally protect develop branch by from merges with failing appveyor status checks

PR #1108, opened 2017-04-11
Meta fixes

mucaho:
Small improvements, see commits.

Among other things, after this PR lands, tests should run with Crafty paused by default. It needs to be unpaused via test hooks, if a specific set of tests needs it.


mucaho:
Additionally added a commit that tell Travis to build Crafty on Mac OSX too. It works!    
The only downside is that builds go from ~3 min wait time to ~30 min wait time, as osx instances are limited. Maybe this feature should be pushed to run only on `testing` branch?

If we enable #1036 Crafty builds and tests on linux, os x, and win! So we can be sure that anyone can contribute no matter what platform (s)he codes on.

EDIT: OSX build is too slow and didn't find a way to disable starting the virtual machine depending on which branch is being tested. Removed the relevant commit from this PR, and will investigate further.

mucaho:
I'm very happy that we included appveyor! It showed a major flaw in our test code: Since webdriver tests were introduced the tests failed on every windows PC! 😢 . Fixed the issue now, sorry for the inconvenience for anybody that tried to build it on windows.   
I think these changes are straightforward, if there are change suggestions, I will happily oblige! Merging.

PR #1109, opened 2017-04-13
Update webdriver

mucaho:
Updates the webdriver dependencies and organizes webdriver test files.
first part of fixing #1078


mucaho:
(just wanted to make sure it builds on travis before merging)
(travis doesn't work on my fork since sauce credentials are missing and the sauce_connect plugin fails :( )

PR #1113, opened 2017-04-14
Fix for fully cancelled tween still firing TweenEnd

### LINKED ISSUE #966: Problem with cancelTween()
I think I found a bug in the Tween component. I'm using the latest nightly.

What happens is:
- I start a tween of say 1 sec.
- I cancel the tween at 0.5sec.
- I start another tween at 0.7sec.
- The first tween isn't fully cancelled! (e.g. still in .tweens array)

It seems the full cancellation happens only after the time the tween would have ended originally. This effect however screws up my game logic since I use tween loops and rely on correct order of these things.

Unfortunately I don't have a small self-contained example, but maybe it is easy to find the bug without it.

mucaho:
Fixes #966 , with solution suggested by airza.
Adds more tests.

### DIFF
--- a/src/core/tween.js
+++ b/src/core/tween.js
@@ -156,9 +156,11 @@ module.exports = {
   * Stops tweening the specified group of properties, and fires the "TweenEnd" event.
   */
   _endTween: function(properties){
+    var notEmpty = false;
     for (var propname in properties){
+      notEmpty = true;
       delete this.tweenGroup[propname];
     }
-    this.trigger("TweenEnd", properties);
+    if (notEmpty) this.trigger("TweenEnd", properties);
   }
 };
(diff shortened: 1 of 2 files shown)


src/core/tween.js
```diff
@@ -156,9 +156,11 @@ module.exports = {
   * Stops tweening the specified group of properties, and fires the "TweenEnd" event.
   */
   _endTween: function(properties){
+    var notEmpty = false;
     for (var propname in properties){
+      notEmpty = true;
       delete this.tweenGroup[propname];
     }
-    this.trigger("TweenEnd", properties);
+    if (notEmpty) this.trigger("TweenEnd", properties);
```
starwed, line 164: 👍  I think this approach is preferable to something like using `object.keys()`, since we're already iterating through the properties.

starwed (approved):
Looks good, thanks for adding more tests.

PR #1115, opened 2017-04-15
Fix mouse scroll

mucaho:
Adds mouseWheel event listeners for new browsers.
Makes viewport.mouseLook respect zoom factor.
See the [craftyjs/demos/devDemos/example_maplike_zoom.html](https://github.com/craftyjs/demos/blob/master/devDemos/example_maplike_zoom.html).

You can test the functionality by downloading the following file and renaming it to `crafty.js` and including it into the example file.
crafty.js

EDIT: Tested on new firefox, edge and chrome versions, seems to work as expected.

### DIFF
--- a/src/controls/inputs.js
+++ b/src/controls/inputs.js
@@ -450,7 +450,7 @@ Crafty.extend({
      */
     findPointerEventTargetByComponent: function (comp, e, target) {
         var tar = target ? target : Crafty.stage.elem,
-            closest, current, q, l, i, pos, layerPos, maxz = -Infinity;
+            closest, current, q, l, i, pos, maxz = -Infinity;
         var x = e.clientX;
         var y = e.clientY;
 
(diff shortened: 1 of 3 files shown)


src/controls/inputs.js
```diff
@@ -492,21 +491,18 @@ Crafty.extend({
...
-        if (!layerPos) {
-            layerPos = Crafty.domHelper.translate(x, y);
-        }
+        // Find the Crafty position in the default coordinate set,
+        // disregard the fact that the pointer event was related to a specific layer.
+        pos = Crafty.domHelper.translate(x, y);
 
         // Update the event coordinates and return the event target
-        e.realX = layerPos.x;
-        e.realY = layerPos.y;
```
starwed, line 509: Huh, I feel like there was a specific reason this was necessary, but looking through the code I can't find uses anywhere but the drag code.   This is definitely simpler, so if it works 👍

starwed (approved):

src/controls/inputs.js
mucaho: Great, thanks for taking a look at it! ❤️  Wish you a happy Easter weekend! See also comments in #1024.

mucaho:
follow-up PR will try to streamline input handling, in order to fix some issues we have

Issue #1013, opened 2016-03-03
Viewport's mouselook conflicts with other Mouse entities


naasking:
Here's the sample showing the problem: https://jsfiddle.net/8va2w9r4/

This might be a little tricky to explain. Click and hold to grab the viewport right next to the shape, then quickly move the mouse over the shape. While you're over the shape, the viewport stops panning with the mouse movement, and if you let go of the mouse button and move the mouse off the shape, the viewport still pans as if you had the button pressed. You can click anywhere on the background to release of course.

May not seem like a big deal, but I've actually gotten into scenarios with fast mouse dragging where Crafty thinks I grabbed the viewport to pan while the mouse is over a shape. So everytime I move off the shape, the shape ends up back under the mouse so I can _never_ release the viewport pan behaviour.

To get out of it, I have to move the mouse super-quick off the shape and click before it follows and ends up under the mouse again. I hope that's clear.

I have no opinion on whether all mouse move events should be respected while over the shape while in mouselook mode (they currently aren't), but I definitely think mouseup should be handled as if you weren't over the shape.

200sc:
(This happens even without Draggable, but just with Mouse on the red cube).

So the mouseUp event is getting consumed by the cube and isn't being passed down back to the viewport. Is it possible to always pass down mouseUp events to the viewport, assuming mouselook is the only thing the viewport will care about with respect to mouse events?

It looks like the viewport is using [Crafty.mouseObjs++](https://github.com/craftyjs/Crafty/blob/develop/src/graphics/viewport.js#L448) to tell Crafty that it wants to receive events.

Mouse events are handled [here](https://github.com/craftyjs/Crafty/blob/develop/src/controls/inputs.js#L159), and mouselook is already special cased out-- this would be a solution given the current architecture, but it might be nice to pull out the active boolean mouselook uses to ignore this call (when it isn't active) and expose it to the mouse event handler?

```
if (closest) {
            //click must mousedown and out on tile
            if (type === "mousedown") {
                closest.trigger("MouseDown", e);
            } else if (type === "mouseup") {
                Crafty.viewport.mouselook('stop');
                closest.trigger("MouseUp", e);
            } else if (type == "dblclick") {
                closest.trigger("DoubleClick", e);
            } else if (type == "click") {
                closest.trigger("Click", e);
            } else if (type === "mousemove") {
            ...
```

starwed:
[...]

PR #1116, opened 2017-04-16
Streamline input implementation

### LINKED ISSUE #1043: Click/Touch doesn't work on mobile via PhoneGap with 0.7.1, works with 0.4.2
Hi,

I crafted a very, very simple minimal CraftyJS game -- something like (minus boiler-plate code):

```
Crafty.e('2D, Alpha, Color, Canvas, Mouse, Touch').attr({ "w": 48, "h": 48 }).bind('Click', function() { Crafty.audio.play(someId) });
```

I tried swapping `Click` with `Touch`. It works fine on the web, but when I shove it into a PhoneGap container and deploy it on my phone (OnePlus X), the click handler never fires.  I even added a ".color('blue")` to make sure that it's not just that audio's broken on my phone.

Is this something that's supposed to work, or is it something that's not supported?

I'm using CraftyJS 0.7.1 with PhoneGap 6.3.0 (I think)

mucaho:
EDIT:
**Current impact: **
* fixes _"Someone could hold down a key while we've lost focus, and then release once it's back, so we probably don't want to fire keyup events in such a case."_ part of #552 
* fixes all points of #1013 
* tries to fix #1043 , needs testing from a friendly soul with access to that :)
* **API change** of `findPointerTargetEventByComponent`, and all documented for clarity, but only internally used `dispatch` methods

### DIFF
--- a/src/controls/controls-system.js
+++ b/src/controls/controls-system.js
@@ -10,16 +10,12 @@ var Crafty = require('../core/core.js');
 
 // MouseButtonToggleInput
 function MouseButtonToggleInput(button) {
-    Crafty.mouseObjs++;
     this.button = button;
 }
 
 MouseButtonToggleInput.prototype = {
     isDown: function() {
         return Crafty.mouseButtonsDown[this.button];
-    },
-    destroy: function() {
-        Crafty.mouseObjs--;
     }
 };
 
@@ -89,6 +85,12 @@ Crafty.s("Controls", {
         // internal object to store definitions
         this._dpads = {};
         this._triggers = {};
+
+        // listen to mouse events
+        Crafty.s("Mouse").bind("MouseDown", this.updateTriggers.bind(this));
+        Crafty.s("Mouse").bind("MouseUp", this.updateTriggers.bind(this));
+        Crafty.s("Mouse").bind("DoubleClick", this.updateTriggers.bind(this));
+        Crafty.s("Mouse").bind("Click", this.updateTriggers.bind(this));
     },
 
     events: {
--- a/src/controls/controls.js
+++ b/src/controls/controls.js
@@ -1,6 +1,278 @@
 var Crafty = require('../core/core.js');
 
 /**@
+ * #KeyboardState
+ * @category Input
+ * @kind Component
+ *
+ * Provides entity with keyboard events.
+ * @note This component requires the `Keyboard` component to be added to the entity, otherwise key events will not be triggered on this entity.
+ *
+ * @trigger KeyDown - is triggered for each entity when the DOM 'keydown' event is triggered. - { key: `Crafty.keys` keyCode (Number), originalEvent: original KeyboardEvent } - Crafty's KeyboardEvent
+ * @trigger KeyUp - is triggered for each entity when the DOM 'keyup' event is triggered. - { key: `Crafty.keys` keyCode (Number), originalEvent: original KeyboardEvent } - Crafty's KeyboardEvent
+ *
+ * The event callbacks are triggered with a native [`KeyboardEvent`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent), accessible by the `originalEvent` property.
+ *
+ * In addition to binding to these events, the current state (pressed/released) of a key can also be queried using the `.isDown` method.
+ *
+ * @example
+ * ~~~
+ * Crafty.e("2D, DOM, Color, Keyboard")
+ *   .attr({x: 100, y: 100, w: 50, h: 50})
+ *   .color("red")
+ *   .bind('KeyDown', function(e) {
+ *     if (e.key == Crafty.keys.LEFT_ARROW) {
+ *       this.x = this.x-1;
+ *     } else if (e.key == Crafty.keys.RIGHT_ARROW) {
+ *       this.x = this.x+1;
+ *     } else if (e.key == Crafty.keys.UP_ARROW) {
+ *       this.y = this.y-1;
+ *     } else if (e.key == Crafty.keys.DOWN_ARROW) {
+ *       this.y = this.y+1;
+ *     }
+ *   });
+ * ~~~
+ *
+ * @see Keyboard, KeyboardSystem
+ * @see Crafty.keys
+ * @see Crafty.keydown
+ * @see Crafty.keyboardDispatch
+ */
+Crafty.__keyboardStateTemplate = {
+    _keyDown: null,
+
+    init: function() {
+        this._keyDown = {};
+    },
+
+    /**@
+     * #.isDown
+     * @comp KeyboardState
+     * @kind Method
+     *
+     * @sign public Boolean isDown(String keyName)
+     * @param keyName - Name of the key to check. See `Crafty.keys`.
+     * @returns The pressed state of the key
+     *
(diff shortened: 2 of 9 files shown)


starwed:
>Route all mouse events without target to the global "Mouse" system,
which controls and viewport can bind to.

I wonder if it would make sense to route *all* events to such a system, with information about whether there was a target or not.

Aside: if this ends up changing the api in a backwards-incompatible way, let's make sure to prep another point release before merging.

mucaho:
> I wonder if it would make sense to route *all* events to such a system, with information about whether there was a target or not.

Sure, but what would be the benefits?

> Aside: if this ends up changing the api in a backwards-incompatible way, let's make sure to prep another point release before merging.

EDIT: Initially I thought there wouldn't be many breaking changes, but alas, there are :)

mucaho:
PR is ready, so starwed if you'd like to take a look at it, it would be greatly appreciated! If you are going to do so, you could omit whitespace with `?w=1` in the diff url. Commit messages should describe in detail all changes. See [current impact](https://github.com/craftyjs/Crafty/pull/1116#issue-221991996), to summarize it's a small API change that fixes some open issues. What do you think, would it make sense to include these fixes in next minor release, too?

What's planned in next PR, will modify API and semantics considerably.
* Split everything into appropriate files. Dispatchers into own file. Mouse & Keyboard system and components into own file. Event setup into own file.
* Let `Crafty.keydown`, `Crafty.resetKeyDown`, `Crafty.lastEvent` live under the appropriate dispatcher namespace. So the whole logic is condensed in each dispatcher.
* Introduce `lastKeyEvent` to KeyboardState and repurpose `Crafty.lastEvent` as `lastMouseEvent` into MouseState
* Additionally, convert the dispatcher itself to a system and let it inherit from MouseState, KeyState. The dispatcher will trigger all events on itself, no matter where it routes them afterwards. This will allow users to always be notified on newest input events.
* Motivation: This "layered" architecture is needed, because we can have the following situation:
  * A KeyDown has been dispatched globally. We can't know which entity modified its state based on this.
  * Another entity requires the lock.
  * We now need to dispatch a KeyUp event globally, since the initial entity won't receive a KeyUp event until the other entity released the lock. It would be stuck in an inconsistent state.
* Trigger Keyboard events only on entities with the Keyboard component and on the Keyboard system. (depends on #1117). Since we need to dispatch artificial KeyUp events (see previous point about motivation), we have to make sure not to cause inconsistent state. Thus only entities with KeyboardState component will receive it, in order to shield them from state corruption.
* Suggestions are welcome!

After that PR, figure out how to make the control system and components get their input from the entity-specific keyboard and mouse states.

starwed:
>Trigger Keyboard events only on entities with the Keyboard component and on the Keyboard system. (depends on #1117). Since we need to dispatch artificial KeyUp events (see previous point about motivation), we have to make sure not to cause inconsistent state. Thus only entities with KeyboardState component will receive it, in order to shield them from state corruption.

The main motivation of the control system I added was to move away from entities listening directly to keyboard/mouse/etc events.

starwed:
Oops, accidentally hit close!  

>Trigger Keyboard events only on entities with the Keyboard component and on the Keyboard system. (depends on #1117). Since we need to dispatch artificial KeyUp events (see previous point about motivation), we have to make sure not to cause inconsistent state. Thus only entities with KeyboardState component will receive it, in order to shield them from state corruption.

The main motivation of the control system I added was to move away from entities listening directly to keyboard/mouse/etc events.  A big part of that was avoiding things like artificial KeyUp events; you'd have the control system dispatch "TriggerUp" events (etc) instead.

mucaho:
> The main motivation of the control system I added was to move away from entities listening directly to keyboard/mouse/etc events. A big part of that was avoiding things like artificial KeyUp events; you'd have the control system dispatch "TriggerUp" events (etc) instead.

Right, but that only takes care of control components. (Although "only" in this case covers probably to the majority of use-cases). Users may want to listen to "KeyDown" and "KeyUp" events for their own implementation, with this proposal it would be safe to do so.

starwed:
I'd argue that the concept of locking events to a specific entity should happen in the control layer not the input layer.  That way you *can* opt in to getting the raw events, which will be as close a representation to the actual input as we can get.

Of course, we don't actually run pointer events through the control system right now...  :)

If we were to move towards that, the rough idea is this sort of structure:
- The input layer handles normalizing how browsers pass in the key/pointer/etc events.  This is pretty much all the existing [keyboardDispatch](https://github.com/craftyjs/Crafty/blob/develop/src/controls/inputs.js#L634-L674) method does right now, but mouseDispatch is doing a bit more.  In my idealized version, it just produces a single event that passes through the crafty stage coordinates, normalized button information, and the original dom target.  Any logic should be happening at a later stage.
- The control layer would then handle interpreting the normalized data and conveying it to entities.  For pointer events, I think that would include finding the target of the event, which is part of dispatch right now, as well as any notion of locking.

An advantage of this approach is that it should be easier to add support for multiple types of input mapping to the same control behavior.  (For instance, extending the current viewport mouselook to be controllable from the keyboard.)

I'm not 100% convinced that my way is right, (and there could be big obstacles I'm not seeing, e.g. adding the notion of locking to the current control system) but I *am* certain we shouldn't work in opposite directions on this.  :)  I'd feel more comfortable if we hashed out the preferred approach before adding any additional public apis.  What are your thoughts?

mucaho:
Thanks for taking the time to think this through!

----------

> I'd argue that the concept of locking events to a specific entity should happen in the control layer not the input layer. That way you can opt in to getting the raw events, which will be as close a representation to the actual input as we can get.

This case is handled by 

> The dispatcher will trigger all events on itself, no matter where it routes them afterwards _[or if something is locked currently]_. This will allow users to always be notified on newest _[raw]_ input events.

-----------------------------------------------------

> An advantage of this approach is that it should be easier to add support for multiple types of input mapping to the same control behavior. (For instance, extending the current viewport mouselook to be controllable from the keyboard.)

Seems like a good idea! Adding an additional layer of indirection on the input events shouldn't conflict with this. On the contrary, it should serve as a foundation for the control system. I firmly believe that the input system itself should filter out invalid events, even if it only serves to let users have some other safe implementation outside the control system.

-------------------------------------------

> The control layer would then handle _[conveying data]_ to entities. For pointer events, I think that would include finding the target of the event, which is part of dispatch right now, as well as any notion of locking.

Now this point is interesting, and I agree that routing and locking of events could be handled by the logic of the control system. However, for this to work the rest of current control components (Draggable and mouselook) need to be adapted to use the control system.

-----------------------------------------------

Are you content to go forward with the following approach?
* Dispatch all mouse events only to the Mouse system and Mouse components
* Dispatch all key events only to the Keyboard system and Keyboard components.
* These systems and components would trigger only valid input events (due to the validation of key / mouse state)
* Control system would either listen to the appropriate system or component and then do its routing based on its control logic

If you are ok with this, i would remove the locking and routing behavior from the input system, but keep the rest of this PR intact.

mucaho:
Thought about it a little bit more: if routing and locking is shifted to the controls there is no need for key state validation on the level of individual components. It suffices if Keyboard dispatch takes care of that on the global level and if it calls resetKeyDown when needed.
There may be some fixes that could be extracted from here, but other than it's obsolete

Issue #1106, opened 2017-04-09
Deprecate older browsers?


starwed:
Right now we support relatively old versions of browsers such as IE9.  But there are newer features in ES6 that might be nice to use.  

As an example, there are several places where we could use [Sets](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) instead of objects, perhaps with a performance gain.  (This was prompted specifically by looking at how we handle spatial searches; right now we use a new object each time to check for dupes, but a Set object would be a more natural fit, and is easier to reuse from search to search.)

Essentially this would mean removing support for IE9+10, as well as older versions of the android browser, before Android switched to just using Chrome.  I believe most people using Firefox, Chrome or Safari will be on recent enough  versions that it's not an issue.

mucaho:
I'm not sure.      

Crafty's support for older browsers is one of the distinguished features among the vast variety of available JS game engines.     
Crafty doesn't need to abandon older browsers. There are transpilers available (like Babel) which transform ES6 into ES5 compliant code if the browser does not support it. Keep in mind that the transformed code has awful performance, but works. e.g. [babel for sets](https://babeljs.io/learn-es2015/#ecmascript-2015-features-map-set-weak-map-weak-set). So code would still work on older browsers, just slower.   

-------

You know that I've been fiddling with more drastic approaches to get considerable performance boosts, like TypeScript or C++ -> emscripten -> asm / WebAssembly. Haven't been completely satisfied with either: TS wouldn't give an order of magnitude better performance. C++ is in my opinion a difficult language with many intrinsics and not suited for open source development, since things can break pretty easily.    

As of recent, [rust can compile to asm / WASM](https://hackernoon.com/compiling-rust-to-webassembly-guide-411066a69fde), but it's still in its infancy. It's a systems programming language with strict compile-time checks, that would enable safe contributions. I've read about upcoming changes to emscripten that could lead to further problems, so it's still questionable if that project will take off.

starwed:
>Crafty doesn't need to abandon older browsers. There are transpilers available (like Babel) which transform ES6 into ES5 compliant code if the browser does not support it. 

True, we could also simply supply separate builds of crafty for older browsers using such transpilers, if the generated code had a perf impact in modern ones.

>You know that I've been fiddling with more drastic approaches to get considerable performance boosts, like TypeScript or C++ -> emscripten -> asm / WebAssembly. 

[...]

PR #1117, opened 2017-04-17
Optimize iteration of entites by componentName

mucaho:
Instead of looping over all entities to check which one has a queried
component, use a map of componentNames to entities that have them.

@starwed There is some [strange piece of code that I removed](https://github.com/craftyjs/Crafty/blob/716ce14ec8b4049a02a267df77f3b72d27efde1b/src/core/core.js#L153-L156) from the Crafty.selector. It's [undocumented](http://craftyjs.com/api/Crafty.html). If I understand correctly when you do a selection like `Crafty('2D')` it will extend the global Crafty object with those properties and methods from `2D`. Huh? Do you know what's going on there?

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -56,7 +56,7 @@ var Crafty = function (selector) {
     return new Crafty.fn.init(selector);
 };
     // Internal variables
-var GUID, frame, components, entities, handlers, onloads,
+var GUID, frame, components, entities, handlers, onloads, compEntities,
 slice, rlist, rspace;
 
 
@@ -70,6 +70,7 @@ var initState = function () {
     frame       = 0;
 
     entities    = {}; // Map of entities and their data
+    compEntities= {}; // Map from componentName to (entityId -> entity)
     handlers    = {}; // Global event handlers
     onloads     = []; // Temporary storage of onload handlers
 };
@@ -95,12 +96,11 @@ Crafty.fn = Crafty.prototype = {
         if (typeof selector === "string") {
             var elem = 0, //index elements
                 e, //entity forEach
-                current,
                 and = false, //flags for multiple
                 or = false,
                 del,
                 comps,
-                score,
+                filteredCompEnts, compEnts, compEnts2, compEnt,
                 i, l;
 
             if (selector === '*') {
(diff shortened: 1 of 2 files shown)


mucaho:
I have tracked the removed code in question down: see 8c242f9835cd1f3a68c7ecf2d5774ddb418de7a1 
A long time ago, in a gal... it was really used to call component specific functionality on a set of entities! ([source](https://github.com/craftyjs/Crafty/blob/8c242f9835cd1f3a68c7ecf2d5774ddb418de7a1/demo.html#L30-L34))
```js
var platform = Crafty.e("2D, wall");
Crafty("gravity 2D").hit(platform,function() {
    console.log("HIT");
    this.y = 1;
});
```
This worked because you declared component methods like so: ([source](https://github.com/craftyjs/Crafty/blob/8c242f9835cd1f3a68c7ecf2d5774ddb418de7a1/components.js#L47-L56))
```js
hit: function(obj, fn) {
    this.each(function() {
        this.bind("hit", fn);
        this.bind("enterframe", function() {
            if(this.intersect(Crafty(obj))) {
                this.trigger("hit");
            }
        });
    });
}
```

I'm not sure why Crafty deviated from this path or didn't setup `this.each` automatically for all component methods.

starwed:
>If I understand correctly when you do a selection like Crafty('2D') it will extend the global Crafty object with those properties and methods from 2D.

You've probably figured this out as you go through the code, but it's not modifying the global Crafty object.  As a function, Crafty is defined as:
```
var Crafty = function(selector) {
  return new Crafty.fn.init(selector);
};
```
This is calling the init method as a constructor,  which means that `this` will generally mean the newly created object.  (Although IIRC the constructor will return an existing entity if there's only one match?)

I'll try to have a good look at the PR this weekend.  Thanks for tackling some fairly ancient parts of the code!  :)

mucaho:
> This is calling the init method as a constructor, which means that this will generally mean the newly created object

Maybe it would make sense to expose some more lightweight ways to get entities? Of course this way you wouldn't get the convenient group operation features, like `trigger` and `each`. `Crafty,getEntities(compName)` could return the internal read-only map `{entityId: entity}` and `Crafty.getEntity(id)` could return the entity itself.

starwed:
>Crafty.getEntity(id) could return the entity itself.

I believe `Crafty(id)` does actually return the entity itself.  (This is why the id of each entity is stored in the `0` property!)

mucaho:
> I believe Crafty(id) does actually return the entity itself.

Yes, but `Crafty()` actually calls a constructor first, so an unneeded object is created before the code realizes it has to return something other than `this`. Hmm maybe the selector `Crafty()` itself should be changed, instead of introducing a new method for it.

Anyways, that is a task for a subsequent PR :)

src/core/core.js
```diff
@@ -442,7 +466,8 @@ Crafty.fn = Crafty.prototype = {
             }
         }
         delete this.__c[id];
-
+        // update map from component to (entityId -> entity)
```
starwed, line 469: I find this description a little confusing.  I'd probably call it a map from component -> (hashset of entity ids).

src/core/core.js
```diff
@@ -130,31 +130,50 @@ Crafty.fn = Crafty.prototype = {
...
+                compEnts = compEntities[comps[0]];
+                compEnts2 = compEntities[comps[1]];
+                for (compEnt in compEnts) {
+                    if (compEnts2[compEnt] !== undefined)
+                        filteredCompEnts[compEnt] = +compEnt; //convert to int
+                }
 
-                        //if anded comps and has all OR ored comps and at least 1
-                    if (and && score === l || or && score > 0) this[elem++] = +e;
+                for (i = 2, l = comps.length; i < l; i++) { // strip initial object keys not in following sets
```
starwed, line 158: It seems kind of reasonable to optimize for 2 component names, but I'm curious if there's a nice way to generalize the`and` case without losing performance.

starwed (commented):
>Instead of looping over all entities to check which one has a queried
component, use a map of componentNames to entities that have them.

Looked over the code, and it seems pretty reasonable to me.  Just a couple comments/questions.

src/core/core.js
starwed: Oh, I guess it  is technically a lookup of `(entityId -> entity)`, but I don't think that is used anywhere in the code?  (You're entirely concerned with getting the ids, not the entities.)

mucaho: In the `Crafty()` selector we only need ids indeed. It could be changed to reflect this by having `{ componentName:  [id1, ..., idn] }`. However it might be useful later to retrieve entities of a component directly, without having to look up an entity from its id  (e.g. the suggested `getEntities` method).     
Which variant would you suggest here?

src/core/core.js
mucaho: I was not trying to optimize it for 2 component names per se :)   
From an algorithmic standpoint I guess the fastest thing to do is create a temporary map holding all ids for the first compName and then remove all ids that are not in common with the ids for the subsequent compNames.   
As we have to iterate all ids once to copy them over to the temporary map at the start, why not use this iteration to do some work already?

src/core/core.js
starwed: > It could be changed to reflect this by having `{ componentName: [id1, ..., idn] }`. 

This would probably be poor perf wise -- it would be expensive to delete ids out of the list.  The minimal approach would be something like `{componentName: {id1: true, id5: true}}` etc, but you're right that we might as well just point to entities instead.

(This is slightly related to #1106, since it's another place that a [set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) might be a slightly better match.).

mucaho:
So are there any still unresolved pointers from your side or can we merge this?

starwed:
Merged!

PR #1118, opened 2017-04-20
Update image component documentation

mucaho:
State how entity's dimensions influence the appearance.
Link to Sprite component for image scaling.
see #1077

### DIFF
--- a/src/graphics/image.js
+++ b/src/graphics/image.js
@@ -32,6 +32,14 @@ Crafty.defaultShader("Image", new Crafty.WebGLShader(
  * @kind Component
  * 
  * Draw an image with or without repeating (tiling).
+ *
+ * If the entity's width and height are smaller than the width and height of the image source, the image will appear cropped.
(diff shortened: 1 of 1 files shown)


src/graphics/image.js
```diff
@@ -32,6 +32,14 @@ Crafty.defaultShader("Image", new Crafty.WebGLShader(
  * @kind Component
  * 
  * Draw an image with or without repeating (tiling).
+ *
+ * If the entity's width and height are smaller than the width and height of the image source, the image will appear cropped.
+ * If the entity's dimensions are larger than the dimensions of the image source, there will be blank space around the image.
```
starwed: The current docs under the [image method](http://craftyjs.com/api/Image.html#-image) imply that this actually varies by renderer: 

>If set to no-repeat and given dimensions larger than that of the image, the exact appearance will depend on what renderer (WebGL, DOM, or Canvas) is used.

IIRC it is WebGL that handles this differently.

Issue #1119, opened 2017-04-23
Indeterministic test causes CI server to fail sometimes


mucaho:
The travis build recently failed so I restarted it, but the link I copied is no longer valid :(
It was something along the line
```shell
"After <someNumber> frames triggered, EndTime <undefined> must be larger than StartTime <someNumber> "
```
[Test in question](https://github.com/craftyjs/Crafty/blob/59d5b0189ed182936cbb9a089b70e989f7edd825/tests/unit/core/time.js#L150-L172) is 
```js
  test("curTime", function(_) {
    _.expect(1);
    var done = _.async(); // pause the QUnit so the timeout has time to complete.
    var startTime, lastKnownTime;

    var framesTriggered = 0;

    Crafty.e("").bind("EnterFrame", function(params) {
      framesTriggered++;
      if (!startTime) {
        startTime = params.gameTime;
      } else {
        lastKnownTime = params.gameTime;
      }
    });

    setTimeout(function() {
      var endTime = lastKnownTime;

      _.ok(endTime > startTime, "After " + framesTriggered + " frames triggered, EndTime " + endTime + " must be larger than StartTime " + startTime);
      done();
    }, 200);
  });
```
If I remember correctly I've had this happen to me locally once or twice.

PR #1121, opened 2017-04-29
New input implementation

mucaho:
Slim version #1116, as mentioned in comments there.


src/controls/controls-system.js
```diff
@@ -89,6 +85,12 @@ Crafty.s("Controls", {
         // internal object to store definitions
         this._dpads = {};
         this._triggers = {};
+
+        // listen to mouse events
+        Crafty.s("Mouse").bind("MouseDown", this.updateTriggers.bind(this));
+        Crafty.s("Mouse").bind("MouseUp", this.updateTriggers.bind(this));
+        Crafty.s("Mouse").bind("DoubleClick", this.updateTriggers.bind(this));
+        Crafty.s("Mouse").bind("Click", this.updateTriggers.bind(this));
```
starwed, line 97: These are also bound in the `events` property to the global events.

Is the issue here that the Mouse system now does not trigger the events globally?  If so, should they be removed from the event block?

src/controls/inputs-keyboard.js
```diff
@@ -0,0 +1,276 @@
...
+                        return null;
+                }
+
+                return evt;
+            },
+
+            changeFocus: function (oldFocusElem, newFocusElem) {
+                // TODO enable these once component iteration improved
+                //Crafty.s('Keyboard').trigger('FocusChange', newFocusElem);
+                //Crafty('Keyboard').trigger('FocusChange', newFocusElem);
```
starwed: This is somewhat related to an idea I originally had when working on systems -- giving a system a way of triggering events on a set of related components (by default with the same name).

In the end I decided that it was simpler to just always trigger events globally, and trust that any entity that subscribed to the event should receive it.  What's the specific rationale for only triggering a subset here (and I think elsewhere in the PR?)

src/controls/inputs-lifecycle.js
```diff
@@ -0,0 +1,68 @@
+var Crafty = require('../core/core.js'),
+    document = window.document;
+
+// figure out which eventName to listen to for mousewheel events
+var mouseWheelEvent = typeof document.onwheel !== 'undefined' ? 'wheel' : // modern browsers
+                        typeof document.onmousewheel !== 'undefined' ? 'mousewheel' : // old Webkit and IE
+                        'DOMMouseScroll'; // old Firefox
+
+//initialize the input events onload
+Crafty._preBind("Load", function () {
```
starwed: Nice that these are broken out.

There might be a way to completely avoid using `_preBind` here, and have the systems handle the event binding and unbinding directly.  I don't recall the exact issue, but the comment for prebind says:
> There are some events that need to be bound to Crafty when it's started/restarted, so store them here
>  Switching Crafty's internals to use the new system idiom should allow removing this hack

starwed (commented):
A lot of the diffs here are code moving around, which makes it a little hard to read through.  Could you highlight what you consider the important changes here?

I tried to [put together a chart](https://www.lucidchart.com/documents/view/95ac4e27-ac53-4640-bb29-a6ed632f5427) that shows the current structure of keyboard inputs/controls, and shows my idea of where locking would occur.  (Let me know if you can't access.)  A big blocker for my idea is that we don't implement mapping pointer events to controls.  I don't want that to hold up this work, but I do want to see if we agree on a possible final goal here.  :)

src/controls/controls-system.js
mucaho: > Is the issue here that the Mouse system now does not trigger the events globally?

Exactly, if Mouse system would trigger the events globally, all mouse entities would receive it. But maybe that's not an issue anymore with the newly proposed implementation?

mucaho:
Thank you for the feedback!   


>  A big blocker for my idea is that we don't implement mapping pointer events to controls _[yet]_

Yep, that's what I was wondering, too. If pointer events are being routed through the control system, the input system does not longer know which specific mouse entity should receive the event. Is this ok and should we just trigger mouse events globally? Anyways I'd leave this task to a subsequent PR.

mucaho:
> Could you highlight what you consider the important changes here?

The changes are mostly non-functional, in preparation for upcoming PRs that will shift the routing of events to the control system, introduce pointer events into the control system and add locking behavior to controls.
As you mentioned, most of the stuff is moving code around into appropriate namespaces. So instead of having `Crafty.resetMouseButtons`, `Crafty.mouseButtonsDown` and `Crafty.mouseDispatch` they would be properties / methods of the Mouse system. (Analogue for Keyboard, Touch).
The notion of dispatching and preventing the bubbling of events is abstracted into a common EventDispatcher base system that all these systems inherit from.
The validation of mousebutton- and keypresses is also abstracted into a common base system that these systems inherit from.
ResetMouseButton method is introduced, which similarly to resetKeyDown, resets actively pressed buttons once Crafty stage looses focus. These two methods are invoked on Crafty pause and stop, too.

I think your idea aligns with the goals of this PR. Will proceed with implementation and note detailed changes once done.

src/controls/inputs-lifecycle.js
starwed: For reference, prebind was added in #960, which provides some more context as to why it was necessary.

starwed:
>So instead of having Crafty.resetMouseButtons, Crafty.mouseButtonsDown and Crafty.mouseDispatch they would be properties / methods of the Mouse system

Makes sense.  Do we want to map the old names to the new system properties through some sort of alias, as done in #1071?

>If pointer events are being routed through the control system, the input system does not longer know which specific mouse entity should receive the event. Is this ok and should we just trigger mouse events globally? 

I think that would be the way to go, but to my mind there's a lot of uncertainty that can only really be pinned down by trying an implementation.

> Anyways I'd leave this task to a subsequent PR.

Agreed.

mucaho:
Implementation is done! Commit messages should contain detailed messages.

------------------------------------

EDIT: removed outdated info from this comment

---------------------

> There might be a way to completely avoid using _preBind here, and have the systems handle the event binding and unbinding directly.

I would tackle that in another PR, if that's ok.

> Do we want to map the old names to the new system properties through some sort of alias

Done.

mucaho:
> I tried to put together a chart that shows the current structure of keyboard inputs/controls ... In the end I decided that it was simpler to just always trigger events globally, and trust that any entity that subscribed to the event should receive it. 

Done! Found a clean way of implementing it, which I had trouble with at first.

mucaho:
>> If pointer events are being routed through the control system, the input system does not longer know which specific mouse entity should receive the event. Is this ok and should we just trigger mouse events globally?
>
> I think that would be the way to go

I may have an idea to get best of both worlds:    
We let the input system find the clicked entity (if it exists) and save it as a e.g. `target` property on the mouse event. The mouse events still get dispatched globally, but user can check if the current `Mouse` entity was clicked by comparing it with that property. Otherwise, if we don't find the `target` for the user, each `Mouse` entity would have call `findPointerEventTargetByComponent` and compare itself to the result, which would be very inefficient.
However, the controls system still remains in charge of routing the thereafter produced Control event, either to the locked / focused element or in its absence to the suggested `target` from the mouse event.

mucaho:
(Just updated this PR for new "UpdateFrame" event, other than that no changes).
Interestingly, bug #1119 occurred on AppVeyor this time. Have a look at the status checks.

mucaho:
To summarize:
* This commit fixes part of #552, quote: "Someone could hold down a key
while we've lost focus, and then release once it's back, so we probably
don't want to fire keyup events in such a case."
* This commit fixes Click on touch devices (as reported in #1043).

I hope I've addressed all of your suggestions, merging this in, and follow-up PR coming soon.

mucaho:
I've updated the list of things to do in the [Controls System project](https://github.com/craftyjs/Crafty/projects/1). If you have some pointers, I'd appreciate the feedback!

PR #1125, opened 2017-05-04
Change partial canvas redrawing

mucaho:
Instead of merging all changed entity rectangles, add spatial map cells of changed entities to a set.
Compute rectangles that represent the bounds of this cells.
Clear and redraw those rectangles.

## Performance
Ran two variants. See [example_partial_redraw.html](https://github.com/craftyjs/demos/blob/master/devDemos/example_partial_redraw.html) for the test code. Toggle the boolean flag to switch between entity distributions.

### Worst-case for new implementation
Worst case for new implementation is when small entities are positioned at the border of two spatial map cells, which triggers a redraw of two grid cells, whereas only a small area for the entity would be redrawn with old implementation.
**Performance of new implementation is in this case ~30% worse than of old implementation.**

#### Old impl chromium profile
canvas-old-old_bc
#### New impl chromium profile
canvas-new-old_bc

### Best-case for new implementation
Best-case for new implementation is a lot of entities near each other, but not overlapping. In this case a redraw of a grid cell redraws all those entities. It's bad for old implementation, since the entity's rectangles are not overlapping, so nothing can be merged and each entity rect needs to be redrawn separately.
**Performance of new implementation is in this case ~20% better than of old implementation.**

#### Old impl chromium profile
canvas-old-new_bc
#### New impl chromium profile
canvas-new-new_bc

### Conclusion

It's hard to say which implementation is better, but intuitively the new impl scales better. Within a more complex scene, I'd argue that better performance is more important than in an already smooth running trivial scene.   
There might be a way to squeeze out a bit more performance in `_createDirtyCell` by creating a number-indexed sparse array instead of a hash-string indexed object. Otherwise I'm not aware of any other things that could be done to improve it even further.   
On top of that, the partial canvas clear bug looks a bit more acceptable, but a fix for that is coming soon (if this impl is chosen, then on top of this, maybe in next PR?)

### DIFF
--- a/src/graphics/canvas-layer.js
+++ b/src/graphics/canvas-layer.js
@@ -1,6 +1,5 @@
 var Crafty = require('../core/core.js');
 
-
 /**@
  * #CanvasLayer
  * @category Graphics
@@ -13,9 +12,15 @@ var Crafty = require('../core/core.js');
 Crafty._registerLayerTemplate("Canvas", {
     type: "Canvas",
     
-    _dirtyRects: [],
-    _changedObjs: [],
     layerCount: 0,
+    _changedObjs: null,
+
+    _dirtyRects: null,
+    _dirtyCells: null,
+    _viewKeys: null,
+    __tempRect: null,
+    __tempSearchRect: null,
+    __tempScreenRect: null,
 
 
     /**@
@@ -99,8 +104,13 @@ Crafty._registerLayerTemplate("Canvas", {
         }
 
         // set referenced objects to initial values -- necessary to avoid shared state between systems
-        this._dirtyRects = [];
         this._changedObjs = [];
+        this._dirtyRects = [];
+        this._dirtyCells = {};
+        this._viewKeys = { x1: 0, y1: 0, x2: 0, y2: 0 };
+        this.__tempRect = { _x: 0, _y: 0, _w: 0, _h: 0 };
+        this.__tempSearchRect = { _x: 0, _y: 0, _w: 0, _h: 0 };
+        this.__tempScreenRect = { _x: 0, _y: 0, _w: 0, _h: 0 };
 
         //create an empty canvas element
         var c;
@@ -134,7 +144,7 @@ Crafty._registerLayerTemplate("Canvas", {
         if (!l && !dirtyViewport) {
             return;
         }
-        
+
         // Set the camera transforms from the combination of the current viewport parameters and this layers 
         var cameraOptions = this.options;
         if (dirtyViewport && cameraOptions) {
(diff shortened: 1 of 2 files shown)


mucaho:
> There might be a way to squeeze out a bit more performance in _createDirtyCell by creating a number-indexed sparse array instead of a hash-string indexed object.

Added. Slight improvement when many cells are drawn. Also avoids a bit of GC.

mucaho:
Why not keep the old implementation for simple scenes?
I've changed the code to redraw the canvas with three different implementations:
* Redraw all by default.
* If number of entities < 32, redraw with dirty entity regions.
* If number of entities < 60% of all entities, redraw with dirty cell
regions.
`32` is a simple metric, and could be improved.

------------------------------

There is an alternate approach to track stale regions to be drawn within the new drawDirtyGridCells method: Instead of keeping track of the object's `staleRect` and recompute the stale grid cells unecessarily again, keep track of these computed stale grid cells it belonged to. After profiling I've found that the cost of maintaining two additional sparse arrays for **each** entity (by means of filling and clearing) isn't worth it for small entities that just span a single grid cell, but it could be worth for bigger entities. Anyways, I would leave this as is for now.   

EDIT: It might make sense to track grid cells the entity belongs to inside the spatial map implementation itself, which would eliminate any re-computation. Will reevaluate performance with dense arrays instead and maybe adapt the more elegant inverse hash function from _updateBounds.

One potential problem I found is that bounds calculation works correctly for entities in range of (-32786, 32767). Is this enough?

------------------------------------------

I've just now discovered a regression bug that occurs when drawing merged `staleRect` and `currrentRect` of the `obj` to be drawn when used in combination with new `rectManager.integerBounds` method (that creates a tighter, but still enclosing integer rect). Need to figure out what's going on there.

src/graphics/canvas-layer.js
```diff
@@ -112,7 +115,11 @@ Crafty._registerLayerTemplate("Canvas", {
 
         // set referenced objects to initial values -- necessary to avoid shared state between systems
         this._dirtyRects = [];
+        this._dirtyCells = [];
         this._changedObjs = [];
+        this._tempKeys = {};
+        this._tempRect = { _x: 0, _y: 0, _w: 0, _h: 0 };
+        this._tempRect2 = { _x: 0, _y: 0, _w: 0, _h: 0 };
```
starwed: The rectangle manager does support an [object pool for rectangles](https://github.com/craftyjs/Crafty/blob/develop/src/spatial/rect-manager.js#L143-L183) -- perhaps you could use that here, instead of explicitly declared temp rects?

src/graphics/canvas-layer.js
```diff
@@ -191,102 +202,118 @@ Crafty._registerLayerTemplate("Canvas", {
...
-            }
+        // Canvas works better with integral coordinates where possible
+        view = rectManager.integerBounds(view);
 
-            // Sort objects by z level
-            objs.sort(this._sort);
+        // Calculate dirty spatial map cells from all changed objects
+        this._createDirtyCells();
+        // Afterwards, calculate dirty rectangles from dirty spatial map cells
+        this._createDirtyRects();
```
starwed: I haven't stepped through these carefully yet, but is it possible that they could be simpler if combined?

starwed (commented):
General thoughts:
- I think it's ok if we drop the previous _drawDirty method -- it's pretty complicated, and I'd be glad to get rid of it.
- I'm a bit concerned about making the implementation here directly linked to the internals of the spatial map, but I'm not sure how else you could do it.
- Each entity does have a pointer to it's spatial map entry, which in turn points to it's hashmap keys.  (so I think you could do `e._entry.keys`).  That might be useful?

I haven't looked through the bit twiddling logic carefully yet, but of course it seems fairly similar to what the spatial map uses.

starwed:
>One potential problem I found is that bounds calculation works correctly for entities in range of (-32786, 32767). Is this enough?

Is that the key index, or the actual x/y position?  I know that's something I considered when implementing the way the spatial map hashes x/y coordinates.  One thing to consider is that if two cells map to the same hash, that might cause extra iteration, but it shouldn't cause any correctness bugs -- since we clip to the region we intend to draw, even if extra entities are added to the list, we won't actually draw them.

src/graphics/canvas-layer.js
mucaho: Not as far as I know. We need one iteration to add cells to a set (we don't want to create duplicate rectangles). Another iteration is needed to create rects out of these cells.

src/graphics/canvas-layer.js
mucaho: Sounds good, but what's the benefit in this case exactly? The code needs only 2 rects at any time, won't it be an overhead to request and recycle them each render frame? Or would they be requested just once, like in the initialization code above? `this._tempRect = Crafty.rectManager._rectPool.get(0, 0, 0, 0)`

mucaho:
> I think it's ok if we drop the previous _drawDirty method -- it's pretty complicated, and I'd be glad to get rid of it.

Are you sure? We would be missing out on "free" performance. The canvas clear bug can be fixed in a single spot for all implementations, it's something that affects _drawDirtyCells as much as _drawDirty.

>> One potential problem I found is that bounds calculation works correctly for entities in range of (-32786, 32767). Is this enough?
>
> Each entity does have a pointer to it's spatial map entry, which in turn points to it's hashmap keys. (so I think you could do e._entry.keys). That might be useful?

Thanks! I have tried multiple complicated alternatives, whereas the simplest one eluded me for some reason :) A fresh set of eyes really helps to get some perspective.

> Is that the key index, or the actual x/y position? I know that's something I considered when implementing the way the spatial map hashes x/y coordinates. One thing to consider is that if two cells map to the same hash, that might cause extra iteration, but it shouldn't cause any correctness bugs

In the case of the [bounds calculations](https://github.com/craftyjs/Crafty/blob/01fd1abd1b71e680b96e47575fd299855bc81743/src/spatial/spatial-grid.js#L366-L434) it will be an issue, as the **exact** inverse coordinates are needed, not the hash bucket of entities. However 32k seems like a reasonable number, it's just something to keep in mind should the issue arise (similar to the WebGL max entity limit). The issue can be solved with some more complicated bit twiddling, but let's cross that bridge when and if we come to it.
```js
for (var h in this.map) {
    if (!this.map[h].length) continue;

    //broad phase coordinate
    var i = h >> 16,
        j = (h << 16) >> 16;
    if (j < 0) {
        i = i ^ -1;
    }

    if (i >= hash.max.x) {
        // i can't be larger than 32k here
        hash.max.x = i;
        ...
    }
...
}
```

mucaho:
>> I think it's ok if we drop the previous _drawDirty method -- it's pretty complicated, and I'd be glad to get rid of it.
>
> Are you sure? We would be missing out on "free" performance. The canvas clear bug can be fixed in a single spot for all implementations, it's something that affects _drawDirtyCells as much as _drawDirty.

On second thought, let's do exactly that.   

I have managed to get some pretty good results with scaled viewport drawing when transforming the "exact" world coordinates used for clearing and clipping the screen to "exact" screen coordinates in the `_drawDirtyGridCells` method.  

`_drawDirty` can be adapted to work with with this too (see [my comment in the relevant issue](https://github.com/craftyjs/Crafty/issues/711#issuecomment-298436897)). However, it's rather a guessing game of extending the rectangle bounds to be drawn, which works well for some scale factors I tested (0.5111 - 10.511111), but might not work well in all cases.
I have been debugging code all day long, but can't find out why it's not possible to use exact coordinates when drawing merged stale-dirty and currently-dirty rectangles. If someone more apt figures that out we can merge the functionality back in.

Will remove the `_drawDirty` method. This allows the common drawing code to be simplified further.

mucaho:
Done. This should fix bug #711.  Tested different examples on different viewport scales on newest versions of chrome, firefox, edge.

There may be some edge cases where 1 px is shaved off objects being drawn (they appear 1 px smaller) when viewport is zoomed heavily (e.g. scale `10.15555`) on the edges of grid cells. At this point I don't think it can get any better than that with partial canvas redrawing. If this becomes noticeable, the only thing left to do is switch to full canvas redraw.

mucaho:
Merging this in now as it seems to fix the issue. If there are any further problems or any other suggestions on your part we can sort it out before release.

src/graphics/canvas-layer.js
mucaho: Ah yes, it could be changed so that dirty rectangles get created alongside dirty cell in a single iteration, if the current cell hasn't been marked as dirty yet.

PR #1127, opened 2017-05-13
New input implementation: Part 2

### LINKED ISSUE #1013: Viewport's mouselook conflicts with other Mouse entities

mucaho:
# Follow-up of #1121 

* Change `MouseWheel` and `Mouse` events to wrap original events in a standard Crafty event object (similar to what KeyboardDispatch does). Additional reason: _"Some browsers (mobile Safari, for one) re-use touch objects between events, so it's best to copy the bits you care about, rather than referencing the entire object."_ ([source](https://developer.mozilla.org/en/docs/Web/API/Touch_events#Copying_a_touch_object))
* Add `target` property to these events, which is a reference to closest entity to pointer event, otherwise `null`
* Convert TouchSystem into a first-class EventDispatcher: Split state handling (TouchState) from DOM event data normalization and dispatch (TouchSystem)
* Fixes Draggable and Crafty.viewport.mouselook to work again together. Also fixes #1013 .
* Updates the documentation.
* **Breaking change:** 
  * TouchStart and TouchEnd are no longer triggered
when a finger enters or leaves entity. They are now triggered only
once, when a finger is pressed or raised on the touch surface, no
matter which entity was targeted.
  * TouchOver and TouchOut replace those events, which are triggered
when a finger enters or leaves the entity.

# About input events
Input events can be organized into two categories:
* Global-context events
  * MouseDown, MouseMove, MouseUp
  * TouchStart, TouchMove, TouchEnd, TouchCancel
  * MouseWheelScroll
  * KeyDown, KeyUp
* Entity-specific events
  * Click, DoubleClick, MouseOver, MouseOut
  * TouchOver, TouchOut

If the input events get triggered globally in future, it wouldn't make much sense to trigger entity-specific events globally, too. For example, a Click is specific to an entity - MouseDown and MouseUp must occur while mouse pointer is over the entity. They should in my opinion either be removed or continue to be triggered on individual entities (probably latter option is better for users?).

One way to handle this would be to make entity-specific input events optional:
* There could be e.g. a `Crafty.setting` which would opt-in / opt-out from triggering any events on individual entities.
* Another option would be to trigger global-context events globally, but also trigger entity-specific events on individual entities. If there are no `Mouse` entities (`mouseObjs === 0`) or `Touch` entities (`touchObjs === 0`), then there is no need to generate or trigger any entity-specific events. (I vote for this option)
* Third option is to leave it as it is now: global-context events get triggered on the MouseSystem / TouchSystem / KeyboardSystem. These also get triggered on the target entity if it exists. Entity-specific events get triggered on the appropriate target entity.

If entity-specific events become optional or get removed, that would allow for additional performance optimization. The current `target` property, which is computed on every pointer event, could be made a lazy getter. Global-context events are handled by MouseState and TouchState and are `target`-agnostic. Entity-specific events are generated by MouseSystem and TouchSystem (mostly in response to global-context events) and need to examine the target entity of the pointer event. Therefore the call to the computation-heavy `findPointerEventTargetByComponent` could be avoided in many cases.

# Keep TouchCancel?

The `TouchCancel` event is fired _"when a touch event has been disrupted in some way"_, e.g. if a popup occurs or (for the moment) when Crafty looses focus it is triggered by `resetTouchDown`.  
The event cause is different, but internally it is handled exactly the same as the `TouchEnd` event, that occurs _"when a finger has been raised from the touch surface"_.  
Should `TouchCancel` be removed and should just `TouchEnd` be called? Or is there a conceivable use-case where a game dev would like to know the difference?

------

Still need to fix some tests and add some test cases, but opened this PR for discussion about the general direction and implementation.

### DIFF
--- a/src/aliases.js
+++ b/src/aliases.js
@@ -23,6 +23,8 @@ module.exports = {
 
         createDeprecatedAlias(Crafty, "touchDispatch", Crafty, "_touchDispatch");
         createDeprecatedAlias(Crafty, "touchObjs", Crafty.s('Touch'), "touchObjs");
+        Crafty.touchHandler = {};
+        createDeprecatedAlias(Crafty.touchHandler, "fingers", Crafty.s('Touch'), "touchPoints");
     }
 };
 
--- a/src/controls/keyboard.js
+++ b/src/controls/keyboard.js
@@ -5,18 +5,18 @@ var Crafty = require('../core/core.js');
  * @category Input
  * @kind Component
  *
- * Provides valid key related events and key states for the entity.
+ * Handles valid key related events and key states for the entity.
+ * @note This is an internally used component, automatically included in the `KeyboardSystem`.
  *
- * @trigger KeyDown - is triggered when the DOM 'keydown' event is triggered. - { key: `Crafty.keys` keyCode (Number), originalEvent: original KeyboardEvent } - Crafty's KeyboardEvent
- * @trigger KeyUp - is triggered when the DOM 'keyup' event is triggered. - { key: `Crafty.keys` keyCode (Number), originalEvent: original KeyboardEvent } - Crafty's KeyboardEvent
+ * @trigger KeyDown - when a key is pressed - KeyboardEvent
+ * @trigger KeyUp - when a key is released - KeyboardEvent
  *
- * The event callbacks are triggered with a native [`KeyboardEvent`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent) received by `window.document`,
- * which is wrapped in a standard Crafty event object consisting of additional properties:
+ * The standard Crafty `KeyboardEvent` object:
  * ~~~
  * // event name of key event
  * e.eventName
  *
- * // Normalized keyCode according to Crafty.keys
+ * // Normalized keyCode number according to `Crafty.keys`
  * e.key
  *
  * // Original keyboard event, containing additional native properties
@@ -108,7 +108,7 @@ Crafty.__keyboardStateTemplate = {
      * @param eventName - Name of the key event to trigger ("KeyDown" or "KeyUp")
      * @param eventData - The key event to trigger
      *
-     * Try to trigger a key event on this entity.
+     * Try to trigger a key event on this entity and persist the key state.
      * This method prevents inconsistent key state.
      * e.g. If this entity didn't receive a "KeyDown" previously, it won't fire a "KeyUp" event.
      *
--- a/src/controls/mouse.js
+++ b/src/controls/mouse.js
@@ -5,18 +5,14 @@ var Crafty = require('../core/core.js');
  * @category Input
  * @kind Component
  *
- * Provides valid mouse related events and button states for the entity.
+ * Handles valid mouse related events and button states for the entity.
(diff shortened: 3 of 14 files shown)


mucaho:
> Still need to fix some tests and add some test cases

Done!

Tested touch events on Chrome, Firefox & Edge. Chrome seems to work, Edge seems to work too after enabling something in `about:flags`. Firefox seems to be currently broken, so touch events don't work on it. See [example_pointer_events.html](https://github.com/craftyjs/demos/blob/master/devDemos/example_pointer_events.html)

starwed:
I'll need some time to take a proper look through this, but a couple things off the top of my head: 

Don't the `MouseUp/Down` events potentially also have a target entity?

I'm still not really convinced that for events that target a particular entity, only that entity cares about the event.  For instance, you might want clicking on an entity to trigger a player action/interaction with that entity (such as moving towards it and then stopping.)  You could work around that with the current approach by having the clicked entity trigger a global event, but we could also just trigger such a global event by default.  Then some combination of components and/or controls could provide behavior analogous to the current click behavior.  

I do think you're correct that the event dispatch system should handle entity picking, rather than my original suggestion of moving that into the control system.

mucaho:
>  For instance, you might want clicking on an entity to trigger a player action/interaction with that entity (such as moving towards it and then stopping.) ...
Then some combination of components and/or controls could provide behavior analogous to the current click behavior.

Agreed. That's what I'm aiming at: Current click behavior on individual entities should be preserved _somehow_, as it's expected by users.

> I do think you're correct that the event dispatch system should handle entity picking, rather than my original suggestion of moving that into the control system.

What do you mean? There are 2 separate things: Finding the target clicked and triggering event on found target. I was under the impression that you had the latter thing (edit:) _for the control system in mind_

starwed:
>What do you mean? There are 2 separate things: Finding the target clicked and triggering event on found target. I was under the impression that you had the latter thing (edit:) for the control system in mind

I was originally thinking that the control system should also be responsible for finding the target, but I'm now convinced that it makes sense for that to happen during event normalization/dispatch.

mucaho:
> but I'm now convinced that it makes sense for that to happen during event normalization/dispatch.

Alright then, merging this in. If you have further suggestions, let me know!

Issue #1128, opened 2017-05-15
Ideas and bugs for Control System


mucaho:
# Bug 1
DPads and Triggers do not get cleaned up when entity is destroyed. They linger in Controls System even if nothing is using them.

Some ideas:
* Reference count of whatever uses them, if triggers / dpads are meant to be used by mulitple objects
* Otherwise, let the object dispose of the trigger once it no longer uses it (e.g. in entity / system `remove` method)

# Bug 2 
DPad is incorrectly computing active directions due to rounding errors when multiple directions are active.  Happens when you use diagonal directions. 
To reproduce, replace [these lines](https://github.com/craftyjs/Crafty/blob/0bdae16933e0e8e8d893ec4a725eb355bd4e6a73/tests/unit/controls/controls.js#L496-L501) with
```js
    // pressing two keys of opposite directions cancels out all movement
    keysDown(Crafty.keys.NUMPAD_2, Crafty.keys.NUMPAD_8);
    Crafty.timer.simulateFrames(1);
    _.strictEqual(e._vx, 0, "vx = 0 when NUMPAD_8 & NUMPAD_2 pressed");
    _.strictEqual(e._vy, 0, "vy = 0 when NUMPAD_8 & NUMPAD_2 pressed");
    keysDown(Crafty.keys.NUMPAD_7, Crafty.keys.NUMPAD_3);
    Crafty.timer.simulateFrames(1);
    _.strictEqual(e._vx, 0, "vx = 0 when NUMPAD_8 & NUMPAD_2 & NUMPAD_7 & NUMPAD_3 pressed");
    _.strictEqual(e._vy, 0, "vy = 0 when NUMPAD_8 & NUMPAD_2 & NUMPAD_7 & NUMPAD_3 pressed");
    keysUp(Crafty.keys.NUMPAD_7, Crafty.keys.NUMPAD_3);
    keysUp(Crafty.keys.NUMPAD_2, Crafty.keys.NUMPAD_8);
```

Some ideas:
[...]

PR #1129, opened 2017-05-17
Fix dpad rounding errors

mucaho:
Fix dpad rounding errors when sum of vectors is close to zero.
This prevents movement at still state and prevents normalization
to convert a should-be zero vector to a unit vector.

Fixes part of #1128 .


starwed:
Looks good 👍

PR #1131, opened 2017-05-17
Setup event callbacks before init

mucaho:
Setup event callbacks defined in the events object before calling
the init method in component and system definition.
This allows triggering bound callbacks in the init method.

see #1126


starwed:
I guess the big question is, are there any common use cases where you want to bind events *after* the init function?  

I can't really think of any, and I think binding events before running init is the least surprising order.  You can always explicitly bind events at the end of `init` if that's what you really want.

mucaho:
> You can always explicitly bind events at the end of init if that's what you really want.

Good point, workaround is available if needed.

PR #1132, opened 2017-05-18
Improve particle implementation

mucaho:
Render particles like other graphics components:
* Update particle position & shape on "UpdateFrame", trigger "Invalidate"
* Render particles on "Draw", but only if inside entity's bounds

### DIFF
--- a/playgrounds/2D/raycast/raycast.js
+++ b/playgrounds/2D/raycast/raycast.js
@@ -154,7 +154,7 @@ function raycast(start, end) {
       'distance=', results[i].distance.toFixed(2),
       '@[', results[i].x.toFixed(2), results[i].y.toFixed(2), ']');
 
-    Crafty.e("Hit, 2D, DOM, VisibleMBR")
+    Crafty.e("Hit, 2D, DOM, SolidMBR")
           .setName("Hit"+i)
           .attr({
             x: results[i].x - 2, w: 4,
--- a/src/debug/debug-layer.js
+++ b/src/debug/debug-layer.js
@@ -13,7 +13,10 @@ var Crafty = require('../core/core.js'),
  *
  * Crafty.debugCanvas.init() will be automatically called if it is not called already to initialize the canvas element.
  *
- * To visualise an object's MBR, use "VisibleMBR".  To visualise a "Collision" object's hitbox, use "WiredHitBox" or "SolidHitBox".
+ * To visualise an object's MBR, use "WiredMBR" or "SolidMBR".
+ * To visualise a "Collision" object's hitbox, use "WiredHitBox" or "SolidHitBox".
+ * To visualize an entity's click area, use "WiredAreaMap" or "SolidAreaMap".
+ *
  * @see DebugPolygon,  DebugRectangle
  */
 Crafty.c("DebugCanvas", {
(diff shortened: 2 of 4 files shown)


starwed:
Nice.

I do wonder if this is somewhat incompatible with partial redraws -- if the particles entity straddles two different rendered regions, the render method will get called twice.

I'm not sure that really matters; I think rendering takes the most time, and that should get skipped becuase it's happening outside the clipped region.

mucaho:
There's a catch with this implementation: It's more complicated to use APi-wise.  
Creating an entity, setting its `x` & `y` and adding particles to it won't work any longer.  
With the new changes, you would have to create an entity, set its `x`, `y` and guess the appropriate `w` & `h` to encompass all particles, plus center the particle emitter using the `originOffset` option.  
That's a lot more work we'd want to save the user from preferably.

I have an idea to make this process simpler:
* Merge #1089 in
* Automatically set `origin('center')`, tell user about this in API docs
* Deprecate `originOffset` and spawn particles at the entity's `_origin`. As it's no longer possible to use another graphics component on the entity, this option is now redundant.
* Tell user to set the entity's position with `ox` and `oy` (#1089), as the particle emitter will emit particles from that exact position (Btw originOffset has a minor issue: it doesn't account for particle size, meaning particles with random sizes won't spawn at same position)
* Automatically adjust the entity's dimensions based on the max distance these randomized particles will travel?.
  * It's possible to compute this, however I'm not sure that is the best way to handle this.
  * At some point I would like to see an additional feature added, namely some kind of floor / wall the particles will bounce off or rest at. That's the only feature from [pEng.js](http://samlancashire.com/peng/) the current implementation is missing, afaik. The entity bounding rectangle naturally describes these borders, thus it would probably be better to leave it up to the user to specify `w` and `h`.

Anyways, I think these features should be added in another PR.

starwed:
>There's a catch with this implementation: It's more complicated to use API-wise.
Creating an entity, setting its x & y and adding particles to it won't work any longer.

Yeah, all the other renderable objects are quads, while particle generators aren't really.  I think most of your suggestions make sense.  👍 

I've been looking a little bit about how you could create a matching implementation in WebGL.  It looks feasible, though you'd have to override most of the methods that other webgl entities use.

mucaho:
Alright, merging.

mucaho:
> I've been looking a little bit about how you could create a matching implementation in WebGL. It looks feasible, though you'd have to override most of the methods that other webgl entities use.

I wonder if it would be ok to use an entity for each particle instead. `Crafty.e("Renderable, Motion, Gravity, WebGL, Color")`. Creating 150+ entities would be very costly, but they could be added to an entity pool. Without the "2D" component there shouldn't be many computations on each frame, so maybe it would be feasible to reuse this existing functionality for updating & rendering particles.

starwed:
>I wonder if it would be ok to use an entity for each particle instead

I suspect that this would prove too costly.

Just considering WebGL rendering, normal entities are drawn as a quad consisting of two triangles connecting four vertices.  If you have a custom renderer for particle effects, it can just use the built-in functionality for rendering points, which requires only a single vertex.

There's another potential advantage of using a custom implementation for particles: For regular entities, we [copy](https://github.com/craftyjs/Crafty/blob/develop/src/graphics/webgl-layer.js#L163) their positions/etc into a typed array, and then load that array into the webgl buffer.  Since the position/etc of individual particles can be completely internal to the "Particles" entity, all that data could just be stored in a typed array to start with, which I suspect would help performance a lot!

mucaho:
>  If you have a custom renderer for particle effects, it can just use the built-in functionality for rendering points, which requires only a single vertex.

Nice, I forgot about `GL_POINTS`. Although there are some limitations to it, e.g. with regards to the max point size. [Test it for your machine](http://alteredqualia.com/tmp/webgl-maxparams-test/).

> all that _[particle]_ data could just be stored in a typed array to start with, which I suspect would help performance a lot!

Sounds good, as long as we provide fallback for browsers without support for typed arrays / WebGL. Older ES5 browsers can still render on Canvas.

PR #1133, opened 2017-05-21
Cleanup no longer used controls

mucaho:
Introduce undefineTriggerGroup and undefineDpad.  
I'm not sure about the names, would e.g. destroyDpad be better?

Fixes part of #1128 .


starwed:
>I'm not sure about the names, would e.g. destroyDpad be better?

I think I"d slightly prefer destroy to undefine.

Possibly we should fire events when defining or destroying controls?

mucaho:
> I think I"d slightly prefer destroy to undefine. Possibly we should fire events when defining or destroying controls?

Done

PR #1135, opened 2017-05-28
Don't create the class when loading the dependency

starwed:
This was causing a React deprecation warning to print anytime a grunt task was run.

We should probably fix that warning at some point, but we also probably shouldn't be creating the React class when the dependency is loaded.


starwed:
Simple fix, so merging.

PR #1136, opened 2017-05-28
Color was creating an array in the inner render loop

starwed:
A very simple performance fix -- the color entity was creating extra garbage because of [this line](https://github.com/craftyjs/Crafty/blob/develop/src/graphics/webgl.js#L232) (and similar lines in other render layers.)


starwed:
Small fix, so merging.  

(I guess you could argue that the render loop should not be attempting to create an array in the first place, but I'll leave that alone for now.)

PR #1138, opened 2017-05-30
Simplify draw signature for performance reasons

starwed:
When using an entity pool, I discovered that the single largest source of allocations in Firefox (~90%) seemed to be instantiating the `arguments` object in the webgl `draw` method.

That method accepts optional arguments, but they're not used anywhere in Crafty, and I'm fairly certain they wouldn't work correctly if you tried to pass them in -- the signature and surrounding logic was cloned from the canvas implementation, and never cleaned up.

So
- Remove the arguments completely from the webgl `draw` method.
- Remove the reference to `arguments.length` from the canvas `draw` method, and simplify the signature

### DIFF
--- a/src/graphics/canvas.js
+++ b/src/graphics/canvas.js
@@ -44,7 +44,7 @@ Crafty.c("Canvas", {
      * @comp Canvas
      * @kind Method
      * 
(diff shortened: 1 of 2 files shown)


mucaho:
It was even causing deoptimizations in chrome. 👍

PR #1141, opened 2017-06-05
Improve particle performance

mucaho:
* Extend API to restart particle animation with previously used options.
* Minimize particle object allocation - reuse particles from particle pool instead.
* Flatten particle vector properties into plain properties.
* Split particle color update from rendering better, to facilitate easier rendering with WebGL backend.
* Simplify and remove some redundant code.

Planned for next PR: Change particles to be a view into a TypedArray (as discussed in #1132 )

### DIFF
--- a/src/graphics/particles.js
+++ b/src/graphics/particles.js
@@ -9,7 +9,8 @@ var Crafty = require('../core/core.js');
  * #Particles
  * @category Graphics
  * @kind Component
- * 
+ *
+ * @trigger ParticleStart - when the particle animation has started
  * @trigger ParticleEnd - when the particle animation has finished
  *
  * Create particle effects.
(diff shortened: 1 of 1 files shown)


src/graphics/particles.js
```diff
@@ -146,223 +160,232 @@ Crafty.c("Particles", {
...
+                    particle.colourR = c > 255 ? 255 : c < 0 ? 0 : ~~c;
+                    c = particle.colourG + particle.deltaColourG;
+                    particle.colourG = c > 255 ? 255 : c < 0 ? 0 : ~~c;
+                    c = particle.colourB + particle.deltaColourB;
+                    particle.colourB = c > 255 ? 255 : c < 0 ? 0 : ~~c;
+                    c = particle.colourA + particle.deltaColourA;
+                    particle.colourA = c > 1 ? 1 : c < 0 ? 0 : (~~(c * 100)) / 100;
+
+                    // Decrease particle's lifespan
+                    particle.timeToLive--;
```
starwed, line 311: With the current implementation, this is really `framesToLive` -- at some point it should get updated to use the proper `dt`.  That would maybe involve changing how velocity is defined.

starwed (approved):
Looks good -- this should help the performance of particles quite a lot.

src/graphics/particles.js
mucaho: Yep, for further roadmap, see the [particle fx project](https://github.com/craftyjs/Crafty/projects/3)

mucaho:
Thanks for taking a look at it!

PR #1142, opened 2017-06-07
Add CDNJS version badge in README.md

sufuf3:
This badge will show the version on CDNJS!


mucaho:
Thanks for the PR!   
Maybe it would be better if we add this link to the website http://craftyjs.com, under the installation section, alongside these other variants.

PR #1143, opened 2017-06-08
Fix pointer entity tracking for areamap

### LINKED ISSUE #1140: Variation in behavior when component 'required' order changed (Mouse component)
Sorry about the cryptic title, this is kind of a weird issue. I have only just started using this library, so I might be doing something wrong.

My issue is the following:
1. Create a component with the required: set to '2D, Mouse, Canvas, Color'
2. Create an entity with this component
3. Destroy the entity
4. Create another entity with this component
5. It no longer correctly gets mouse events (it did before destroying the original entity)

On the other hand, if I specify the required components as '2D, Canvas, Color, Mouse', it works correctly.

My reproducer:
https://jsfiddle.net/ks8ryt69/1/

If you run this you'll find that the mousemove event will not fire. If you uncomment the commented out line, it will now work.

I couldn't find any notes on this gotcha and it took me quite a while to figure out the problem, so I'm assuming it's a bug.

It's not critical, since there's a simple workaround, but if I'm doing something wrong please let me know.

starwed:
We need to decrement the number of pointer entities when detached from a layer, or when the component is removed.

However, when the entity is destroyed and both these conditions are true, only decrement once.

Several unit tests added for this behavior.

Fixes #1140 (The sensitivity to the order of components was because of the "Renderable" check in AreaMap's `remove()`.)

### DIFF
--- a/src/inputs/pointer.js
+++ b/src/inputs/pointer.js
@@ -129,8 +129,8 @@ Crafty.c("AreaMap", {
         }
     },
 
-    remove: function () {
-        if (this.has("Renderable") && this._drawLayer) {
+    remove: function (isDestruction) {
+        if (!isDestruction && this.has("Renderable") && this._drawLayer) {
             this._drawLayer._pointerEntities--;
         }
     },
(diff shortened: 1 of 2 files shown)


mucaho:
Looks good! I'm unable to verify at the moment, if the code from the linked jsfiddle works as expected with this change.  
Would you kindly check that please before merging?

starwed:
Confirmed that the jsfiddle now works with either order.

PR #1144, opened 2017-06-24
Allocate fewer objects when searching

starwed:
A few changes to Crafty.map.search

- Maximum of one array created per search
- Possibility to provide 'output' array
- Split filtererd/unfiltered search into two separate methods

Because of the last two points, I still need to update some of the docs/comments, but figured I'd check whether we wanted to change the api like this first.

(I also have a follow-up PR that simplifies a couple of the core 'Collision' methods in a similar way.)


starwed:
Merging.  There is probably some follow-up work to make more internal uses of `search` pass in a results object.

PR #1145, opened 2017-06-24
Don't generate objects on each rotation

starwed:
Right now rotation info is passed around as an object that is constructed each time, which is somewhat expensive.

To avoid this:
- Alter the `rotate` method so that rotation paramters are arguments of the function, rather than properties of the object
- Don't generate a rotation object when triggering the "Rotate" event, instead passing through the amount of rotation
- Split rotation cascade into its own method, and generate more complex rotation parameters only when it's called

This also required updating other calls to `rotate`, but I think this is a general improvement.

An alternative approach I'd considered was to use an object pool, as is done for the 'Move' event.  However, it didn't seem like we actually needed that complexity here.


starwed:
The api change here is the change to the Rotate event argument.  Anyone relying on the cos/sin properties would now have to calculated them; that seems like a decent compromise to get a good performance boost for rotations in general.

PR #1146, opened 2017-06-25
Remove the 'Moved" event

starwed:
We currently emit a 'Move' event when the position is set,
and a separate 'Moved' event when the position is set by the 'Motion' event.
(I believe this is baggage left over from how Multiway/etc used to work?)

This has a fairly high cost, since on a typical motion tick we emit 4 events,
two for each axis.

This PR simplifies things by
- Removing the 'Moved' event completely
- Adding a method that handles both x and y change in one pass.

The result is that now linearMotionTick will only emit one move-related event per frame.
In limited performance testing, this had a fairly large impact.
In a test involving several hundred falling entities, linearMotion + event callbacks went from
consuming 13% of the total CPU time to 5.5%. 

In one sense this is a fairly large change to the api, but I don't think emitting these separate 'Moved' events is very useful; you can always recover the same info from the 'Move' event.  But perhaps there is a specific utility in separate events for x/y motion that makes it worth the performance cost?

### DIFF
--- a/src/spatial/2d.js
+++ b/src/spatial/2d.js
@@ -678,8 +678,7 @@ Crafty.c("2D", {
      * for an opposite direction.
      */
     shift: function (x, y, w, h) {
-        if (x) this.x += x;
-        if (y) this.y += y;
+        if (x || y) this._setPosition(this._x + x, this._y + y);
         if (w) this.w += w;
         if (h) this.h += h;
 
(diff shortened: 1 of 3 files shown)


starwed:
@mucaho Can you think of any reason (other than backwards-compatibility) we should keep both sets of events around?

starwed:
Eh, merging.  We can always revert before release if this turns out to be an issue.

PR #1147, opened 2017-07-01
Optimize hit() and _SAT()

starwed:
The main win here is reducing the number of allocations.  

`hit()` was also doing an intermediate check between the broad-phase search and the _SAT collision test.  I believe this increases the amount of work in the common case (and in checking performance in Firefox, the check was certainly using up about the same amount of CPU time as the _SAT check itself, so...) 

- Allow results object to be passed to hit()
- Flatten the results of _SAT to reduce allocations a bit when hits are found.
- Optimize hit a bit to reduce the amount of work done
  - Remove an intermediate overlap check; it's unnecessary when we're already using broad-phase + SAT
  - Assume that the collision component always has map
  - Return as early as possible
- Updated the documentation to include the `nx` and `ny` properties.

### DIFF
--- a/src/spatial/collision.js
+++ b/src/spatial/collision.js
@@ -425,18 +425,23 @@ Crafty.c("Collision", {
      * @comp Collision
      * @kind Method
      * 
-     * @sign public Array .hit(String component)
+     * @sign public Array .hit(String component[, Array results])
      * @param component - Check collision with entities that have this component
      * applied to them.
+     * @param results - If a results array is supplied, any collisions will be appended to it 
      * @return `null` if there is no collision. If a collision is detected,
      * returns an Array of collision data objects (see below).
+     * If the results parameter was passed, it will be used as the return value.
      *
-     * Tests for collisions with entities that have the specified component
-     * applied to them.
-     * If a collision is detected, data regarding the collision will be present in
-     * the array returned by this method.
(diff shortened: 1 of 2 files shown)


src/spatial/collision.js
```diff
@@ -483,46 +487,46 @@ Crafty.c("Collision", {
...
-        }
-
-        for (key in dupes) {
-            obj = dupes[key];
-
-            if (hasMap && 'map' in obj) {
-                var SAT = this._SAT(this.map, obj.map);
-                SAT.obj = obj;
-                SAT.type = "SAT";
-                if (SAT) finalresult.push(SAT);
```
starwed, line 520: Fun fact: when `_SAT` returned `false`, this code was previously trying to set the `obj` and `type` properties on that value.  This 'works' in javascript in the sense that it doesn't produce an error; instead, the boolean primitive is implicitly coerced to an object, the value is set on that object, and then the object is ignored (since it isn't assigned or referenced anywhere.)  I think the intermediate object was probably getting optimized away (since I didn't see this code get flagged as a source of allocations) but this was at best very confusing.  (Here is [an article that describes what happens when you set properties on primitives.](https://javascriptweblog.wordpress.com/2010/09/27/the-secret-life-of-javascript-primitives/))

PR #1154, opened 2017-09-17
Handle stage element without explicit w and h

### LINKED ISSUE #1153: Components don't receive mouse events when stage doesn't have defined size
I ran into an issue where I was trying to attach mouse events to a component and the event handlers weren't firing. After stepping through the event handling code, I found the problem was that it wasn't locating the target object because the viewport's width was `NaN`, so when translating the mouse event's location to the local coordinates, it ended up with `x: NaN`.

I reduced the bug down to how the stage gets initialized: it happens if you don't give the stage a defined size. For example:

```javascript
<html>
  <head></head>
  <body>
    <div id="game"></div>
    <script type="text/javascript" src="https://rawgithub.com/craftyjs/Crafty/release/dist/crafty-min.js"></script>
    
    <script>
      const buggy = true;
      
      if (buggy) {
        Crafty.init(document.getElementById('game'));
      } else {
        Crafty.init(500, 350, document.getElementById('game'));
      }

      Crafty.e("2D, DOM, Text, Mouse")
        .text("Click me")
        .bind('Click', function() {
          alert("Clicked!");
        })
        .attr({x: 0, y: 0, w: 100, h: 50});
    </script>
  </body>
</html>
```

https://codepen.io/michaelmelanson/pen/LzpGmb?editors=1000

In this example, the event handler will not get fired if `buggy=true`.

starwed:
Hopefully fixes #1153 by allowing the stage element to be specified without explicit dimensions.

### DIFF
--- a/src/graphics/viewport.js
+++ b/src/graphics/viewport.js
@@ -568,8 +568,8 @@ Crafty.extend({
          * @comp Crafty.stage
          * @kind Method
          * 
-         * @sign public void Crafty.viewport.init([Number width, Number height, String stage_elem])
-         * @sign public void Crafty.viewport.init([Number width, Number height, HTMLElement stage_elem])
+         * @sign public void Crafty.viewport.init([Number width, Number height][, String stage_elem])
+         * @sign public void Crafty.viewport.init([Number width, Number height][, HTMLElement stage_elem])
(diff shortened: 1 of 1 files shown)


michaelmelanson:
👍 LGTM!

Issue #1160, opened 2018-01-02
Add .delaySpeed and .tweenSpeed just like .animationSpeed


matthijsgroen:
I'm trying to make the death of the player extra dramatic, by slowing town time when the player explodes. This works create with sprite animations, by using `.animationSpeed` and slowing that one down.

These functions do not exist for delay and tween, causing some weird behaviour.
It would be cool to have the speed property there as well.

(I will try to create a PR for this this week)

PR #1161, opened 2018-01-03
Add speed property to delay and tween

matthijsgroen:
Fixes https://github.com/craftyjs/Crafty/issues/1160

This adds a `.delaySpeed` and `.tweenSpeed` property to the `Delay` and `Tween` components.

This works like the `.animationSpeed` for `SpriteAnimation`. The default is 1, setting it to 2 makes everything twice as fast, setting it to 0.5 makes it twice as slow. This can be applied when delays and tweens are in progress, and is useful to create slow-motion effects in games.

### DIFF
--- a/src/core/time.js
+++ b/src/core/time.js
@@ -8,6 +8,14 @@
  * This syncs with Crafty's internal clock, and so should generally be preferred to using methods such as `setTimeout`.
  */
 module.exports = {
+    /**@
+     * #.delaySpeed
+     * @comp Delay
+     *
+     * The rate of the delay.  This property defaults to 1.
+     */
+    delaySpeed: 1,
+
     init: function () {
         this._delays = [];
         this._delaysPaused = false;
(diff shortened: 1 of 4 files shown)


matthijsgroen:
To see the effect in action:

slow-mo

src/core/time.js
```diff
@@ -8,6 +8,14 @@
  * This syncs with Crafty's internal clock, and so should generally be preferred to using methods such as `setTimeout`.
  */
 module.exports = {
+    /**@
+     * #.delaySpeed
+     * @comp Delay
+     *
+     * The rate of the delay.  This property defaults to 1.
```
starwed: Maybe add a clarifying comment, e.g. that a speed of 2 means the delay will complete in half the time?  (I feel like it's less intuitive what "speed" or "rate" means for delays.)

starwed:
Makes sense to me that these components would have this functionality.

I do wonder if it might make more sense for your use case to be able to alter the rate of time at the source -- to allow changing the relative speed of the entire game as part of `Crafty.timer`?

matthijsgroen:
It is best to do it per component for me. For instance, when I pause the game, I stop a GameLoop event that is bound to the EnterFrame, and do not stop the EnterFrame. That is because when I pause the game, I need Crafty to render a Pause menu, and need the controls to navigate that menu.

starwed:
Looks good, merging.

PR #1162, opened 2018-01-03
Fix issue with removing a component that was not there

matthijsgroen:
When removing a component that was never added, Crafty produces an error.

```
Uncaught TypeError: Cannot convert undefined or null to object
    at init.removeComponent (crafty.js:2699)
```


starwed (approved):
Thanks for updating the unit test as well!

PR #1164, opened 2018-01-09
Remove unused line of code

matthijsgroen:
nothing special, but this line is not used since the Moved event is ... removed


starwed (approved):

starwed:
Good catch, merged.

PR #1168, opened 2018-01-15
Fix MBR narrow-phase collision logic

starwed:
The narrow phase collision check was always returning true for 'MBR' type collisions.

(If you look at the fix, you'll see this was a pretty dumb goof on my part!)

I added unit tests for this specific case.

### DIFF
--- a/tests/unit/spatial/collision.js
+++ b/tests/unit/spatial/collision.js
@@ -103,6 +103,22 @@
     _.ok('overlap' in results[0], "expected overlap value");
   });
 
+  test("hit -- collision vs. non collision tests", function(_){
+    var e1 = Crafty.e("2D, Collision")
+      .attr({x: 0, y: 0, w: 2, h: 2});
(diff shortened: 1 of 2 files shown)


tests/unit/spatial/collision.js
```diff
@@ -103,6 +103,22 @@
...
   });
 
+  test("hit -- collision vs. non collision tests", function(_){
+    var e1 = Crafty.e("2D, Collision")
+      .attr({x: 0, y: 0, w: 2, h: 2});
+    var e2 = Crafty.e("2D")
+      .attr({x: 0, y: 0, w: 2, h: 2});
+    var results = e1.hit('2D');
+    _.strictEqual(results.length, 1, "exactly one collision");
+    _.strictEqual(results[0].type, "MBR", "expecdted MBR collision type");
```
matthijsgroen, line 113: Typo 😅

Issue #1166, opened 2018-01-13
Release 0.9.0


starwed:
- [x] Create testing branch and upload beta docs
- [x] Automated tests pass
- [x] Prepare updates to site + non-autogenerated docs.

**Browser checks**
Let's make sure that we at least nominally work in the full list of browsers (i.e. a basic scene will render.)

- [ ] IE9 (oldest browser we claim to support)
- [x] IE11
- [x] Edge (I saw some choppiness here, but I think it was my testing setup?  Rendered correctly, though.)
- [x] Firefox latest
- [x] Firefox for Android
- [x] Chrome latest
- [x] Chrome for Android
- [x] Safari
- [x] Safari on iPhone (<-- I have no easy way to check this, but we do run at least some basic tests on iphone devices)

starwed:
There are a couple of tests failing on saucelabs, due to "Unsupported OS/browser/version/device combo".

I'll probably have to update some combos to more recent systems.

starwed:
Updated the tests targets in [this commit](https://github.com/craftyjs/Crafty/commit/c2aa19538925e856e0654be87f29e899a78992d8), they now pass.

(Essentially we try to test the oldest browsers we support in addition to the most recent.  SauceLabs makes it easy to specify the most recent with the `latest` keyword, but  we have to manually update the tests targeting old browsers as they remove support for them.)

starwed:
Here is the first [release candidate](https://github.com/craftyjs/Crafty/releases/tag/0.9.0-rc1).

starwed:
[Second release candidate](https://github.com/craftyjs/Crafty/releases/tag/0.0.9-rc2)

starwed:
In doing some browser checks, I noticed the following small issues:
- Chrome (desktop+android) seems to be applying the viewport transformations slightly differently for DOM layers.  I think this is essentially fine -- the only reason why I noticed it is that my demo file has canvas/dom/webgl layers setup with background elements which are flush against each other, so even a single pixel difference in rendering the background image is noticeable.  
- If you don't enable `multitouch`,  I found that firefox/chrome would sometimes trigger mouse events twice on a tap -- not 100% sure why that is.  I'll see if I can figure that out, but I'm not sure I consider it a blocker; you can work around it by handling touch directly, and I suspect it might be present in older releases too?

starwed:
I think I've figured out the touch issue; potential fix [here](https://github.com/craftyjs/Crafty/pull/1172)

[...]

PR #1173, opened 2018-01-26
Update docs to use Move instead of Moved

starwed:
This should address the misleading documentation mentioned [here](https://github.com/craftyjs/Crafty/issues/1166#issuecomment-359474536).


matthijsgroen:
Trick with this (for me at least) was if a movement is heading south-east, and the entity has a collision only on the south side, to keep the motion eastward intact (gliding over the other entity). with reverting both axis, the entity gets 'stuck' when going south-east

starwed:
That's true, but in terms of just updating this example I think the simpler collision resolution is fine.

*e:* Probably we should have a general [doc](http://craftyjs.com/documentation/) section on collision management that explains these things.

PR #1189, opened 2018-09-11
[ Feature ]  Support Assets encoded in Data URI

### LINKED ISSUE #1188: Support data-uri ?
It seems that Cratfy does not support data-uri encoded assets. I want to wrap all my assets (data-uri) up in one big single js file, but once using the data uri as assets' url, it doesn't work.

Simply I changed the `isValidImage` method of [loader.js](https://github.com/chux0519/Crafty/blob/develop/src/core/loader.js#L266)

```
return f.startsWith('data:') || Crafty.imageWhitelist.indexOf(getExt(f)) !== -1;
```

It works, but will cause very long key in Crafty.assets.

Any idea / suggestion of this ?

TechQuery:
- [x] [ Add ]  Support Assets encoded in Data URI
- [x] [ Optimize ]  Simplify core logic of the loader

Resolves #1188

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,10 +82,5 @@
     "wdio-qunit-framework": "mucaho/wdio-qunit-framework#master",
     "wdio-sauce-service": "^0.3.1",
     "webdriverio": "^4.6.2"
-  },
-  "browserify": {
-    "transform": [
-      "brfs"
-    ]
   }
 }
--- a/src/core/loader.js
+++ b/src/core/loader.js
@@ -1,11 +1,11 @@
-var Crafty = require('../core/core.js');
+var Crafty = require('../core/core.js'), Utility = require('./utility');
 
 module.exports = {
     /**@
      * #Crafty.assets
      * @category Assets
      * @kind Property
-     * 
+     *
      * An object containing every asset used in the current Crafty game.
      * The key is the URL and the value is the `Audio` or `Image` object.
      *
@@ -23,7 +23,7 @@ module.exports = {
      * #Crafty.paths
      * @category Assets
      * @kind Method
-     * 
+     *
      * @sign public void Crafty.paths([Object paths])
      * @param paths - Object containing paths for audio and images folders
      *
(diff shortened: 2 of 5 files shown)


starwed:
Hey, thanks for the PR!  I might not be able to get to looking over it until this weekend.

One thing to note is that using the `Promise` object would [break compatibility with IE 9 - 11 completely](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) which Crafty nominally supports.  (But for many reasons, it might be time to just break that chain.)

TechQuery:
@starwed 
To import a Promise polyfill from CDN is so easy now, and it's good for making the code simpler & stronger.

At the same time, I also suggest to refactor this project with ECMAScript 6+ & Babel 7, it will make the core logic clearer.

starwed:
Looks good, merged!  Thanks for the PR.

I made some non-functional tweaks (updating qunit in our package.json to match the version you'd manually updated to, upgrading our jshint rules to allow es6, and some minor formatting to whitespace.)  [Merge commit here](https://github.com/craftyjs/Crafty/commit/3806f551211e085aa9cce7fc60e34b93600c3226).

> I also suggest to refactor this project with ECMAScript 6+ & Babel 7, it will make the core logic clearer.

I think I agree with this (and see #1106), at least with moving to ES6 where possible.  I updated our linting rules to allow es6.

TechQuery:
@starwed When is this feature published as a new version? A product of my company is based on it.

starwed:
>@starwed When is this feature published as a new version? A product of my company is based on it.

Each commit to develop pushes a build to the [Crafty-Distro](https://github.com/craftyjs/Crafty-Distro) repository, so if you just need the `crafty.js` file that can work.

A full release (including pushing to npm, etc) generally happens when there's a useful enough set of changes to warrant it, and there aren't any additional breaking changes planned that would cause churn.  In this case the fix includes breaking changes (using ES6 functionality), so I'd want to make some related changes before doing a full release.

If publishing to the npm is important, I can probably publish a beta/alpha build?  Let me know if that's useful.

Issue #1201, opened 2018-12-04
Collision does not work for rotated entities with updated origin


adbo28:
I am using the following piece of code to check for collisions and to prevent them from moving, if needed.
It works well until I change the rotation origin (`this.origin("center")`) and rotate the `this` object. The erroneous behavior can be easily shown by adding the `SolidHitBox` component to the respective entity.

``` javascript
var hitDatas, hitData;
if ((hitDatas = this.hit('Wall'))) { // check for collision with walls
    hitData = hitDatas[0]; // resolving collision for just one collider
    if (hitData.type === 'SAT') { // SAT, advanced collision resolution
        // move player back by amount of overlap
        this.x -= hitData.overlap * hitData.nx;
        this.y -= hitData.overlap * hitData.ny;
    } else { // MBR, simple collision resolution
        // move player to previous position
        this.x = evt._x;
        this.y = evt._y;
    }
}
```

starwed:
>It works well until I change the rotation origin (this.origin("center")) and rotate the this object. The erroneous behavior can be easily shown by adding the SolidHitBox component to the respective entity.

Thanks for the bug report!  Could you describe the error in a bit more detail?  (Or provide a self-contained example?)

adbo28:
Hi, sure the subset of my code which displays the reported behaviour follows:

`<html>
    <head>
    </head>
    <body>

        <div id="game"></div>
        
        <script type="text/javascript" src="https://rawgithub.com/craftyjs/Crafty/release/dist/crafty-min.js">
        </script>
            
        <script>
            Crafty.init(500,500, document.getElementById('game'));

            Crafty.c("Rocket", {
                init: function() {
                    this.addComponent("2D, DOM, Color, Motion, Collision, SolidHitBox");
                    this.rotating = 0;
                    this.rotation = Crafty.math.randomNumber(0, 359);
                    this.x = Crafty.math.randomNumber(50, 450);
                    this.y = Crafty.math.randomNumber(50, 450);
                    this.w = 30;
                    this.h = 30;
                    this.origin("center");
                },

                moveOneStep: function(evt) {
                    //rotation
                    this.rotation += this.rotating;

                    //check for collision after rotation
                    var hitDatas, hitData;
                    if ((hitDatas = this.hit('Wall'))) { // check for collision with walls
                        hitData = hitDatas[0]; // resolving collision for just one collider
                        this.rotation -= this.rotating;
                        return this;
                    }

                    //move
[...]

Issue #1200, opened 2018-12-03
Random tile of tilemap not rendering with canvas/WebGL when rotated


r3ndd:
I have the following code for my scene:

```
var tiles;

var addTile = function (type, x, y, rot, tile_size) {
    var tile = Crafty.e("Tile").attr({x: x * tile_size, y: y * tile_size, rotation: rot || 0});
    tile.type(type);
    tile.origin("center");
    tiles.push(tile);
    if (type == "dirt_water")
        console.log(tile);
};

var addTileRows = function (type, startY, endY, width, rot, tile_size) {
    for (var y = startY; y < endY; y++) {
        for (var x = 0; x < width; x++) {
            addTile(type, x, y, rot, tile_size);
        }
    }
}

Crafty.scene("Main", function (data) {
    var width = 30;
    tiles = [];

    addTileRows("water", 0, 2, width, 0, data.tile_size);
    addTileRows("grass_water", 2, 3, width, -90, data.tile_size);
    addTileRows("grass", 3, 5, width, 0, data.tile_size);
    addTileRows("dirt_grass", 5, 6, width, -90, data.tile_size);
    addTileRows("dirt", 6, 10, width, 0, data.tile_size);
    addTileRows("dirt_water", 10, 11, width, 90, data.tile_size);
    addTileRows("water", 11, 12, width, 0, data.tile_size);
    addTileRows("sand_water", 12, 13, width, -90, data.tile_size);
    addTileRows("sand", 13, 16, width, 0, data.tile_size);
});
```

The tile component:
```
Crafty.c("Tile", {
    required: "2D, WebGL",

    init: function () {
        this.size = window.tile_size;
    },

    type: function (tile_name) {
        this.addComponent("spr_" + tile_name);
    }
});
```

However, I am getting an issue where one of the rotated tiles is not rendered: https://i.gyazo.com/9a526fd6febb991011f32043ca73fdce.png

Only happens on canvas and WebGL, DOM works fine. Adding 1 pixel to the x position of the tile makes the issue disappear, but that obviously isn't a viable fix for a game.

r3ndd:
I should add that this is version 0.9.0, but after checking it appears 0.8.0 has the same issue.

starwed:
Hmm, definitely seems like a bug.  If it occurs in WebGL *and* Canvas layers, but not DOM, and only for rotated entities, I bet it's an error determining whether the entity falls within the viewport.

[...]

PR #1202, opened 2018-12-08
Update dependent positions on origin change

starwed:
Setting the origin 'mid scene' was not properly updating the overall entity state.

When the origin of a rotated entity changes, we need to:
- Adjust the position of attached children (such as collision hitboxes)
- Recalculate the MBR
- Update the spatial map, since the MBR might have changed
- Flag for re-rendering

### DIFF
--- a/src/spatial/2d.js
+++ b/src/spatial/2d.js
@@ -925,6 +925,7 @@ Crafty.c("2D", {
         }
 
         this.trigger("OriginChanged");
+        this.trigger("Invalidate");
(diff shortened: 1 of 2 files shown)


starwed:
This should fix  #1201, and *might* also resolve #1200.  (For the latter, it could be that the MBR/map not being updated upon setting the origin could cause a rotated entity on the edge of the view to not be rendered -- the pre origin change position would have been off of the screen)

